PR #1, opened 2017-05-03
removed the delay for spi_comm

neelrparikh:
I think this should work. I wasn't too sure on the pin number used by boot0, I have used GPIO4 on the ESP and B0 on the ST

### DIFF
--- a/board/main.c
+++ b/board/main.c
@@ -732,6 +732,12 @@ int spi_total_count = 0;
 uint8_t spi_tx_buf[0x44];
 
 void handle_spi(uint8_t *data, int len) {
+  // handshake with esp
+  // Set boot0 to output and pull high
+  GPIOB->MODER &= ~(GPIO_MODER_MODER0);
+  GPIOB->MODER |= GPIO_MODER_MODER0_0;
+  GPIOB->ODR |= GPIO_ODR_OD0; 
+  
   memset(spi_tx_buf, 0xaa, 0x44);
   // data[0]  = endpoint
   // data[2]  = length
@@ -740,22 +746,10 @@ void handle_spi(uint8_t *data, int len) {
   *resp_len = 0;
   switch (data[0]) {
     case 0:
-	  // handshake with esp
-	  // Set boot0 to output and pull high
-	  GPIOB->MODER &= ~(GPIO_MODER_MODER0);
-	  GPIOB->MODER |= GPIO_MODER_MODER0_0;
-	  GPIOB->ODR |= GPIO_ODR_OD0; 
-  
       // control transfer
       *resp_len = usb_cb_control_msg((USB_Setup_TypeDef *)(data+4), spi_tx_buf+4, 0);
       break;
     case 1:
-	  // handshake with esp
-	  // Set boot0 to output and pull high
-	  GPIOB->MODER &= ~(GPIO_MODER_MODER0);
-	  GPIOB->MODER |= GPIO_MODER_MODER0_0;
-	  GPIOB->ODR |= GPIO_ODR_OD0; 
-  
       // ep 1, read
       *resp_len = usb_cb_ep1_in(spi_tx_buf+4, 0x40, 0);
       break;
--- a/board/main.c
+++ b/board/main.c
@@ -737,7 +737,7 @@ void handle_spi(uint8_t *data, int len) {
   GPIOB->MODER &= ~(GPIO_MODER_MODER0);
   GPIOB->MODER |= GPIO_MODER_MODER0_0;
   GPIOB->ODR |= GPIO_ODR_OD0; 
-  
+
   memset(spi_tx_buf, 0xaa, 0x44);
   // data[0]  = endpoint
   // data[2]  = length
@@ -763,7 +763,7 @@ void handle_spi(uint8_t *data, int len) {
       break;
   }
   spi_tx_dma(spi_tx_buf, 0x44);
-  
+
   // signal transfer ready
   GPIOB->ODR &= ~(GPIO_ODR_OD0); 
 }
@@ -819,7 +819,7 @@ void DMA2_Stream3_IRQHandler(void) {
   // Set boot0 to old state
   GPIOB->MODER &= ~(GPIO_MODER_MODER0); 
   GPIOB->PUPDR |= GPIO_PUPDR_PUPDR0_0;
-  
+
   // reenable interrupt
   //EXTI->IMR |= (1 << 4);
   //puts("stop\n");
--- a/board/main.c
+++ b/board/main.c
@@ -740,10 +740,22 @@ void handle_spi(uint8_t *data, int len) {
   *resp_len = 0;
   switch (data[0]) {
     case 0:
+	  // handshake with esp
+	  // Set boot0 to output and pull high
+	  GPIOB->MODER &= ~(0x3);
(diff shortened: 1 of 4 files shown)


board/main.c
```diff
@@ -740,10 +740,22 @@ void handle_spi(uint8_t *data, int len) {
   *resp_len = 0;
   switch (data[0]) {
     case 0:
+	  // handshake with esp
+	  // Set boot0 to output and pull high
+	  GPIOB->MODER &= ~(0x3);
```
geohot: Don't use a constant here

boardesp/proxy.c
```diff
@@ -51,10 +51,14 @@ static int ICACHE_FLASH_ATTR __spi_comm(char *dat, int len, uint32_t *recvData,
...
 
-  // give the ST time to be ready, 1 ms
-  // TODO: how can we do this better?
-  os_delay_us(1000);
-
+  // use the boot0 pin as a handshake with the ST
+  // set boot0 as input
+  gpio_output_set(0,0,0,(1 << 4));
+  //wait for boot0 to be set to LOW by ST, counter to avoid infinite loop
+  for(int i = 0;(gpio_input_get() & (1 << 4)) && i < 100; i++) {
```
geohot: What happens if this counter expires and data still isn't ready?

board/main.c
```diff
@@ -757,6 +769,9 @@ void handle_spi(uint8_t *data, int len) {
       break;
   }
   spi_tx_dma(spi_tx_buf, 0x44);
```
geohot: Is this the only place we use SPI?

board/main.c
```diff
@@ -740,10 +740,22 @@ void handle_spi(uint8_t *data, int len) {
   *resp_len = 0;
   switch (data[0]) {
     case 0:
+	  // handshake with esp
+	  // Set boot0 to output and pull high
+	  GPIOB->MODER &= ~(0x3);
+	  GPIOB->MODER |= GPIO_MODER_MODER0_0;
+	  GPIOB->ODR |= (1);
```
geohot: Why is 1 in parens?

boardesp/proxy.c
neelrparikh: The counter makes it wait for 1ms, same as before. If the data isn't ready it will be an issue. The only other way I can think of would be to use an infinite loop and always poll but that would crash the system if spi fails. Even using a ack would require a counter to avoid an infinite loop. Any suggestions?

board/main.c
neelrparikh: Yes, this is the only place where we need to receive data over SPI, we transmit data in spi_flasher.h:122 but there's no need for this there as we are no waiting for receiving data to be ready.

geohot: Hmm, but does the ESP know this? Won't it still wait?

neelrparikh:
Done, sorry about that I got messy with the constants. Let me know if we can make some changes for the counter, can't think of anything right now (worst case we can increase the delay).

geohot:
lol yea, i'd just increase the delay to 10ms.

are you 100% sure you don't need this on the spi_flasher?

board/main.c
neelrparikh: Let me just go through it again to be 100% sure

neelrparikh:
I don't think this method will work, I just realized spi_flasher uses boot0 while it is using spi. I added the ack code to spi_flasher but doubt it will work. Let me think over it and post changes tomorrow.
For now I increased the delay and added an os_print may be later we can add some kind of error message?

geohot:
it helps to understand what we use the boot pin for, it's just a gpio connecting the processors.

also, the library to interact with this is also open source in lib/panda.py

neelrparikh:
I think I got a work around for this, when accessing spi_flasher the spi has not yet been initialized so we can use the cs pin for handshaking to indicate that it has entered the flasher, after which we initialize the spi and use the boot0 pin for handshaking. Do you think it will work? I think it should.

geohot:
maybe, write it and i'll try.

neelrparikh:
okay i am pushing the code

boardesp/webserver.c
```diff
@@ -62,6 +62,13 @@ void st_set_boot_mode(int boot_mode) {
     // boot mode (pull low)
     gpio_output_set(0, (1 << 4), (1 << 4), 0);
     st_reset();
+	// use cs as a handshake for spi_flasher
+	// set cs as input and wait for it to go high
+	gpio_output_set(0, 0, 0, (1 << 5));
+	//TODO: remove infinite loop
```
geohot: Not okay.

board/spi_flasher.h
```diff
@@ -53,6 +62,12 @@ void spi_flasher() {
     lastval = val;
 
     if (rcv && (DMA2->LISR & DMA_LISR_TCIF2)) {
+	  // handshake with esp
```
geohot: Are these tabs!?

board/spi_flasher.h
```diff
@@ -53,6 +62,12 @@ void spi_flasher() {
...
+	  // Set boot0 to output and pull high
+	  GPIOB->MODER &= ~(GPIO_MODER_MODER0);
+	  GPIOB->MODER |= GPIO_MODER_MODER0_0;
+	  GPIOB->ODR |= GPIO_ODR_OD0;
```
geohot: This doesn't build

boardesp/proxy.c
```diff
@@ -51,10 +51,17 @@ static int ICACHE_FLASH_ATTR __spi_comm(char *dat, int len, uint32_t *recvData,
...
+  // use the boot0 pin as a handshake with the ST
+  // set boot0 as input
+  gpio_output_set(0,0,0,(1 << 4));
+  //wait for boot0 to be set to LOW by ST, counter to avoid infinite loop
+  for(int i = 0;(gpio_input_get() & (1 << 4)) && i < 1000; i++) {
+	os_delay_us(10);
+  }
+  // print error if handshake not successful 
+  if(gpio_input_get() & (1 << 4))
+	os_print("ERROR: SPI receive failed\n");
```
geohot: What is os_print? Did you build this?

boardesp/webserver.c
neelrparikh: I can remove the infinite loop but did not know the delay to keep for the ST to start up would a worst case of 10ms be good?

board/spi_flasher.h
neelrparikh: Ahh! my ide inserted tabs I'll get rid of them

boardesp/proxy.c
neelrparikh: It's suppose to be os_printf sorry about that.

board/spi_flasher.h
neelrparikh: Can't figure why this is not building, I will try building and may be run it on an ESP I have.

neelrparikh:
Sorry about the build errors, I was not able to compile in my environment. I managed to compile the files and removed all the build errors. Thanks.

geohot:
Good enough for a microinternship. We were in a rush to get pandas out and I ended up writing a variant of this, so I won't actually merge this. (you can read, it's committed). Viv will be in touch with microinternship details.

neelrparikh:
Thank you, sorry for the delay I was stuck in school today. I will have a look on what you have done! Sounds good thanks again looking forward to meeting you.

PR #3, opened 2017-05-09
Single-wire GMLAN support

vntarasov:
- Fixes off-by-one error in can bus speed calculation, that was preventing single-wire GMLAN from working on my Chevy Volt gen 2.
- Instead of just replacing 0x001c002d with 0x001c002c, this PR makes it easier to understand, debug, and change the CAN speed configuration code, should someone need a different bus rate in the future.
- Enables sending/receiving extended addresses (majority of messages on single-wire GMLAN on my car).


geohot:
CAN speed should probably be refactored even more to support changing over API. But this is certainly an improvement, will merge. And nice with the extended, it's in boardd but wasn't ported.

PR #5, opened 2017-06-09
Clean up download mode script and add dependency to installer.

diamondman:
Added dependencies to get_sdk.sh. Cleaned up code in enter_download_m…ode.py and made compatible with py2 and py3.

### DIFF
--- a/board/tools/enter_download_mode.py
+++ b/board/tools/enter_download_mode.py
@@ -1,27 +1,33 @@
 #!/usr/bin/env python
-import usb1
+from __future__ import print_function
+
+import sys
 import time
-import traceback
+import usb1
 
-if __name__ == "__main__":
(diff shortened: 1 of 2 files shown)


board/tools/enter_download_mode.py
```diff
@@ -1,27 +1,33 @@
...
+    handle.controlWrite(usb1.TYPE_VENDOR | usb1.RECIPIENT_DEVICE, 0xd1, 0, 0, b'')
+  except usb1.USBErrorIO as e:
+    print("Device download mode enabled.")
+    time.sleep(1)
+  else:
+    print("Device failed to enter download mode.")
+    sys.exit(1)
 
-  handle = None
+def find_first_panda(context=None):
```
geohot: This should really call into lib/panda.py

PR #6, opened 2017-06-12
Remove inline keywords, cleaned up build.md, set compiler standard.

### LINKED ISSUE #4: Panda firmware fails to compile with arm-none-eabi-gcc 5.4.1 20160919 (default std=gnu11)
Compiling the panda firmware on a Debian Stretch machine (using the debian provided gcc-arm-none-eabi version 5.4.1) fails complaining about inline functions not being available at link time. For example:
`undefined reference to 'early'`.

This stems from the C standard being used by GCC changing from `gnu90` to `gnu11` between gcc version 4.9 to 5.1 as detailed [here](https://stackoverflow.com/a/14737642/627525). In short, the semantics around the `inline` keyword has [changed between standards](https://stackoverflow.com/questions/16245521/c99-inline-function-in-c-file/16245669#16245669).

It appears that the original developer wrote the firmware using ubuntu's Xenial gcc-arm-none-eabi package which is currently on [version 4.9.3](https://packages.ubuntu.com/xenial/gcc-arm-none-eabi). Since Ubuntu will inevitably update to the [5.x](https://packages.debian.org/stretch/gcc-arm-none-eabi) line sometime in the future, we should fix this now.

The easiest solution is to explicitly mark the std type as  `gnu90`, thus restoring the default behavior of the older GCC version. No matter what is done, I believe that a C standard should be selected for the project to prevent these types of issues in the future.

But this issue raises a question: why are these functions inline?
The compiler is free to ignore the inline directive, as well as choose to inline a function that is not explicitly marked to be inlined [[1]](https://stackoverflow.com/a/1626269/627525). Furthermore, the firmware is compiled with -O0, and inlining is only done on optimizing compilations: "Inlining of functions is an optimization and it really “works” only in optimizing compilation. If you don't use -O, no function is really inline." [[gcc manual]](https://gcc.gnu.org/onlinedocs/gcc-3.3.6/gcc/Inline.html).

I argue that most of these inline functions should be stripped of their inline qualifier and split into a header file and a source file, as inlining can still be done across compilation units. This would also allow us to use more modern C semantics in the future. 

I am happy to make a PR with either of the proposed changes. What is the preferred way to resolve this issue?

diamondman:
Closes #4 
Removed inline keywords on all functions. 
Split early.h into a header and c file.
Updated Makefile rules to be more consistent.

### DIFF
--- a/board/Makefile
+++ b/board/Makefile
@@ -1,10 +1,8 @@
-# :set noet
 PROJ_NAME = panda
-CFLAGS  = -g -O0 -Wall
+CFLAGS  = -g -O0 -Wall -std=gnu90
 
 CFLAGS += -mlittle-endian -mthumb -mcpu=cortex-m4
 CFLAGS += -mhard-float -DSTM32F4 -DSTM32F413xx
 STARTUP_FILE = startup_stm32f413xx
 
 include build.mk
-
--- a/board/build.mk
+++ b/board/build.mk
@@ -23,7 +23,7 @@ endif
 DFU_UTIL = "./tools/dfu-util-$(MACHINE)"
 
 # this pushes the unchangable bootstub too
-all: compileall #dfu
+all: compileall dfu
 
 compileall: obj/bootstub.$(PROJ_NAME).bin obj/$(PROJ_NAME).bin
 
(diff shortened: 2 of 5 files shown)


diamondman:
Needs to be tested on hardware before merged. I can do that tomorrow.

geohot:
Excellent, will merge

PR #7, opened 2017-06-13
Add steering safety check for cr-v

energee:


energee:
@rbiasini This commit is in HEAD but not in the hash referenced by openpilot so it looks like the safety check is in not place for CR-V drivers running the release branch.

rbiasini:
@energee , you are correct. We'll cherry-pick this commit into the openpilot's submodule

PR #10, opened 2017-06-14
Created python package and implemented industry best practices.

diamondman:
newly created panda package is ready for upload to pypy with all recommended fields and features.
Supports python 2 and 3 (don't have hardware for all tests, so a few may fail in Python3)
    Support for python 3 is greatly encouraged due to the ever approaching sunset date on 2.x http://www.python3statement.org/
Tests can be run from any directory without having to have env variables set. These tests should be split into unit tests and tools in the future.


geohot:
Can we call it panda? There is already a panda on pypi.

https://pypi.python.org/pypi/panda

diamondman:
We can make it anything we want, but you are correct that there will be a conflict on pipi if we upload it. If the goal is to make it available on the package manager, lets come up with a better name.

diamondman:
Possible options are
commaaipanda
pandaai
driverless_panda
canpanda
etc

We can totally leave it like it is and change the name when we go to put it on a package manager.

PR #13, opened 2017-06-16
Usb bulk to interrupt

diamondman:
This patch should not be merged until a satisfactory driver is available.

This patch relies on the (Reformatted usb descriptor table for readability. #12).

### DIFF
--- a/board/usb.h
+++ b/board/usb.h
@@ -128,29 +128,49 @@ uint8_t device_desc[] = {
 
 uint8_t configuration_desc[] = {
   DSCR_CONFIG_LEN, DSCR_CONFIG_TYPE, // Length, Type,
-  0x27, 0x00, // Total Len (uint16)
+  TOUSBORDER(0x0045), // Total Len (uint16)
   0x01, 0x01, 0x00, // Num Interface, Config Value, Configuration
   0xc0, 0x32, // Attributes, Max Power
-  // interface 0
+  // interface 0 ALT 0
   DSCR_INTERFACE_LEN, DSCR_INTERFACE_TYPE, // Length, Type
   0x00, 0x00, 0x03, // Index, Alt Index idx, Endpoint count
-  0xff, 0xFF, 0xFF, // Class, Subclass, Protocol
+  0XFF, 0xFF, 0xFF, // Class, Subclass, Protocol
   0x00, // Interface
-  // endpoint 1, read CAN
-  DSCR_ENDPOINT_LEN, DSCR_ENDPOINT_TYPE, // Length, Type
-  ENDPOINT_RCV | 1, ENDPOINT_TYPE_INT, // Endpoint Num/Direction, Type
-  TOUSBORDER(0x0040), // Max Packet (0x0040)
-  0x05, // Polling Interval (5 frames)
-  // endpoint 2, send serial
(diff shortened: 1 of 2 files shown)


board/usb.h
```diff
@@ -84,37 +84,77 @@ int  usb_cb_ep1_in(uint8_t *usbdata, int len, int hardwired);
...
+  DSCR_INTERFACE_LEN, DSCR_INTERFACE_TYPE, // Length, Type
+  0x00, 0x00, 0x03, // Index, Alt Index idx, Endpoint count
+  0xff, 0xFF, 0xFF, // Class, Subclass, Protocol
+  0x00, // Interface
   // endpoint 1, read CAN
-  0x07, 0x05, 0x81, 0x02, 0x40, 0x00, 0x00,
+  DSCR_ENDPOINT_LEN, DSCR_ENDPOINT_TYPE, // Length, Type
+  ENDPOINT_RCV | 1, ENDPOINT_TYPE_INT, // Endpoint Num/Direction, Type
+  TOUSBORDER(0x0040), // Max Packet (0x0040)
+  0x0A, // Polling Interval (10 frames)
```
geohot: Currently we are polling at 200hz with boardd. Can we make this 5?

diamondman: This can be anything as long as the kernel is able to handle it. I will set it to 5 and check that it works.

diamondman:
So far the new interval of 5 works fine with the new python library.

board/usb.h
```diff
@@ -643,29 +672,51 @@ void usb_irqhandler(void) {
...
+
+    // Handle the two interface alternate settings. Setting 0 is has
+    // EP1 as bulk. Setting 1 has EP1 as interrupt. The code to handle
+    // these two EP variations are very similar and can be
+    // restructured for smaller code footprint. Keeping split out for
+    // now for clarity.
+
+    //TODO add default case. Should it NAK?
+    switch(current_int0_alt_setting){
+    case 0: ////// Bulk config
```
geohot: Looks easy to write this with half as many lines. if (can_rx_q.w_ptr != can_rx_q.r_ptr || current_int0_alt_setting == 0)

board/usb.h
```diff
@@ -201,11 +224,13 @@ void USB_WritePacket(const uint8_t *src, uint16_t len, uint32_t ep) {
     puts("writing ");
     hexdump(src, len);
   #endif
+    uint8_t numpacket = (len+(MAX_RESP_LEN-1))/MAX_RESP_LEN;
```
geohot: Why so many spaces?

board/usb.h
diamondman: Agreed, the comment above explained my reasoning. I am happy to leave it as it is or compress it.

board/usb.h
diamondman: My editor decided to be clever.

board/usb.h
diamondman: The reasoning being that compressing the logic into that form can make it harder to edit in the future, or even understand the goal of the code. If a 3rd alternate setting were added, the compressed form of the code could have weird and unexpected behavior. The editor could not just edit this code. First it would need to be expanded, edited, and recompressed. Yes many can do it in their head, but this form was just to make it explicitly clear what is happening.

PR #14, opened 2017-06-19
New int python driver

diamondman:
First pass at new Python 2/3 api for interfacint with the new Panda firmware (with USB interrupt endpoints for reading data).

Relies on #10 (Created python package and implemented industry best practices.).

Should be merged with #13 (Firmware Usb bulk to interrupt).


geohot:
Will close. No need for interrupt from Python driver.

PR #15, opened 2017-06-19
Removed dependence on precompiled dfu-util binaries.

diamondman:
The forked dfu-util --reset-stm32 is included in mainstream under a different name.
https://github.com/dsigma/dfu-util/issues/3#issuecomment-309504691


geohot:
This breaks functionality on the phone. NEOS can't apt-get dfu-util.

Suggested solution: Build updated dfu-util for phone and leave only that one in the package. Eventually we will ship it with NEOS.

diamondman:
Which of the three binaries are used for phones (dfu-util-aarch64-linux or dfu-util-aarch64)? That way I can edit the Makefile to use dist under all other conditions.

geohot:
dfu-util-aarch64 is used for the phone.

diamondman:
outdated.

PR #18, opened 2017-07-04
Fix legacy board build

vntarasov:
- CAN3 is not available for legacy board.
- Minor stylecop.


diamondman:
Thanks for cleaning up the style. The CAN3 stuff will be unnecessary soon, but I am happy to add it back for now since it seems to be causing you some issues.

PR #22, opened 2017-07-13
Redux/can speed

diamondman:
New Feature: Set and read can/GMLAN baud rate (with wifi support).
USB Messages needed to have their direction bit set so that wifi could know to read data.
Tested with loopback and throughput tests.


geohot:
Will not merge. Please break into smaller pieces.

PR #23, opened 2017-07-13
Modularize safety modes to encourage 3rd party safety code contribution.

diamondman:
Allows multiple safety modes to be written and selected from at runtime.

### DIFF
--- a/board/can.h
+++ b/board/can.h
@@ -1,4 +1,4 @@
-void can_init(uint8_t bus_number) {
+void can_init(uint8_t bus_number, int silent) {
   CAN_TypeDef *CAN = CANIF_FROM_BUS_NUM(bus_number);
   set_can_enable(CAN, 1);
 
@@ -22,7 +22,7 @@ void can_init(uint8_t bus_number) {
     CAN->BTR |= CAN_BTR_SILM | CAN_BTR_LBKM;
   }
 
-  if (!controls_allowed) {
+  if (silent) {
     CAN->BTR |= CAN_BTR_SILM;
   }
 
--- a/board/can.h
+++ b/board/can.h
@@ -1,4 +1,4 @@
-void can_init(uint8_t bus_number, int silent) {
+void can_init(uint8_t bus_number) {
   CAN_TypeDef *CAN = CANIF_FROM_BUS_NUM(bus_number);
   set_can_enable(CAN, 1);
 
@@ -22,7 +22,7 @@ void can_init(uint8_t bus_number, int silent) {
     CAN->BTR |= CAN_BTR_SILM | CAN_BTR_LBKM;
   }
 
-  if (silent) {
+  if (!controls_allowed) {
     CAN->BTR |= CAN_BTR_SILM;
   }
 
--- a/board/main.c
+++ b/board/main.c
@@ -14,8 +14,8 @@
 // can_num_lookup: Translates from 'bus number' to 'can number'.
 // can_forwarding: Given a bus num, lookup bus num to forward to. -1 means no forward.
 
-// old:         CAN1 = 1   CAN2 = 0
-// panda:       CAN1 = 0   CAN2 = 1   CAN3 = 4
+// NEO:         Bus 1=CAN1   Bus 2=CAN2
+// Panda:       Bus 0=CAN1   Bus 1=CAN2   Bus 3=CAN3
 #ifdef PANDA
   CAN_TypeDef *cans[] = {CAN1, CAN2, CAN3};
   uint8_t bus_lookup[] = {0,1,2};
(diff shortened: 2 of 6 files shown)


board/main.c
```diff
@@ -1,21 +1,42 @@
...
+// bus num: Can bus number on ODB connector. Sent to/from USB
+//    Min: 0; Max: 127; Bit 7 marks message as receipt (bus 129 is receipt for but 1)
+// cans: Look up MCU can interface from bus number
+// can number: numeric lookup for MCU CAN interfaces (0 = CAN1, 1 = CAN2, etc);
+// bus_lookup: Translates from 'can number' to 'bus number'.
+// can_num_lookup: Translates from 'bus number' to 'can number'.
+// can_forwarding: Given a bus num, lookup bus num to forward to. -1 means no forward.
+
 // old:         CAN1 = 1   CAN2 = 0
 // panda:       CAN1 = 0   CAN2 = 1   CAN3 = 4
```
vntarasov: Comment seems to be out of date.

vntarasov:
Is the intent to allow bus_num to be assigned independently from can number, even though currently they are the same?

board/main.c
```diff
@@ -875,19 +860,8 @@ int main() {
...
-    can_init(CAN3, 1);
-  #endif
-#else
-  can_init(CAN1, 0);
-  can_init(CAN2, 0);
-  #ifdef CAN3
-    can_init(CAN3, 0);
-  #endif
-#endif
+  for(i=0; i < CAN_MAX; i++)
```
vntarasov: `for (int i`?

board/can.h
```diff
@@ -21,7 +22,7 @@ void can_init(CAN_TypeDef *CAN, int silent) {
     CAN->BTR |= CAN_BTR_SILM | CAN_BTR_LBKM;
   #endif
 
-  if (silent) {
+  if (!controls_allowed) {
```
geohot: Messages are allowed in some safety modes while controls aren't.

geohot:
@diamondman please rebase on top of master and address @vntarasov comments.

board/main.c
diamondman: When you use `for(int i...)` without setting the c standard to a new enough version, this error happens.

```
main.c: In function 'main':
main.c:875:3: error: 'for' loop initial declarations are only allowed in C99 or C11 mode
   for(int i=0; i < CAN_MAX; i++)
   ^
main.c:875:3: note: use option -std=c99, -std=gnu99, -std=c11 or -std=gnu11 to compile your code
build.mk:71: recipe for target 'obj/main.panda.o' failed
make: *** [obj/main.panda.o] Error 1
```

I did not feel this PR deserved to have a change in the C standard flags so I put the declaration of i at the top of the function as is suggested.

board/can.h
diamondman: Good to know. 
I was convinced they were having the same behavior because silent defaulted to on with PANDA_SAFETY, but as soon as the usb  'controls enabled' message was received (to enable or disable), silent would be disabled and never reenabled.
https://github.com/commaai/panda/pull/23/files#diff-6cd306808dae15d8f629c38dc9e817a0L655

board/main.c
diamondman: Yes that needs to be updated (can 3 = 2, not 4). Thanks for identifying it.

board/can.h
diamondman: Should silent have a way to be re-enable after controls allowed is called?

diamondman: I will split silent and controls allowed. Should silent have a way to be re-enable after the "controls allowed" usb command is called?

diamondman:
@vntarasov correct, bus_num allows us to decouple the bus number on the obd port (and the USB api) from the microcontroller's can interfaces. But no, they are not the same now. The panda has bus numbers and can numbers that coincide, but the neo has them flipped and makes translation back and forth necessary.

diamondman:
The new changes are minor, but I will run full tests on this when I get into the office.

panda/__init__.py
```diff
@@ -147,6 +151,12 @@ def get_secret(self):
 
   # ******************* configuration *******************
 
+  def set_controls_mode(self, mode=SAFETY_ALLOUTPUT):
+    self._handle.controlWrite(Panda.REQUEST_TYPE, 0xdc, mode, 0, b'')
+
+  def set_controls_allowed(self, on):
```
geohot: Huh, this function is in here twice

panda/__init__.py
```diff
@@ -147,6 +151,12 @@ def get_secret(self):
...
+    self.set_controls_mode(SAFETY_ALLOUTPUT if on else SAFETY_NOOUTPUT)
+
   def set_controls_allowed(self, on):
```
geohot: There's two set_controls_allowed

Issue #30, opened 2017-08-01
How to update panda code? How to connect with computer?


kwasnyty:
Is 'Panda Paw' needed to update the code on the Panda? Can I use a standard USB cord? How do I transfer the code to the board?

To use PandaCAN or SocketCAN with my computer, do I just plug the panda into my computer with a standard USB cord?

edsammy:
A paw is not needed to flash new firmware onto the panda (as long as you don't modify bootstub.c or anything in main.c that mentions bootstub or bootloader). You can flash the panda by running `make` in the board directory. 

You can use a standard A to A USB cable to connect your panda to your computer. When your panda is plugged into an OBDII port and you want to connect it to a computer, it is recommended that you cut the power line of the USB cable. We have found that a few computers do not like that pandas output 5.2V on the power line (this helps fast charge phones). I know that this is not an issue on my Macbook Air but cannot speak for every computer.

If you ordered a panda through our [preorder page](panda.comma.ai) then you will not have to deal with the issue mentioned above. We have updated the panda hardware to automatically shutdown its power output if it is plugged into a computer.

PR #31, opened 2017-08-04
Update README.md

palmerit:
Add a note about entering DFU mode.


edsammy:
DFU mode shouldn't be required for normal flashing (see my answer to [issue 30](https://github.com/commaai/panda/issues/30))

PR #32, opened 2017-08-07
SAFETY: disallowing controls on brake switch as this is the main canc…

rbiasini:
…ellation bit considered by cruise in pcm

### DIFF
--- a/board/honda_safety.h
+++ b/board/honda_safety.h
@@ -25,8 +25,8 @@ void safety_rx_hook(CAN_FIFOMailBox_TypeDef *to_push) {
 
   // exit controls on brake press
   if ((to_push->RIR>>21) == 0x17C) {
-    // bit 50
-    if (to_push->RDHR & 0x200000) {
+    // bit 53 or 32
+    if (to_push->RDHR & 0x200001) {
       controls_allowed = 0;
     }
   }


board/honda_safety.h
```diff
@@ -25,8 +25,8 @@ void safety_rx_hook(CAN_FIFOMailBox_TypeDef *to_push) {
 
   // exit controls on brake press
   if ((to_push->RIR>>21) == 0x17C) {
-    // bit 50
-    if (to_push->RDHR & 0x200000) {
```
rbiasini: @geohot , we need to slightly change the rule on brake cancellation. In order to avoid pcm state mismatch, we are now canceling on both BRAKE_PRESS (bit 53 in 0x17c) and BRAKE_SWITCH (bit 32 in 0x17c). Before we were only canceling on BRAKE_PRESS, however PCM seems to cancel on BRAKE_SWITCH instead. Since BRAKE_SWITCH is triggered a bit sooner, this was causing states mismatch when brake pedal was slowly pressed (very repeatable bug).

This change is already implemented in carstate.py and I think we need to change it here too. 2 questions:
- is this change right? :)
- should I make this change and merge it in top of 'panda' master? I believe 'one' is currently referencing a pretty old git# of 'panda'.

PR #42, opened 2017-09-14
Toyota torque limit

rbiasini:
Adding a speed dependent steer torque limitation


geohot:
Looks good. Once we test to validate the limits are good (torque injection), let's merge.

PR #43, opened 2017-09-17
board: standardized how health voltage is reported between revisions.

diamondman:
The REVB boards have a different voltage divider leading up to the ADC. When left uncompensated, a panda may appear to report 5V when in reality it is reading 12V, or vice versa.

Previously, health reported the raw output of the ADC. Now, voltage is reported in millivolts.

Math included in comments to deal with introduction of 'magic numbers'.


geohot:
Thanks! Merging.

PR #60, opened 2017-11-10
Adds Mac-specific instructions for setting up board and esp

energee:
Details on installing on mac and readme updates


geohot:
Nice!

PR #61, opened 2017-11-11
tool to reverse engineer a single bit field

adhintz:
Given an interesting CSV file of CAN messages and a list of background CAN messages, print which bits in the interesting file have never appeared in the background files.


geohot:
Cool!

PR #62, opened 2017-11-11
Webserver control

appleguru:
Add basic controls for setting USB mode using the built in web interface.


boardesp/webserver.c
```diff
@@ -193,8 +193,35 @@ static void ICACHE_FLASH_ATTR web_rx_cb(void *arg, char *data, uint16_t len) {
...
+        "<button onclick=\"location.href='/dcp'\" type='button'>"
+        "DCP</button>"
+        "<button onclick=\"location.href='/cdp'\" type='button'>"
+        "CDP</button>"
+        "<button onclick=\"location.href='/client'\" type='button'>"
+        "Client</button>");
+      
       espconn_send_string(&web_conn, resp);
       espconn_disconnect(conn);
+    } else if (memcmp(data, "GET /client ", 12) == 0) {
```
geohot: Instead of writing this out three times, can you write a function?

boardesp/webserver.c
```diff
@@ -193,8 +193,35 @@ static void ICACHE_FLASH_ATTR web_rx_cb(void *arg, char *data, uint16_t len) {
       } else {
         ets_strcat(resp, "<br/>esp flash file: user1.bin");
       }
+      
+      ets_strcat(resp,"<br/>Set USB Mode:"
+        "<button onclick=\"location.href='/dcp'\" type='button'>"
```
geohot: So this navigates away from the main page, can you use XHR to not?

appleguru: Yes, XHR should be fine. Might want to look at something like WebSockets or mqtt long term if we're going to be sending lots of data back and forth.

PR #84, opened 2018-02-01
Bounty: panda J2534 Driver Fix

gregjhogan:
This fixes the major issues with getting HDS and i-HDS working.


gregjhogan:
Here is a debug build with instructions on how to set it up
[j2534-driver.zip](https://github.com/commaai/panda/files/1684473/j2534-driver.zip)

nanamiwang:
@gregjhogan @diamondman @geohot 
A question regarding to the code structure:
I have seen a multi threading Tx/Rx queue mechanism in PandaJ2534Device.h/.cpp, A lot of lines of code are handling queue push/pop and inter-thread communication. This pull request also add another msg process thread.
According to the two docs blew, J2534 DLL client app is restricted to single threaded design:
https://web.archive.org/web/20130805013326/https://tunertools.com/prodimages/DrewTech/Manuals/PassThru_API-1.pdf  
page 3
```
In order to reduce programming complexity the J2534 DLL does
not support multi-threading so the User Application is restricted to a single thread design.
```
https://archive.org/stream/gov.law.sae.j2534.2002/sae.j2534.2002_djvu.txt
Chapter 6.2
```
The OEM application is limited to a single thread 
for communication with the tool manufacturer DLL/API. Multiple protocols may 
be connected and communicated on sequentially (serialized) from the single 
application thread. This will prevent the unnecessary complexity of determining 
what message responses belong to which application thread
```
In my understanding, this rule ensure that client will not send multiple Tx/Rx requests simultaneously, so the Tx/Rx queue mechanism is not necessary, for CAN bus message Tx we can just call WinUSB_WritePipe & WinUSB_ReadPipe(for echo msg recv) directly in PassthruWriteMsgs implementation, and the same for Rx. the whole DLL can be single threaded(maybe periodical msg sending need a dedicate thread). This will greatly reduce the code complexity.
Hoping for your comments.

gregjhogan:
@nanamiwang if there isn't something constantly reading the RX queue off the panda it fills up (on my vehicle in about 2 seconds) and once it is full messages are dropped.  Therefore, if you start reading messages only after you transmit a message you are pretty much guaranteed to miss the response because it got dropped.

On top of that, I found that if there wasn't a thread reading messages as fast as possible sometimes a read command would time out because it took to long to get processed through all the code.

I spent a lot of time iterating on making the RX processing more performant until I finally reached a point where HDS worked reliably.  I did most of this work while using interrupt based communication (because that is how it was originally configured) but then ultimately ended up switching to bulk mode, so it is possible I went further than I needed to.

I was trying to retain the original architecture as much as possible while dramatically increasing RX processing performance.  I will let others decide if a more significant reactor is in order.

drivers/windows/pandaJ2534DLL/J2534Frame.h
```diff
@@ -10,7 +10,7 @@ class J2534Frame {
 
 	J2534Frame(const panda::PANDA_CAN_MSG& msg_in) {
 		ProtocolID = CAN;
-		ExtraDataIndex = 0;
+		ExtraDataIndex = msg_in.len + 4;
```
gregjhogan: This was changed to match how tactrix operates

drivers/windows/pandaJ2534DLL/MessageTx_ISO15765.cpp
```diff
@@ -95,7 +95,7 @@ BOOL MessageTx_ISO15765::checkTxReceipt(J2534Frame frame) {
 
 				J2534Frame outframe(ISO15765);
 				outframe.Timestamp = frame.Timestamp;
-				outframe.RxStatus = TX_INDICATION | (flags & (ISO15765_ADDR_TYPE | CAN_29BIT_ID));
+				outframe.RxStatus = TX_MSG_TYPE | TX_INDICATION | (flags & (ISO15765_ADDR_TYPE | CAN_29BIT_ID));
```
gregjhogan: This was changed to match how tactrix operates

drivers/windows/pandaJ2534DLL/PandaJ2534Device.cpp
```diff
@@ -8,11 +8,15 @@ PandaJ2534Device::PandaJ2534Device(std::unique_ptr<panda::Panda> new_panda) : tx
 	this->panda->set_esp_power(FALSE);
 	this->panda->set_safety_mode(panda::SAFETY_ALLOUTPUT);
 	this->panda->set_can_loopback(FALSE);
-	this->panda->set_alt_setting(1);
+	this->panda->set_alt_setting(0);
```
gregjhogan: Couldn't get enough throughput in interrupt mode (full speed USB max is 64 bytes per Ms)

drivers/windows/pandaJ2534DLL/PandaJ2534Device.cpp
```diff
@@ -67,11 +74,28 @@ DWORD PandaJ2534Device::addChannel(std::shared_ptr<J2534Connection>& conn, unsig
 }
 
 DWORD PandaJ2534Device::can_recv_thread() {
```
gregjhogan: Separated reading USB data from message processing to prevent data reading and message processing from getting behind

geohot:
Nice, we'll test this on TIS on Monday

gregjhogan:
The old zip file only worked if you had visual studio installed, and it was missing a step about adding the DLL directory to your PATH environment variable.

Here is an updated zip with DLLs such that visual studio doesn't need to be installed, and with improved instructions:
[panda-j2534.zip](https://github.com/commaai/panda/files/1692262/panda-j2534.zip)

geohot:
Why do we need to add a DLL to PATH? Can we put this in the setup?

gregjhogan:
The registry points to this DLL, so it doesn't need to be in the path:
pandaJ2534_0404_32.dll

and that DLL loads this DLL:
panda.dll

An easy way to accomplish that is to put it in your PATH env var.
The installer (which I haven't figured out how to build) puts it in the system32 folder to accomplish this.

gregjhogan:
I was hoping there is an easy way to combine the two DLLs into one DLL, but I haven't found an easy way to do that yet.

gregjhogan:
Figured out how to build the installer, here is a new package:
[panda J2534 driver install.zip](https://github.com/commaai/panda/files/1737763/panda.J2534.driver.install.zip)
(requires panda firmware changes in #87)

Also made the J2534 DLL standalone (for convenience - no need to modify PATH)

gregjhogan:
New installer - this should now work with techstream!

[panda J2534 driver install.zip](https://github.com/commaai/panda/files/1742802/panda.J2534.driver.install.zip)

PR #87, opened 2018-02-18
fix sending WinUSB Extended Properties Feature Descriptor

gregjhogan:
`USB_WritePacket` couldn't handle the `len` parameter being greater than 127 and the WinUSB Extended Properties Feature Descriptor length is 142.

The way I read the reference manual, if you are not using interrupts you need to check `DTXFSTS` for room in the TX FIFO and wait for available space, so I added that in addition to a loop to send all the packets one at a time.

### DIFF
--- a/board/drivers/usb.h
+++ b/board/drivers/usb.h
@@ -72,7 +72,7 @@ uint8_t resp[MAX_RESP_LEN];
 #define ENDPOINT_TYPE_INT 3
 
 // This is an arbitrary value used in bRequest
-#define  MS_VENDOR_CODE 0xFF
+#define  MS_VENDOR_CODE 0x20
 
 //Convert machine byte order to USB byte order
 #define TOUSBORDER(num)\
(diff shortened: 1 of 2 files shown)


board/drivers/usb.h
```diff
@@ -231,25 +224,45 @@ void *USB_ReadPacket(void *dest, uint16_t len) {
...
+    // limit size to single packet
+    len = MAX_RESP_LEN;
+  }
 
-  // bullshit
-  USBx_INEP(ep)->DIEPTSIZ = ((numpacket << 19) & USB_OTG_DIEPTSIZ_PKTCNT) |
-                            (len               & USB_OTG_DIEPTSIZ_XFRSIZ);
+  if (len != 0) {
+    // wait for space in in the TX FIFO
+    while((USBx_INEP(0)->DTXFSTS & USB_OTG_DTXFSTS_INEPTFSAV) < MAX_RESP_LEN);
```
geohot: This shouldn't be done like this. This should be done on interrupt.

gregjhogan:
updated to use interrupts instead of polling `DTXFSTS`

PR #89, opened 2018-02-23
Chevy Volt safety

vntarasov:
Resets controls_allowed not only gas and brake, but on pressing regen paddle as well.
Radar configuration, dashboard status, and chime messages are always allowed through.


geohot:
Nice! Looks quite thorough.

PR #90, opened 2018-02-28
Honda Bosch Safety Hooks

gregjhogan:
Questions:
* CAN forwarding solution acceptable?
* share file for nidec and bosch safety hooks?
* identify interceptor pedal vs bosch hardware for addr 0x201 using message length?

TODO:
* [x] cancel on brakes (need to find new signal since nidec brake signal is used by bosch ACC)
* [x] cancel on gas pedal (since we use stock ACC this would only cancel lateral if we add it)
* [x] update pedal interceptor check (addr 0x201 is used by bosch hardware)
* [ ] test on nidec hardware (make sure we didn't break anything)
* [ ] test on vehicle with interceptor pedal (make sure we didn't break anything)


geohot:
Looks good, merging!

PR #96, opened 2018-03-05
Bounty: panda high quality CAN autobaud

gregjhogan:
Currently this attempts to initialize at 500, 250, 125 kbps.  Should any other speeds be included?

### DIFF
--- a/board/main.c
+++ b/board/main.c
@@ -303,6 +303,7 @@ int usb_cb_control_msg(USB_Setup_TypeDef *setup, uint8_t *resp, int hardwired) {
     // **** 0xde: set can bitrate
     case 0xde:
       if (setup->b.wValue.w < BUS_MAX) {
+        can_autobaud_enabled[setup->b.wValue.w] = false;
         can_speed[setup->b.wValue.w] = setup->b.wIndex.w;
         can_init(CAN_NUM_FROM_BUS_NUM(setup->b.wValue.w));
       }
@@ -551,6 +552,11 @@ int main() {
   puts("**** INTERRUPTS ON ****\n");
 
   __enable_irq();
+
+  // if the error interrupt is enabled to quickly when the CAN bus is active
+  // something bad happens and you can't connect to the device over USB
+  delay(10000000);
+  CAN1->IER |= CAN_IER_ERRIE | CAN_IER_LECIE;
 
   // LED should keep on blinking all the time
   uint64_t cnt = 0;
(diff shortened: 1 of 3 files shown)


cbenhagen:
VW is using 500 and 100 kbps. So adding 100 kbps would be nice.

atomic991:
Accord EU 2009 camera CAN uses 1000kbps so I think that one should be added as well

geohot:
I don't think this works. can_init_autobaud_all is only called once

Ideally, we want to measure the baud.

gregjhogan:
@geohot it definitely works! Hook two pandas together to try it out.

`can_init_autobaud_all` is called repeatedly in [this](https://github.com/gregjhogan/panda/blob/29d2d7692f583ddd28eaf3f06d0d7206e4c0494f/board/main.c#L565) loop.

The way the code is now it only auto-bauds once per power cycle/reset, but that can be changed if desired.

This works because a CAN bus automatically goes into passive mode if 127 bad messages occur.  A register holds the status indicating it went into passive mode due to to many message errors.  What I found is that being at the wrong bus speed causes this to happen, and it automatically clears the error register and leaves passive mode once good messages start flowing (once you have the correct bus speed).

Note that this solution only works on an active bus (it is reliant on messages quickly causing it to enter/leave silent mode).

geohot:
Oh I see. Cute!

We probably want to rethink how this interacts with silent mode. And I think a speed list is okay, but it should be more complete.

gregjhogan:
@geohot updated it to switch frequencies using interrupt that triggers on every invalid frame detected.  Still a few issues to work through, but it flips through the list of frequencies and locks in wicked fast now!

gregjhogan:
This is working now.

Questions:
* Is the list of bus speeds complete? 1000kbps, 500kbps, 250kbps, 125kbps, 100kbps
* is the way I prevent TX until bus speed is detected acceptable?
* is the delay before enabling the CAN error interrupt acceptable? (see comment below)

board/main.c
```diff
@@ -552,6 +553,11 @@ int main() {
 
   __enable_irq();
 
+  // if the error interrupt is enabled to quickly when the CAN bus is active
+  // something bad happens and you can't connect to the device over USB
+  delay(10000000);
+  CAN1->IER |= CAN_IER_ERRIE | CAN_IER_LECIE;
```
gregjhogan: Without this delay, if you hook a panda into an active can bus (which causes `can_init` to loop on setting the speed as opposed to the error interrupt triggering to iterate on setting the speed) you can't connect to it over USB (even though the lights are all blinking as if it detected the speed and is receiving messages)

chrisinajar:
I rebased this pull request onto master and added a tiny bit to disable auto-baud when entering a safety mode activates a bus. This allows the change to also work on a silent bus, which can happen with certain cars (and you use ELM to talk to them) or also in the testing environment with two panda's talking directly to each other.

tACK :+1:

Issue #103, opened 2018-03-27
1.0.8+ appears to break chffr


energee:
It looks like when we flash 1.0.8+ manually, it breaks wifi. It results in panda-xxxxx-BROKEN-pair SSID which blocks chffr pairing and probably wifi connections. Latest packaged release is OK.

adamurban:
This appears to be due to this commit to enable compiler optimization:

https://github.com/commaai/panda/commit/786a004af51e1e54713a03f5c311b63a3703d48c

Setting the CFLAGS back to -O0 allows for a successful flash and a non-broken SSID, and also allows cffr to function again...

PR #107, opened 2018-04-06
Implement WebUSB and upgrade WinUSB to 2.0

chrisinajar:
Closes: #103 

Add a Binary Object Store (BOS) descriptor to the USB driver code. Adds WebUSB descriptors and also WinUSB 2.0 (we currently only support 1.0, which is dominantly used on Windows XP).

I left lots of comments as to where I got most of the information / constants / spec descriptions.

* [x] works on osx
* [x] works on eon
* [x] works on pandajs
* [x] works on pandajs over wifi
* [x] works on windows
* [x] un-breaks SPI communications


chrisinajar:
Also exposes string configuration name, upgrades USB support value from 2.0 to 2.1. Also changes the name from "panda" to "Panda" because it looks a lot better, we should make sure that change doesn't break anything.

Tested and working on white panda, haven't tested on grey yet.
image

chrisinajar:
I believe this is ready now.

lomal788:
i have a same issues if i firmware upgrade my panda and use zadig to install web-usb and it will work fine?
is this version released on 1.5.2?

https://github.com/commaai/panda-artifacts

robbederks:
@lomal788 This has been implemented for a very long time, so yes, 1.5.2 should have it!

PR #109, opened 2018-04-09
Angle rate2

rbiasini:


board/safety/safety_toyota.h
```diff
@@ -20,6 +10,22 @@ const int32_t MAX_RATE_UP = 10;        // ramp up slow
 const int32_t MAX_RATE_DOWN = 25;      // ramp down fast
 const int32_t MAX_TORQUE_ERROR = 350;  // max torque cmd in excess of torque motor
 
+struct Lookup {
```
geohot: Rename struct "lookup_t"

Change all "double" to "float"

board/safety/safety_toyota.h
```diff
@@ -29,14 +35,59 @@ const int32_t RT_INTERVAL = 250000;    // 250ms between real time checks
 const int16_t MAX_ACCEL = 1500;        // 1.5 m/s2
 const int16_t MIN_ACCEL = -3000;       // 3.0 m/s2
 
+int cruise_engaged_last = 0;           // cruise state
+int ipas_state = 1;                    // 1 disabled, 3 executing angle control, 5 override
+int angle_control = 0;                 // 1 if direct angle control packets are seen
+double speed = 0.;
+
+// track the torque measured for limiting
+int16_t torque_meas[3] = {0, 0, 0};    // last 3 motor torques produced by the eps
```
geohot: This should be a struct, and a function should update it. No repeating code three times

board/safety/safety_toyota.h
```diff
@@ -29,14 +35,59 @@ const int32_t RT_INTERVAL = 250000;    // 250ms between real time checks
...
+// interp function that holds extreme values
+double interpolate(struct Lookup xy, double x) {
+  int size = sizeof(xy.x) / sizeof(xy.x[0]);
+  if (x <= xy.x[0]) {
+    return xy.y[0];
+
+  } else {
+    for (int i=0; i < size-1; i++) {
+      if (x < xy.x[i+1]) {
+        return (xy.y[i+1] - xy.y[i]) * (x - xy.x[i]) / (xy.x[i+1] - xy.x[i]) + xy.y[i];
```
geohot: This is unreadable, and it's not immediately obvious why it won't divide by 0 (I don't think it will, but that should be so obvious)

PR #110, opened 2018-04-11
Add safety hook for ignition state for car platforms that don't support GPIO pin

Jamezz:
The main contender of this change is for the GM Volt port.

* Add safety hook for ignition state. All interfaces except GM should default to using the GPIO pin state for determining ignition.

* Override the safety hook created for ignition on GM safety to key the powertrain gear selection to ignition. This is needed because the Volt is incredibly cryptic about vehicle power status (can't exactly use engine parameters from ELM327). We know about the gear selection, so any time we're out of Park we consider the ignition to be "on."

Tested on Volt 2017 Premier drive: https://community.comma.ai/cabana/?route=f89c604cf653e2bf|2018-04-11--13-18-55


geohot:
I support this idea, but a potential issue is controls doesn't fingerprint until the car starts. We'd have to rethink that as well.

Jamezz:
That is potentially an issue with other cars, but the Volt (hopefully most GM) sends these messages immediately on power on and very often during operation. And changing the gear state should reasonably require the vehicle to be powered on.

I suspect other cars may need to use this technique, but hopefully they can find a nice way like the Volt has to ensure the proper order of operations.

geohot:
Upstreamed, there's still an issue with controls starting, but I think I have a way to fix it.

PR #111, opened 2018-04-18
Addition to Bosch safety to support Hatchback

energee:
The computer brake is included on 0x17c brake sensor on Accord and CR-V. We assumed all Bosch Hondas had a new 0x1be message which reports only manual brake, but Civic Hatchback is not like this- It doesn't have this message and 0x17c works like the other Hondas so passing a parameter from Openpilot for this.

### DIFF
--- a/board/safety/safety_honda.h
+++ b/board/safety/safety_honda.h
@@ -15,7 +15,7 @@ int gas_prev = 0;
 int gas_interceptor_prev = 0;
 int ego_speed = 0;
 // TODO: auto-detect bosch hardware based on CAN messages?
-bool bosch_hardware = false;
+bool alternate_brake = false;
 
 static void honda_rx_hook(CAN_FIFOMailBox_TypeDef *to_push) {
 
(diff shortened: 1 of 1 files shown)


geohot:
Can we just always check both?

geohot:
Oh, I think I understand. Can you write a comment explaining this, then we'll upstream.

rbiasini:
this change is going to conflict with https://github.com/commaai/panda/pull/113. 

Actually, I can close my PR if you can include it into yours (and add a regression test too?)

energee:
Includes https://github.com/commaai/panda/pull/113

energee:
Should be good to go

tests/safety/test_honda.py
```diff
@@ -94,6 +101,12 @@ def test_disengage_on_brake(self):
     self.safety.honda_rx_hook(self._brake_msg(1))
     self.assertFalse(self.safety.get_controls_allowed())
 
+  def test_alt_disengage_on_brake(self):
+    self.safety.set_alt_brake_signal(1)
+    self.safety.set_controls_allowed(1)
+    self.safety.honda_rx_hook(self._alt_brake_msg(1))
+    self.assertFalse(self.safety.get_controls_allowed())
```
rbiasini: I would add one last test:

```
self.safety.set_alt_brake_signal(0)
self.safety.set_controls_allowed(1)
self.safety.honda_rx_hook(self._alt_brake_msg(1))
```
Other than that looks good!

energee:
Looks like this needs more work

board/safety/safety_honda.h
```diff
@@ -115,6 +119,12 @@ static int honda_tx_hook(CAN_FIFOMailBox_TypeDef *to_send) {
       if ((to_send->RDLR & 0xFFFF0000) != to_send->RDLR) return 0;
     }
   }
+  
+  // FORCE CANCEL: safety check only relevant when spamming the cancel button in Bosch HW
+  // ensuring that only the cancel button press is sent (VAL 2)
+  if (((to_send->RIR>>21) == 0x296) && bosch_hardware) {
```
rbiasini: && !controls_allowed
we want to spam the resume button when acc is on

PR #113, opened 2018-04-24
Honda Bosch: enforcing safety check on cancel spam

rbiasini:
@energee @gregjhogan This PR to ensure that when spamming cancel button we are not engaging/resuming by mistake.

Need to add some safety regression test.


rbiasini:
included in https://github.com/commaai/panda/pull/111

PR #117, opened 2018-05-22
Update gas_interceptor_threshold due to experiments wth RDX and Odyssey

joeljacobs:
This new threshold value was arrived at through experimentation with the RDX and Odyssey. Values much lower than this would cause OP to suddenly disengage due to anomolies with the Gas Pedal analog readings.

### DIFF
--- a/board/safety/safety_honda.h
+++ b/board/safety/safety_honda.h
@@ -8,7 +8,7 @@
 //      brake > 0mph
 
 // these are set in the Honda safety hooks...this is the wrong place
-const int gas_interceptor_threshold = 328;
+const int gas_interceptor_threshold = 800;
 int gas_interceptor_detected = 0;
 int brake_prev = 0;
 int gas_prev = 0;


joeljacobs:
This PR goes along with [My Pedal OP PR](https://github.com/commaai/openpilot/pull/251)

board/safety/safety_honda.h
```diff
@@ -8,7 +8,7 @@
 //      brake > 0mph
 
 // these are set in the Honda safety hooks...this is the wrong place
-const int gas_interceptor_threshold = 328;
+const int gas_interceptor_threshold = 800;
```
rbiasini: As we were discussing on the other PR, this should be 120/0.2539 (+1 to be slightly conservative and not false trigger panda cancellations before openpilot does) in order to be consistent with opendbc and openpilot. If this threshold does not work, we should figure out why.

joeljacobs: I know that 600 does not work, so 472 would definitely not work.  If I set it higher, occasional momentary gas signal spikes (either a-to-d or physical pedal anomaly) will cause OP to throw a "controls mismatch" error.

joeljacobs: I should note that it does a good job knowing it should disengage when I press on the gas with these settings.

rbiasini: "I don't know why it seems the dbc is more forgiving." <-- yeah... this needs to be figured out before we can merge changes.

joeljacobs: Could it have anything to do with the GAS2 value? If you divide 120 by 0.126992032, you get a much closer value of 944.94.

PR #120, opened 2018-06-03
Change Volt ignition detection to real ignition

Jamezz:
Parsing the 0x1F1 message has the vehicle ignition status in Bit 5. This should improve the ignition experience.

(Sorry about the VERSION)


VERSION
```diff
@@ -1 +1 @@
-v1.1.1
\ No newline at end of file
+v1.1.1
```
rbiasini: can't leave white space at the end. I use the `echo -n` command

rbiasini: ah, realized you have been fighting it for a while :)

rbiasini:
Done here: https://github.com/commaai/panda/commit/293fd1ac70250a09b511cc8af99aa7f260d586ef

PR #121, opened 2018-06-08
Support for Classic Tesla Model S

zax123:
Contains RX and TX safety hooks for Classic (non-AP) Tesla Model S

Also contains forwarding code to bypass the EPAS kill message sent from the gateway.

Requires the harness as designed by @kalud and @zax123 and the Tesla giraffe.

### DIFF
--- a/board/safety/safety_tesla.h
+++ b/board/safety/safety_tesla.h
@@ -1,0 +1,159 @@
+// board enforces
+//   in-state
+//      accel set/resume
+//   out-state
+//      cancel button
+//      regen paddle
+//      accel rising edge
+//      brake rising edge
+//      brake > 0mph
+
+// lateral limits
+const int16_t MAX_ANGLE = 20; //Degrees
+
+int tesla_brake_prev = 0;
+int tesla_gas_prev = 0;
+int tesla_speed = 0;
+int current_car_time = -1;
+int time_at_last_stalk_pull = -1;
+int eac_status = 0;
+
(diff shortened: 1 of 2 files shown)


jeankalud:
We want and will add those safety in Panda code.

- EPS internal firmware already limits angle rate quite and its speed dependant
- Absolute max angle is limited in EON's code, its not fixed 20 degreee but a function of speed:

https://github.com/jeankalud/openpilot/blob/tesla/selfdrive/car/tesla/carcontroller.py#L119-L129

But now that we have some more people helping us, the progress is going well and we will support the code without issues.

board/safety/safety_tesla.h
```diff
@@ -0,0 +1,159 @@
...
+
+// lateral limits
+const int16_t MAX_ANGLE = 20; //Degrees
+
+int tesla_brake_prev = 0;
+int tesla_gas_prev = 0;
+int tesla_speed = 0;
+int current_car_time = -1;
+int time_at_last_stalk_pull = -1;
+int eac_status = 0;
```
rbiasini: can you make all those shared variables starting with `tesla_`?

zax123:
Hi @rbiasini,

Absolutely, I will rename those shared variables.  We are working on solidifying all the safety parameters on Thursday evening and will send new cleaned-up code on Friday.

Thanks for looking into this!

rbiasini:
superseded by https://github.com/commaai/panda/pull/132

PR #124, opened 2018-06-15
Strengthen Tesla VIN detection

AllWashedOut:
Replaces a counter with a check of the dictionary contents to work around a potential bug: if the same VIN piece were received twice, it would still have incremented the counter.


adamurban:
This is a good change... I also propose you fix my typos in lines 32 and 36 before merging...

Specifically, ```bus_num``` should be changed to ```body_ bus_num```

appleguru:
This is a good change... I also propose you fix my typos in lines 32 and 36 before merging...

Specifically, ```bus_num``` should be changed to ```body_bus_num```

AllWashedOut:
Done, corrected 'bus_num' to 'body_bus_num'

PR #128, opened 2018-08-01
Add an alternate function to use GMLAN output on panda as a GPIO

appleguru:
This PR adds an alternate mode for the GMLAN output on the panda where it can be used as a general purpose output (set high or low). To accomplish this, a timer forces the GMLAN output low if every 15ms if the desired output is high to bypass the GMLAN transceiver hardware's 17ms timeout

```
gmlan_switch_init(0); //0 = timeout disabled; 1 = timeout enabled.
// With timeout enabled you need to call reset_gmlan_switch_timeout()
// faster than every 1s or the GPIO mode will be disabled

set_gmlan_digital_output(GMLAN_HIGH); //set output high
set_gmlan_digital_output(GMLAN_LOW); //set output low
```

### DIFF
--- a/board/main.c
+++ b/board/main.c
@@ -4,6 +4,7 @@
 // ********************* includes *********************
 
 #include "libc.h"
+#include "safety.h"
 #include "provision.h"
 
 #include "drivers/drivers.h"
@@ -14,12 +15,11 @@
 #include "drivers/uart.h"
 #include "drivers/adc.h"
 #include "drivers/usb.h"
+#include "drivers/gmlan_alt.h"
+#include "drivers/can.h"
 #include "drivers/spi.h"
 #include "drivers/timer.h"
 
-#include "drivers/gmlan_alt.h"
-#include "safety.h"
-#include "drivers/can.h"
 
 // ***************************** fan *****************************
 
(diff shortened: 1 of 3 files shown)


jeankalud:
I support this, works very well. All Tesla port users are using it.

geohot:
This is hard to review, since the file was moved without git mv. Can you clean up the diff so I can review it?

appleguru:
Rebased.

main.c includes were reordered because:
1) we use gmlan_alt.h in safety (on the Tesla fork), so needs to be included before safety.h
2) can.h requires gmlan_alt for bitbang (so gmlan_alt needs to be before can)
3) gmlan_alt requires drivers.h (set_gpio_mode(), so needs to be after drivers.

I am open to alternative suggestions for how to best handle.

board/main.c
```diff
@@ -4,7 +4,6 @@
 // ********************* includes *********************
 
 #include "libc.h"
-#include "safety.h"
```
geohot: Why can't this stay here?

geohot:
Fix the include order for now to match master (put gmlan_alt.h before can.h) and I'll merge. We'll deal with the Tesla safety merge later.

appleguru:
Done. @jeankalud won't be happy ;-)

PR #129, opened 2018-08-02
Kia and Hyundai Safety

nullku7:
This compiles, and flashes fine on the panda.
OP operates, and the car continues to steer.

No further testing performed.
Unit Tests to follow.


board/safety/safety_hyundai.h
```diff
@@ -0,0 +1,112 @@
...
+      violation |= rt_rate_limit_check(desired_torque, hyundai_rt_torque_last, HYUNDAI_MAX_RT_DELTA);
+
+      // every RT_INTERVAL set the new limits
+      uint32_t ts_elapsed = get_ts_elapsed(ts, hyundai_ts_last);
+      if (ts_elapsed > HYUNDAI_RT_INTERVAL) {
+        hyundai_rt_torque_last = desired_torque;
+        hyundai_ts_last = ts;
+      }
+    }
+
```
rbiasini: we need something like:
```
    // no torque if controls is not allowed
    if (!current_controls_allowed && (desired_torque != 0)) {
      violation = 1;
    }

    // reset to 0 if either controls is not allowed or there's a violation
    if (violation || !current_controls_allowed) {
      hyundai_desired_torque_last = 0;
      hyundai_rt_torque_last = 0;
      hyundai_ts_last = ts;
    }
```

board/safety/safety_hyundai.h
```diff
@@ -0,0 +1,112 @@
...
+    if (violation) {
+      return false;
+    }
+  }
+
+  return true;
+}
+
+static void hyundai_rx_hook(CAN_FIFOMailBox_TypeDef *to_push) {
```
rbiasini: this should be pretty much like gm_rx_hook.

board/safety/safety_hyundai.h
emmertex: Line 76

rbiasini: ops :)

board/safety/safety_hyundai.h
rbiasini: yeah, it's fine. then you should use the cruise state in line 102 (safety_toyota does the same).
if acc is enabled <==> controls_allowed.

rbiasini:
superseded by https://github.com/commaai/panda/commit/8a1dcbec54e14033aad3bfdb07d380dd49640b9e
Thanks though, it was the right start.

PR #130, opened 2018-08-02
Chrysler safety controls

adhintz:
Tested on Pacifica 2017 Hybrid and Pacifica 2018 non-Hybrid.


adhintz:
This is ready for a review.

board/safety/safety_chrysler.h
```diff
@@ -0,0 +1,129 @@
...
+    // Extended
+    // Not looked at, but have to be separated
+    // to avoid address collision
+    addr = to_push->RIR >> 3;
+  } else {
+    // Normal
+    addr = to_push->RIR >> 21;
+  }
+
+  // Gear selector used for determining ignition.
```
rbiasini: instead of gear I would use any CAN message that represents "car on"

board/safety/safety_chrysler.h
```diff
@@ -0,0 +1,129 @@
...
+    // Normal
+    addr = to_send->RIR >> 21;
+  }
+
+  // LKA STEER: Too large of values are ignored by the car, but still filter it out.
+  if (addr == 0x292) {
+    int rdlr = to_send->RDLR;
+    int steer = ((rdlr & 0x7) << 8) + ((rdlr & 0xFF00) >> 8);
+    int straight = 1024;
+    int max_steer = 250;
```
rbiasini: so basically the car would not respond (fault?) if the sent value exceed 1024 +- 250?

tests/safety/test_chrysler.py
```diff
@@ -0,0 +1,67 @@
...
+
+  def test_steer_calculation(self):
+    self.assertEqual(0x14, self._steer_msg(0)[0].RDLR)  # straight, no steering
+
+  def test_steer_tx(self):
+    self.safety.set_controls_allowed(1)
+    self.assertTrue(self.safety.chrysler_tx_hook(self._steer_msg(0)))
+    self.assertTrue(self.safety.chrysler_tx_hook(self._steer_msg(100)))
+    self.assertTrue(self.safety.chrysler_tx_hook(self._steer_msg(-100)))
+    self.assertFalse(self.safety.chrysler_tx_hook(self._steer_msg(300)))
```
rbiasini: I would test 250 and 251, which are right across the ok vs non-ok boundary

rbiasini:
@adhintz overall looks good! few comments. 
What's left after are subjective things like max level of torque allowed and how much effort the driver needs to do to override the max steer torque. Another way to do safety is to look at `safety_hyundai.h`, for example, and copy the same structure (just different params and CAN details).

tests/safety/test_chrysler.py
adhintz: Good idea -- done.

board/safety/safety_chrysler.h
adhintz: If the sent value exceeds 1024 +- 250 then the steering wheel will no longer actuate from the LKAS messages. The driver can still steer normally. I guess that means the ECU does a silent fault.

board/safety/safety_chrysler.h
adhintz: I actually switched it to use the default_ign_hook. The chrysler giraffe only has power if the car's ignition is on, and I've wired that 12V to the IGN GPIO on panda.

adhintz:
Thanks for the review! I haven't yet tested this new code on a real car, but I think I've addressed your comments.

As for the max torque, the current limits should be totally fine. If anything, I'd like it to have more torque than the ECU allows so it can handle sharper turns.

I'll let you know once I've actually tried this new version on a car.

python/__init__.py
```diff
@@ -106,6 +106,7 @@ class Panda(object):
   SAFETY_HONDA = 1
   SAFETY_TOYOTA = 2
   SAFETY_HONDA_BOSCH = 4
+  SAFETY_CHRYSLER = 8
```
rbiasini: the python code is kind of deprecated, but for consistency I would set this to 9

rbiasini:
Ok, let me know how the test goes. And thank for writing the test part as well.

adhintz:
Testing worked well this weekend in my car. The PR is ready for review.

board/safety/safety_chrysler.h
```diff
@@ -0,0 +1,136 @@
...
+    if (((to_push->RDLR & 0x380000) >> 19) == 7) {
+      controls_allowed = 1;
+    } else {
+      controls_allowed = 0;
+    }
+  }
+
+  // exit controls on brake press by human
+  if (addr == 0x140) {
+    if (to_push->RDLR & 0x4) {
```
rbiasini: can we exit on gas too?

board/safety/safety_chrysler.h
```diff
@@ -0,0 +1,136 @@
...
+    }
+    if (steer < -max_steer) {
+      return false;
+    }
+    if (!controls_allowed && steer != 0) {
+      // If controls not allowed, only allow steering to move closer to 0.
+      if (chrysler_desired_torque_last == 0) {
+	return false;
+      }
+      if ((chrysler_desired_torque_last > 0) && (steer >= chrysler_desired_torque_last)) {
```
rbiasini: so it can still take 2.3s to ramp down (assuming you send steer torque at 100Hz). Why not enforcing 0 torque when `!controls_allowed`?

board/safety/safety_chrysler.h
adhintz: The stock ACC does not exit on a gas press. The gas pedal is also used to resume ACC when at 0mph for longer than 3 seconds.

If you feel strongly that gas should exit, then I could try to add it. I'd also need to add logic in carstate.py so that OpenPilot doesn't immediately try to turn back on because ACC will still be green:
self.main_on = cp.vl["ACC_2"]['ACC_STATUS_2'] == 7  # ACC is green.
Perhaps add a disabled state that's set when gas is pressed. Then change main_on to be (not disabled) and (acc is green). Disabled would be set back to false when ACC is not green.

rbiasini: basically you would need to find a way to disable ACC as well on gas press (that's what we do on Hyundai and Toyota with DSU plugged in). It's fine for now, will be required in order to list this port more officially.

board/safety/safety_chrysler.h
adhintz: The steering actuator ECU silently faults if the there's a change of more than 100 between two messages. I know that from testing on the ramp-up, but have probably not verified that on the ramp-down.  Perhaps it allows faster ramp-down.

The stock ramp-down is -3 per message, which is what I have OpenPilot. Enforcing that would be easy and reduce the ramp-down time to 0.77s. I could easily change it to -20 and it works fine, which would be a ramp-down time of 0.12s. What do you think is a good ramp-down time?
```
if ((chrysler_desired_torque_last > 0) && (steer > (chrysler_desired_torque_last - 20))) {
  return false;
```

rbiasini: If a quick ramp-down does not cause a eps fault (I would be surprised it it does), then I would just enforce 0 torque whenever controls aren't allowed (this is also for simplicity). Otherwise <0.5s rampdown seems legit.

board/safety/safety_chrysler.h
adhintz: Makes sense. I can try to send the ACC cancel button message. It has a counter, so hopefully it doesn't confuse the ECU that typically sends it. I'll add that to my TODO list.

PR #132, opened 2018-08-14
Added Tesla safety changes.

zax123:
Strictly the safety changes to panda code to add support for pre-autopilot Tesla Model S with the EPAS harness and giraffe designed by @Kalud and @zax123 installed.


board/safety/safety_tesla.h
```diff
@@ -0,0 +1,350 @@
...
+int tesla_gas_prev = 0;
+int tesla_speed = 0;
+int current_car_time = -1;
+int time_at_last_stalk_pull = -1;
+int eac_status = 0;
+
+int tesla_ignition_started = 0;
+
+// interp function that holds extreme values
+float tesla_interpolate(struct lookup_t xy, float x)
```
rbiasini: just moved a generic interpolation function to safety.h I recommend using that one.

zax123:
OK, I removed the function from safety_tesla.h and I'm using the generic interpolate function in safety.h now, thanks!

board/safety.h
```diff
@@ -49,6 +49,7 @@ int controls_allowed = 0;
 #include "safety/safety_gm.h"
 #include "safety/safety_ford.h"
 #include "safety/safety_cadillac.h"
+#include "safety/safety_tesla.h"
```
rbiasini: move this inside `#ifdef PANDA`: legacy boards and pedal don't support floats.

board/safety.h
```diff
@@ -100,6 +102,7 @@ const safety_hook_config safety_hook_registry[] = {
   {SAFETY_GM, &gm_hooks},
   {SAFETY_FORD, &ford_hooks},
   {SAFETY_CADILLAC, &cadillac_hooks},
+  {SAFETY_TESLA, &tesla_hooks},
```
rbiasini: move this inside `#ifdef PANDA`: legacy boards and pedal don't support floats.

zax123:
Done, good eye for finding that!

board/safety.h
```diff
@@ -49,6 +49,9 @@ int controls_allowed = 0;
 #include "safety/safety_gm.h"
 #include "safety/safety_ford.h"
 #include "safety/safety_cadillac.h"
+#ifdef PANDA
+#include "safety/safety_tesla.h"
```
rbiasini: move this after https://github.com/commaai/panda/pull/132/files#diff-7b552a9c774a81e90fcf2183ff134492R47

board/safety/safety_tesla.h
```diff
@@ -0,0 +1,319 @@
...
+    addr = to_push->RIR >> 21;
+  }
+
+  // Record the current car time in current_car_time (for use with double-pulling cruise stalk)
+  if (addr == 0x318)
+  {
+    int hour = (to_push->RDLR & 0x1F000000) >> 24;
+    int minute = (to_push->RDHR & 0x3F00) >> 8;
+    int second = (to_push->RDLR & 0x3F0000) >> 16;
+    current_car_time = (hour * 3600) + (minute * 60) + second;
```
rbiasini: using car time from CAN seems unnecessary complicated and potentially unreliable. I recommend re-using:
```
    uint32_t ts = TIM2->CNT;
    uint32_t ts_elapsed = get_ts_elapsed(ts, tesla_ts_angle_last);
```

board/safety/safety_tesla.h
```diff
@@ -0,0 +1,319 @@
...
+//      cancel button
+//      regen paddle
+//      accel rising edge
+//      brake rising edge
+//      brake > 0mph
+
+// 2m/s are added to be less restrictive
+const struct lookup_t TESLA_LOOKUP_ANGLE_RATE_UP = {
+    {2., 7., 17.},
+    {5., .8, .25}};
```
rbiasini: above 17mph, this is twice than what is in `toyota_safety_ipas.h`. Isn't it too much?

board/safety/safety_tesla.h
```diff
@@ -0,0 +1,319 @@
...
+
+  // exit controls on EPAS error
+  // EPAS_sysStatus::EPAS_eacStatus 0x370
+  if (addr == 0x370)
+  {
+    // if EPAS_eacStatus is not 1 or 2, disable control
+    eac_status = ((to_push->RDHR >> 21)) & 0x7;
+    // For human steering override we must not disable controls when eac_status == 0
+    // Additional safety: we could only allow eac_status == 0 when we have human steerign allowed
+    if ((controls_allowed == 1) && (eac_status != 0) && (eac_status != 1) && (eac_status != 2))
```
rbiasini: If we don't want to disable on user steer override (let's think about it), then I would remove this check. Does not seem safety relevant.

board/safety/safety_tesla.h
```diff
@@ -0,0 +1,319 @@
...
+    {
+      controls_allowed = 0;
+      puts("EPAS error! \n");
+    }
+  }
+  //get latest steering wheel angle
+  if (addr == 0x00E)
+  {
+    int angle_meas_now = (int)((((to_push->RDLR & 0x3F) << 8) + ((to_push->RDLR >> 8) & 0xFF)) * 0.1 - 819.2);
```
rbiasini: we should just use floats and define a new `sample_t` struct that contains floats

board/safety/safety_tesla.h
zax123: I got rid of this section of code.  It was going to be used if we only used double-pulls to activate openpilot.  We now are activating on a single pull.  Double-pull may activate other features, but safety-wise, we only need to monitor a single pull because nothing changes from a safety perspective when you pull twice.

board/safety.h
zax123: This is done and will be in my next commit.

board/safety/safety_tesla.h
rbiasini: cool

board/safety/safety_tesla.h
zax123: It looks like this code was copied from the Toyota IPAS safety code.  Did you want to make this generic somehow and move it into safety.h?

board/safety/safety_tesla.h
rbiasini: yeah, it would be safety relevant only if we want to disengage on user steer override.
I assume that if EAC is faulted, then it won't respond to steer commands.

board/safety/safety_tesla.h
```diff
@@ -0,0 +1,319 @@
...
+      return true;
+    }
+
+    if (controls_allowed)
+    {
+      if (steer_allowed)
+      {
+
+        // add 1 to not false trigger the violation
+        int delta_angle_up = (int)(interpolate(TESLA_LOOKUP_ANGLE_RATE_UP, tesla_speed) * 25. + 1.);
```
rbiasini: I think all of these should be floats

board/safety/safety_tesla.h
zax123: Yes, as soon as EAC faults, it disables steer commands and the user must take control.  However, there is no feedback to openpilot indicating this, so I'm thinking it would be good for us to set controls_allowed = 0 because then feedback gets back to OP and it can throw an alert on the screen and/or play an alert sound.  What do you think?

board/safety/safety_tesla.h
zax123: I'm sorry for my ignorance, I didn't copy this piece of code.  It looks like the difference is the .25 in Tesla, which is .15 in Toyota.  Is that the part that concerns you?

board/safety/safety_tesla.h
zax123: Yeah, so, we could change these to floats, but then the toyota_ipas safety code is different.  Should we make them the same?

board/safety/safety_tesla.h
rbiasini: yes

board/safety/safety_tesla.h
rbiasini: exactly. `safety_toyota_ipas.h` is a good starting point, but hasn't reached the level of safety we want. `safety_tesla.h` will be the first "approved" angle control safety logic.

board/safety/safety_tesla.h
rbiasini: yes, I would be happy to generalize those functions and use them in both places

zax123: If you would be so kind to do that, I would appreciate it, just because I didn't write that code myself and would likely introduce more bugs trying to fix it.

rbiasini: yes, once the new functions are defined, I'll take care of using them in safety_toyota_ipas

board/safety/safety_tesla.h
zax123: So we have a couple engineers and coders who tested and tweaked the values in this code and I believe that that value was necessary in order to take even some modest curves at a decent speed.  I don't think it's excessive.  There are still some curves that the model can't take which I believe it should.  I can definitely change this to .15 for upstreaming purposes, but I have a feeling users will likely change it back because of the turns it would disengage on.  Definitely your call, please let me know.

board/safety/safety_tesla.h
zax123: Oh, so you'd like the Tesla team to generalize them and put them in safety.h using floats, or were you going to do that?  Just want to be clear...

rbiasini: tesla team shoul'd do that :)

board/safety/safety_tesla.h
rbiasini: ok, let's keep these numerical details last. It comes down to testing the max steer injection case:
https://medium.com/@comma_ai/how-to-write-a-car-port-for-openpilot-7ce0785eda84

board/safety/safety_tesla.h
```diff
@@ -0,0 +1,305 @@
...
+        int delta_angle_down = (int)(interpolate(TESLA_LOOKUP_ANGLE_RATE_DOWN, tesla_speed) * 25. + 1.);
+        int highest_desired_angle = tesla_desired_angle_last + (tesla_desired_angle_last > 0 ? delta_angle_up : delta_angle_down);
+        int lowest_desired_angle = tesla_desired_angle_last - (tesla_desired_angle_last > 0 ? delta_angle_down : delta_angle_up);
+        int TESLA_MAX_ANGLE = (int)(interpolate(TESLA_LOOKUP_MAX_ANGLE, tesla_speed) + 1.);
```
rbiasini: we also need a simple check that ensures the rate limits are observed at each step (I only see the real time check). See safety_gm.h, for example.

board/safety/safety_tesla.h
```diff
@@ -0,0 +1,305 @@
...
+
+    if (st_enabled == 0) {
+      //steering is not enabled, do not check angles and do send
+      tesla_desired_angle_last = desired_angle;
+      return true;
+    }
+
+    if (controls_allowed)
+    {
+      if (steer_allowed)
```
rbiasini: why distinguishing between controls_allowed and steer_allowed? can we only use controls_allowed?

rbiasini: why need to have steer_allowed? can't we just use controls_allowed only?

board/safety/safety_tesla.h
```diff
@@ -0,0 +1,305 @@
...
+
+    // update array of samples
+    update_sample(&tesla_angle_meas, angle_meas_now);
+
+    // check for violation;
+    if (max_limit_check(angle_meas_now, highest_rt_angle, lowest_rt_angle))
+    {
+      // We should not be able to STEER under these conditions
+      // Other sending is fine (to allow human override)
+      steer_allowed = 0;
```
rbiasini: can't this be replaced by controls_allowed?

board/safety/safety_tesla.h
```diff
@@ -0,0 +1,305 @@
...
+          puts("Angle limit - max! \n");
+        }
+      }
+      else
+      {
+        violation = 1;
+        controls_allowed = 0;
+        puts("Steering commads disallowed");
+      }
+    }
```
rbiasini: what is blocking messages to be sent when `controls_allowed == 0`? Seems like this function would return True.

zax123:
Hi @rbiasini, we've made all the requested changes as per our conversations above.  Can you please check on the new code and let me know if anything else is missing?

Thanks!

rbiasini:
@zax123 , thanks. Will review within this week

board/safety/safety_tesla.h
```diff
@@ -0,0 +1,308 @@
...
+    float lowest_rt_angle = tesla_rt_angle_last - (tesla_rt_angle_last > 0 ? rt_delta_angle_down : rt_delta_angle_up);
+
+    if ((ts_elapsed > TESLA_RT_INTERVAL) || (controls_allowed && !tesla_controls_allowed_last))
+    {
+      tesla_rt_angle_last = angle_meas_now;
+      tesla_ts_angle_last = ts;
+    }
+
+    // update array of samples
+    update_fsample(&tesla_angle_meas, angle_meas_now);
```
rbiasini: tesla_angle_meas is unused

board/safety/safety_tesla.h
```diff
@@ -0,0 +1,308 @@
...
+      //steering is not enabled, do not check angles and do send
+      tesla_desired_angle_last = desired_angle;
+      return true;
+    }
+
+    if (controls_allowed)
+    {
+      // add 1 to not false trigger the violation
+      float delta_angle_up = interpolate(TESLA_LOOKUP_ANGLE_RATE_UP, tesla_speed) * 25. + 1.;
+      float delta_angle_down = interpolate(TESLA_LOOKUP_ANGLE_RATE_DOWN, tesla_speed) * 25. + 1.;
```
rbiasini: why the `25` factor? here the check should be done each step, right?

rbiasini:
@zax123 , there is a minor conflict and a bug in the angle commanded rate limiter check (see my comment). Other than that it looks good.
My only worry is that the RT measured steering angle check will false trigger. I don't have a good solution in mind for this, except that maybe you can do the RT rate limiter check on the commanded angle (but I realize that checking the measured angle is more meaningful...).

Anyway, Once the merge conflict + bug are fixed, we can merge this PR. After that, you guys should write a regression test (see `panda/tests/safety/`) and we'll add the Tesla to the list of community supported cars (assuming the official branch uses this safety code :)). Finally, bounty will be paid!
Good job Tesla guys!

zax123:
Hi @rbiasini ,

Thanks for keeping this alive.  I've made the necessary changes.  For some reason, there's still a merge conflict with the SAFETY_TESLA number.  It no longer conflicts with the SAFETY_HYUNDAI number.

The other changes are done.  I'd just like it if you could confirm this is all ok before we move on to making the unit test file.

Thank you!

Rob

zax123:
PS. Congratulations on your promotion!!

PR #134, opened 2018-09-05
Hyundai Safety Auto-Detect which CAN bus Camera is on

emmertex:


rbiasini:
temporary solution to support legacy giraffe until official hyundai/kia giraffe is out

PR #136, opened 2018-09-06
Increase Volt's steering limit

vntarasov:


vntarasov:
Kudos to @TheMutley for pointing out the limit can be increased without causing SCM faults.

board/safety/safety_gm.h
```diff
@@ -8,7 +8,7 @@
 //      brake rising edge
 //      brake > 0mph
 
-const int GM_MAX_STEER = 255;
+const int GM_MAX_STEER = 300;
```
rbiasini: `panda/tests/safety/test_gm.py` also need to be updated

rbiasini:
fixed in https://github.com/commaai/panda/commit/627713948a6941b95d013d1fd2b08ab524598074

PR #138, opened 2018-10-25
XACC_ACT signal detection instead of CRUISE_STATE for engaging controls

wocsor:
This signal works on Corolla, Camry, and Lexus GS 300h so far. CRUISE_STATE is always 0000 for GS 300h so OP cannot engage


rbiasini:
already merged

energee:
@wocsor For changes like this, you would want to add a safety unit test or modify the current one. You can see this in @rbiasini commit 8291971d9a2d83a071f9c3e190aeb298e5a251a2

PR #144, opened 2018-11-21
GM: Stock ECU presence monitoring

vntarasov:
Allow Openpilot to engage after stock ECUs are unplugged, without Panda reboot.

Common GM setup is to plug in a Panda first, and to figure out ASCM removal later. Existing Panda safety goes to silent mode indefinitely, meaning even after ASCM is removed, Openpilot still can't send car control commands. That adds "have you tried unplugging Panda from OBD port and plugging it back in" to one of the many troubleshooting steps for new GM users. Instead, this change makes Panda to forget it saw ASCM in one minute.

Also, this change unifies bus number check in rx_hook to single check at the top, since all of the safety car state only comes from powertrain bus.


vntarasov:
Need to double check if any changes are needed now that we have GM_PASSIVE safety model. Closing for now.

PR #145, opened 2018-12-17
Unified Diagnostic Services (UDS) panda library (ISO 14229)

gregjhogan:
This library makes it a lot easier to flash ECUs over CAN using a panda, as well as perform general scanning of ECUs for interesting data

some services are not fully tested because I don't have an ECU that responds to them, and some are not fully implemented because they don't seem that useful and are not trivial (see TODO comments)

https://en.wikipedia.org/wiki/Unified_Diagnostic_Services


BogGyver:
any chance we can add code to allow for Tesla return address calculations (addr + 0x10 instead of addr + 0x08) as a setting?

gregjhogan:
@bogyver you can do that now! see `rx_addr` parameter!

BogGyver:
Awesome!

python/uds.py
```diff
@@ -0,0 +1,717 @@
...
+    try:
+      rx_frame = {"size": 0, "data": "", "idx": 0, "done": True}
+      tx_frame = {"size": 0, "data": "", "idx": 0, "done": True}
+
+      # allow all output
+      self.panda.set_safety_mode(0x1337)
+      # clear tx buffer
+      self.panda.can_clear(self.bus)
+      # clear rx buffer
+      self.panda.can_clear(0xFFFF)
```
geohot: Is that what this does?

python/uds.py
```diff
@@ -0,0 +1,717 @@
+#!/usr/bin/env python
+import time
+import struct
+from enum import IntEnum
+from Queue import Queue, Empty
```
geohot: Python 3?

python/uds.py
gregjhogan: https://github.com/commaai/panda/blob/4972376de36aff020694d06d2458d73a5f4da829/board/main.c#L499

python/uds.py
gregjhogan: should be good now

gregjhogan:
if the API for this library looks good, it is ready to merge

gregjhogan:
before we merge this, make the CAN i/o thread work off a rx/tx queue pair and move the `panda.can_send()`/`panda.can_recv()` calls only in this thread

PR #147, opened 2018-12-26
GM: handle simultaneous GMLAN sends

vntarasov:
Handle simultaneous low-speed GMLAN bitbang sends by queuing up a few last messages.
Needed for simultaneous LKA icon and chime messages: when openpilot is disengaged, both LKA icon needs to be cleared, and disengagement chime needs to be sent.

Had to ifdef-out debug strings, to keep firmware size under 32k. To have debug strings _and_ have working Panda, I had to exclude some safety models in my local testing.

Test: send a bunch of low-speed GMLAN messages through bitbang at once, and check that the last one still goes through:
```
from panda import Panda
import time

panda = Panda()
panda.set_gmlan(0)
while (True):
  panda.can_send_many([
    [0x104c006c, 0, "\x40\x40\x18", 3], # LKA icon
    [0x104c006c, 0, "\x40\x40\x18", 3],
    [0x104c006c, 0, "\x40\x40\x18", 3],
    [0x104c006c, 0, "\x40\xc0\x14", 3],
    [0x10400060, 0, "\x82\x10\x01\xff\x00", 3], # chime
  ])
  time.sleep(1)
```

### DIFF
--- a/board/drivers/gmlan_alt.h
+++ b/board/drivers/gmlan_alt.h
@@ -3,6 +3,14 @@
 #define GMLAN_HIGH 0 //0 is high on bus (dominant)
 #define GMLAN_LOW 1 //1 is low on bus
 
+// Excluding debug strings to save on firmware size
+// #define GMLAN_DEBUG
+#ifdef GMLAN_DEBUG
+#define gmlan_puts(a) puts(a)
+#else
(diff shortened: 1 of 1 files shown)


vntarasov:
Rebasing & testing needed, closing for now.

PR #150, opened 2019-01-22
optimize board build for size to avoid going over the limit.

adhintz:
Here are the sizes of board/obj/panda.bin with different settings:
-O2 32228
-Og 30172
-Os 28028

### DIFF
--- a/board/build.mk
+++ b/board/build.mk
@@ -1,4 +1,4 @@
-CFLAGS += -I inc -I ../ -nostdlib -fno-builtin -std=gnu11 -O2
+CFLAGS += -I inc -I ../ -nostdlib -fno-builtin -std=gnu11 -Os
 
 CFLAGS += -Tstm32_flash.ld
 


adhintz:
For now I just left it as a warning if the size is too big. Perhaps doing `exit 1` would also be approrpriate.

rbiasini:
Nice. I think an error would be appropriate.

adhintz:
Changed it to an error and added the exit

adhintz:
How did you flash it?

I've been doing mine by running 
cd /data/openpilot/panda ; pkill -f boardd ; PYTHONPATH=.. python -c "from panda import Panda; Panda().flash()"

From an EON connected to a panda that's on a debug board.

adhintz:
I'll try it from a PC in a bit.

rbiasini:
oh, I'm just doing:
```
cd /data/openpilot/panda/board/
make clean
make
```

From a pc connected to a Panda that it's on a debug board.

rbiasini:
oh wait, I might have lied: I was using our internal repo. Trying now from the public openpilot and I get the green fast blinking when bin size exceeds 32kB...

energee:
This will probably fix https://github.com/commaai/panda/commit/20e8fa9716195d84a1e59f4c804ff117845cff8a#diff-36d67b6d2c9d538cd0d183030956fbabR54

energee:
That makes sense, I've gotten rapid green before during FW flashes that should work. These failures are hard to replicate, they seem more common when flashing a panda that is several versions behind but could be coincidence.

PR #161, opened 2019-02-23
change Chrysler rate checks so that they do not fault on actual drives.

adhintz:


rbiasini:
weird, so you see `dist_to_meas_check` sporadically blocking messages? `chrysler_torque_meas` contains the last 6 samples and they are using conservatively. So there should be no problems even if panda/EON have some timing issues.
Note that this safety mechanism "should" be a 1:1: copy from Toyota

adhintz:
Yes, basically any OpenPilot drive will trigger the dist_to_meas_check to block messages. Is the intent of dist_to_meas_check to disable OpenPilot if the driver moves the wheel? Chrysler OpenPilot doesn't disengage if the driver moves the wheel, and if the LKAS_COMMAND messages are blocked for 100ms+ then the EPS faults for the rest of the drive.

For the dist_to_meas_check, you're using EPS_STATUS TORQUE_MOTOR, correct? Just want to make sure that I'm not reading the RDHR bit math wrong. I'm not sure what the TORQUE_MOTOR value is supposed to be, but it changes even when OP / LKAS are not in use, just by the driver turning the wheel.

For an example of what dist_to_meas_check would block with existing code, see drive https://community.comma.ai/cabana/?route=03efb1fda29e30fe|2019-02-21--18-03-45#
The first blocked message would be ['359.0192872819998', '658', '128', '14120000c07c0000']
Then 14 seconds later a bunch of messages would be blocked starting with 
374.5361421719999,658,128,14690000b0f80000
Let me know if you want me to email you the Cabana Share Link if you don't have easy magic access to the drive.

rbiasini:
The intent of checking `dist_to_meas` is to ensure that the difference between commanded torque (`LKAS_COMMAND LKAS_STEERING_TORQUE`) and produced torque by EPS (`EPS_STATUS TORQUE_MOTOR`) is within a threshold. Indeed, when the userapplies torque on the wheel, the difference between commanded and produced torques enlarges. When this happens, we want to reduce the allowed commanded torque to not "fight" the driver's intent. 

Regardless, the safety logic in panda should only be replication of what's in `carcontroller.py`. It's surprising that messages get blocked by panda, if panda has the same exact logic that is in carcontroller.

adhintz:
Thanks for the explanation! I'll look for the difference in logic between `carcontroller.py` and `safety_chrysler.h` If the logic is the exact same, then I'll see if there's a timing issue where they're looking at different sets of data.

Unrelated to this, when there's a safety violation are you fine with the Panda both blocking the original message and sending its own LKAS_COMMAND with LKAS_STEERING_TORQUE=0 ? Those types of messages are what's sent even if OP or stock LKAS is not engaged.

Similarly, I'm considering having Panda send LKAS_COMMAND with LKAS_STEERING_TORQUE=0 messages if OP appears to be lagging. The EPS will fault if there's no LKAS_COMMAND for 200ms+.  I see OP+EON lag for that amount of time perhaps once every 6 hours of operation. I could either spend time tracking down the lag or just have the Panda send LKAS_COMMAND messages if it hasn't seen one in a while. Sound reasonable?

rbiasini:
> then I'll see if there's a timing issue where they're looking at different sets of data.

Race condition between `EPS_STATUS TORQUE_MOTOR` and 
`LKAS_COMMAND LKAS_STEERING_TORQUE` might be an issue. This is currently mitigated by panda taking the last 6 valid samples of `EPS_STATUS TORQUE_MOTOR` and use them conservatively. This is probably the source of the issue you see, but the root cause seems to be openpilot hanging for too long and I don't see this happening on any other car (Toyota has the same logic).

> Unrelated to this, when there's a safety violation are you fine with the Panda both blocking the original message and sending its own LKAS_COMMAND with LKAS_STEERING_TORQUE=0 ? Those types of messages are what's sent even if OP or stock LKAS is not engaged. 

I would prefer not to, because it would involve panda knowing about checksum and counters, which I would prefer to avoid to be 100% sure that panda isn't capable of sending valid steering messages. In other words, I prefer seeing the panda only as a blocker, rather than a corrector.

> I could either spend time tracking down the lag or just have the Panda send LKAS_COMMAND messages if it hasn't seen one in a while. Sound reasonable?

I would be a lot happier if we can track down why openpilot sometime hangs, but I realize that a EPS fault every 6 hours of operation is unacceptable from a quality standpoint.

adhintz:
I have not seen any issues with this recently. I'm not completely certain why, but I never found out why it was triggering in the first place.

Thanks for pushing back on this! Closing this as it seems to not be needed.

PR #157, opened 2019-02-06
WIP: VW safety

cbenhagen:
This is an initial version of the safety for VW cars. 

TODO:

- [ ] Test drive
- [ ] Tune constants

PR #162, opened 2019-03-01
unify honda and toyota pedal code

geohot:
Pull request #159 was extremely ugly and should not have been merged. Hundreds of lines of code should never be copy and pasted for a 2 line change. It is a big cost going forward.

This pull request unifies the Honda and Toyota pedals. This is untested, and should be tested on a Honda and Toyota before it is merged.

I also question comma's position on the pedal. If this code is going to be upstreamed, it tacitly implies that comma.ai will maintain the pedal firmware, even it not the pedal itself. If they do not intend to maintain this code, it should be deleted from the upstream panda repo. No code is better than unmaintained code.


geohot:
Actually thinking about this for a minute, there's no reason these builds should even be different! openpilot should be refactored to support multiple different checksum types. Without a counter, the Toyota checksum is not as safe as the Honda checksum.

rbiasini:
Below clarifications on safety rules that applies to any repo maintained by comma.ai.

**What are comma.ai's safety rules?** 
A high level abstraction of our safety rules can be found at https://github.com/commaai/openpilot/blob/devel/SAFETY.md
The safety file is contained in openpilot.comma.ai repository, since https://github.com/commaai/opendbc and https://github.com/commaai/panda are openpilot's submodules.

**How are those safety rules enforced?**
We run extensive in-vehicle tests and we have continuous integration tests running both on our internal openpilot repo and as well on the public panda repo to ensure that code changes won't create regressions in the safety code.

**What parts of the code must meet those safety rules?**
Everything that can interact with car's actuators (gas, steer, brake) using an unmodified version of openpilot.

**So, which parts of the code do not have to meet those safety rules?** 
Any code that cannot be utilized with an unmodified version of openpilot, or that does not interact with the car;s actuators (gas, steer, brake). Examples:
- Tesla safety code: https://github.com/commaai/panda/blob/master/board/safety/safety_tesla.h.
- VW safety code (not merged yet): https://github.com/commaai/panda/pull/157.
Both Tesla and VW complete ports aren't upstreamed to openpilot yet and they only live within community maintained forks of openpilot, but we are happy to merge WIP safety models for new brands of vehicles into panda before the their openpilot port is complete. 

That said, we'll do our due diligence to make sure that any proposed safety model is reasonably within our safety rules before being merged.

**And the pedal interceptor FW?**
The pedal interceptor is a community project that is not officially supported nor endorsed by comma. However, it is hosted on the official openpilot repo and it can be actuated using an unmodified version of openpilot. From now on, the pedal interceptor related code will undergo the same safety scrutiny as any other part of our code that must obey to our safety rules.  #164 addresses most of the concerns.

PR #163, opened 2019-03-02
Unified pedal counter

rbiasini:


geohot:
Why do this? Why should they be different at all?

rbiasini:
all right, makes sense. openpilot side, I still want to do the checksum/counter checks in can-parser: you can't really do the counter check outside the parser. So we'll define a 4-bits counter, 8-bits CRC checksum, 6 bytes message for both 0x200 and 0x201. The can-parser will perform the counter+checksum checks for 0x201 (the one sent by the pedal).

geohot:
Cool, as long it's the same firmware for both.

PR #166, opened 2019-03-08
L-line relay

legonigel:


legonigel:
@rbiasini Fixed the clicking on startup/flashing

board/drivers/llgpio.h
```diff
@@ -14,6 +14,10 @@ void set_gpio_output(GPIO_TypeDef *GPIO, int pin, int val) {
   set_gpio_mode(GPIO, pin, MODE_OUTPUT);
 }
 
+int get_gpio_output(GPIO_TypeDef *GPIO, int pin) {
```
rbiasini: not used, I can remove it, right? @legonigel

legonigel: Sure.

board/drivers/can.h
```diff
@@ -4,6 +4,7 @@
 #define ALL_CAN_LIVE 0
 
 int can_live = 0, pending_can_live = 0, can_loopback = 0, can_silent = ALL_CAN_SILENT;
+int relay_control = 0;
```
legonigel: I think this int should live in lline_relay.h and use extern here.

legonigel:
LGTM!

board/pedal/main.c
```diff
@@ -295,6 +297,7 @@ int main() {
   puts("**** INTERRUPTS ON ****\n");
   __enable_irq();
 
+  power_save_init();
```
rbiasini: @legonigel is this needed?

legonigel: Nope. This was removed in master, but got left in here on accident.

board/safety.h
```diff
@@ -91,6 +97,10 @@ int safety_fwd_hook(int bus_num, CAN_FIFOMailBox_TypeDef *to_fwd) {
   return current_hooks->fwd(bus_num, to_fwd);
 }
 
+int safety_relay_hook(int to_set) {
+  return current_hooks->relay(to_set);
```
rbiasini: no reason to pass along the input `to_set`, right? this function and the relay hooks can all have `void` as input

legonigel: Sure. This was done to match the style of the other safety hooks.

rbiasini: you mean we have safety hooks with unused input?

legonigel: Just remove it

appleguru:
@legonigel We have some conflicts with our downstream Tesla branches and this change. Can you help me understand the functionality here so I can work on some changes that help them coexist better? If I understand correctly, the goal here was to use one of the LIN outputs as a relay control? Are you using this in conjunction with the GMLAN output? Or? ...on our branches, we are using TIM5 and L-Line together for some actual LIN output. So definitely should be doable to coexist as they shouldn't need to ever be enabled at the same time; just need to sort out the best way how.

legonigel:
@appleguru This L-Line relay is a relay over the LIN line. There is a relay with + side connected to +12v in the giraffe, and the - side is connected to L-Line on LIN. When not energized this relay connects CAN 0 with CAN 2, and when it is energized it disconnects the two cans and connects terminating resistors for both CANS. This relay currently only exists in the Subaru Giraffe, but will probably be added to other cars in the future. The relay allows the car to start in stock mode, then switch on the relay and go to openpilot mode once openpilot tries to send can messages. This prevents the car from faulting if the EON turns on after the car is started.
With this PR the panda can use the L-Line to turn on or off this relay, and the timer is used to toggle the pin so that the LIN transceiver doesn't time out. The code for this is actually somewhat based on the GMLAN code from the Tesla branches. 
Let me know if you have any more questions.

appleguru:
Ok, so the biggest issue I see so far is that in gpio.h you do:
`// LLine now used for relay output
    set_gpio_output(GPIOC, 10, 1);
    //set_gpio_alternate(GPIOC, 10, GPIO_AF7_USART3);`

If I read that correctly, you are essentially disabling LIN and hardcoding L-line to be relay control. Not sure why this is needed, as you have `set_gpio_output(GPIOC, 10, 1);` in `lline_relay_init()` also (which seems more appropriate). Can we revert this line?

legonigel:
I think we need to come up with a better solution. When using the relay
this line would cause the relay to click once, which we would like to
avoid. I will try to think of a better solution.

appleguru:
Maybe leave it how it was in the upstream branch, but change it in your downstream branch where you're actually using it? For reference, this is how we're working around this for now (restoring this line and calling our TIM5_IRQHandler from yours conditionally):

https://github.com/BogGyver/openpilot/commit/55cbddd319b7974dd717218a4b02b8e722219ff1

PR #168, opened 2019-03-12
Dcp remove

rbiasini:
This PR removes DCP mode when EON builds panda code. It avoids entering in DCP mode when EON takes long time to boot.


geohot:
There's no need to remove most of this stuff, like entering DCP on key over serial line. You just need to remove the auto DCP.

rbiasini:
Done. Merged for in-car testing.

PR #170, opened 2019-03-14
More Power saving

legonigel:
Not ready to merge. Needs to be tested.

 - CAN is turned off via USB.
 - CAN is turned on when either try to transmit or can is received
 - If only transmit is asleep, all messages should send okay
 - If receive is alseep, will miss first message while waking up
 - Sometimes will report error on second message while CAN perif wakes up
 - Saves 130mW!


legonigel:
@rbiasini Needs to be tested in a car, but this is the rest of the power saving for the panda.
Grey panda 650mW -> 325mW
White panda 460mW -> 260mW

board/power_saving.h
```diff
@@ -0,0 +1,160 @@
...
+  TIM6->EGR = TIM_EGR_UG;
+  TIM6->CR1 |= TIM_CR1_CEN;
+}
+
+
+
+// Reset timer when activity
+void power_save_reset_timer() {
+  TIM6->CNT = 0;
+#ifdef EON
```
rbiasini: this ifdef is not needed

board/main.c
```diff
@@ -141,6 +143,9 @@ void usb_cb_ep2_out(uint8_t *usbdata, int len, int hardwired) {
   uart_ring *ur = get_ring_by_number(usbdata[0]);
   if (!ur) return;
   if ((usbdata[0] < 2) || safety_tx_lin_hook(usbdata[0]-2, usbdata+1, len-1)) {
+#ifdef EON
```
rbiasini: not needed

board/power_saving.h
```diff
@@ -0,0 +1,160 @@
...
+}
+
+void TIM6_DAC_IRQHandler(void) {
+  puts("TIM6 ");
+  putui(TIM6->SR);
+  puts("\n");
+  //Timeout switch to power saving mode.
+  if (TIM6->SR & TIM_SR_UIF) {
+    TIM6->SR = 0;
+#ifdef EON
```
rbiasini: this is ok, the only needed ifdef EON

Issue #154, opened 2019-02-05
pandacan doesn't work: USBErrorBusy: LIBUSB_ERROR_BUSY on python 2 and ModuleNotFoundError: No module named 'dfu' on python 3


shurikrulik:
python 2:

Python 2.7.15rc1 (default, Nov 12 2018, 14:31:15) 
[GCC 7.3.0] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> from panda import Panda
>>> Panda.list()
[u'480031000551363338383037']
>>> panda = Panda()
opening device 480031000551363338383037 0xddcc
exception LIBUSB_ERROR_BUSY [-6]
Traceback (most recent call last):
  File "/home/burningman/.local/lib/python2.7/site-packages/panda/__init__.py", line 163, in connect
    self._handle.claimInterface(0)
  File "/home/burningman/.local/lib/python2.7/site-packages/usb1/__init__.py", line 1300, in claimInterface
    libusb1.libusb_claim_interface(self.__handle, interface),
  File "/home/burningman/.local/lib/python2.7/site-packages/usb1/__init__.py", line 133, in mayRaiseUSBError
    __raiseUSBError(value)
  File "/home/burningman/.local/lib/python2.7/site-packages/usb1/__init__.py", line 125, in raiseUSBError
    raise __STATUS_TO_EXCEPTION_DICT.get(value, __USBError)(value)
USBErrorBusy: LIBUSB_ERROR_BUSY [-6]
connected
>>> panda.can_recv()
CAN: BAD RECV, RETRYING
CAN: BAD RECV, RETRYING
CAN: BAD RECV, RETRYING
CAN: BAD RECV, RETRYING
CAN: BAD RECV, RETRYING

python 3:

Python 3.6.7 (default, Oct 22 2018, 11:32:17) 
[GCC 8.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from panda import Panda
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/burningman/.local/lib/python3.6/site-packages/panda/__init__.py", line 11, in <module>
    from dfu import PandaDFU
ModuleNotFoundError: No module named 'dfu'

Drivers installed, 11-panda.rules edited, panda connected through USB cable and led indicates ready to work status. What is the problem ?

shreyansh073:
The problem lies in the script panda/python/__init__.py at line 193.
[...]

PR #174, opened 2019-03-25
Fixing libusb busy error

shreyansh073:
Detaching all kernels before attempting to claim the usb

### DIFF
--- a/python/__init__.py
+++ b/python/__init__.py
@@ -173,6 +173,7 @@ class Panda(object):
                 self.bootstub = device.getProductID() == 0xddee
                 self.legacy = (device.getbcdDevice() != 0x2300)
                 self._handle = device.open()
+                self._handle.setAutoDetachKernelDriver(True)
                 if claim:
                   self._handle.claimInterface(0)
                   #self._handle.setInterfaceAltSetting(0, 0) #Issue in USB stack


legonigel:
Does this actually fix a problem? And what OS did you see this on?

I've never seen where a kernel driver needs to be detached for the panda connection. Only seen busy error when it was claimed by some other code running on my system.

shreyansh073:
> Does this actually fix a problem? And what OS did you see this on?
> 
> I've never seen where a kernel driver needs to be detached for the panda connection. Only seen busy error when it was claimed by some other code running on my system.

Issue https://github.com/commaai/panda/issues/154

I was facing the same problem with python 2.7 on Ubuntu 16.04. Detaching kernel before claiming them fixed the issue for me. You can refer to the following thread on stack overflow for more details: https://stackoverflow.com/questions/6365314/libusb-interface-already-claimed

legonigel:
Could you provide steps to reproduce this? I would like to know what driver is claiming the device when this happens.

shreyansh073:
I am not sure which driver is claiming the usb before panda. But I have
been troubled by this bug for more than a month. There have been
discussions happening in the discord community regarding this bug for quite
some time now.

Detaching kernels solved the bug for me. I hope this helps. Please let me
know if it is not the correct solution, because I am still not able to
cansend().

Thanks and regards,
Shreyansh.

zorrobyte:
Same issue here. Let's get it merged! 🎉

legonigel:
@zorrobyte That stack trace looks like it didn't find a panda, given it failed at the `assert(self._handle != None)`. If this issue is what you are seeing, it should fail at `self._handle.claimInterface(0)`

@shreyansh073 Can you confirm what stack trace you get?

Also, can you [rebase](https://github.com/edx/edx-platform/wiki/How-to-Rebase-a-Pull-Request) your pull request on the latest master?

dekerr:
This is a permission error. launch python with sudo.

Sluder:
Was this never merged? Im having this issue, and it looks like it was never merged. https://github.com/commaai/panda/blob/master/python/__init__.py#L175

PaulSD:
For Linux, to see what process has already claimed the device:
```
$ sudo -s
# lsusb
Bus 003 Device 009: ID bbaa:ddcc
# lsof | grep /dev/bus/usb/003/009
```

I came across this PR because I got a LIBUSB_ERROR_BUSY error on Ubuntu 18.04, but it turns out that was because I actually had two processes connected to the device at the same time.  As long as I don't have two processes connected at the same time, I don't have any issues with this.

PaulSD:
I ran into LIBUSB_ERROR_BUSY again after installing the SocketCAN kernel module (https://github.com/commaai/panda/tree/master/drivers/linux).

Userspace software is unable to access the Panda at the same time as the SocketCAN kernel module.  Calling `self._handle.setAutoDetachKernelDriver(True)` in this case does not help.  Note that `lsof` does not show devices claimed by kernel modules, so `lsof` will not identify that the SocketCAN kernel module has claimed the device in this case.

To temporarily disable the SocketCAN kernel module so that Python can access the Panda, run `modprobe -r panda`.  To re-enable it, run `modprobe panda`.

PR #176, opened 2019-04-02
Stop charging EON when battery voltage is low

legonigel:
Untested, not ready to merge.


board/main.c
```diff
@@ -627,7 +643,15 @@ int main() {
           }
           break;
         case USB_POWER_CDP:
-#ifndef EON
+#ifdef EON
+          if (power_save_status == POWER_SAVE_STATUS_ENABLED) {
```
rbiasini: I think `POWER_SAVE_STATUS_ENABLED` can only be reached if `#ifdef EON` is true, so I'll just remove it.

rbiasini:
Ended up making a new state (CLIENT_TEMP) to not overload the existing CLIENT mode.
I have considered to force a transition from CLIENT mode to CDP when exiting power saving mode, but that would make a panda flashed from a EON unusable when connected to a pc, which is annoying.

Maybe there is a more elegant solution? @legonigel

PR #177, opened 2019-04-03
chrysler: forward bus 0 to bus 2

adhintz:
This keeps the stock LKAS camera active so that we can copy data from its messages.

I'll open a PR in openpilot with code that uses this to support arbitrary Chrysler models, including 2019.

### DIFF
--- a/board/safety/safety_chrysler.h
+++ b/board/safety/safety_chrysler.h
@@ -5,7 +5,7 @@ const int CHRYSLER_MAX_RATE_UP = 3;
 const int CHRYSLER_MAX_RATE_DOWN = 3;
 const int CHRYSLER_MAX_TORQUE_ERROR = 80;    // max torque cmd in excess of torque motor
 
-int chrysler_camera_detected = 0;
+int chrysler_camera_detected = 0;             // is giraffe switch 2 high?
 int chrysler_rt_torque_last = 0;
 int chrysler_desired_torque_last = 0;
 int chrysler_cruise_engaged_last = 0;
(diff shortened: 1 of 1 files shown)


board/safety/safety_chrysler.h
```diff
@@ -125,12 +125,21 @@ static int chrysler_tx_hook(CAN_FIFOMailBox_TypeDef *to_send) {
...
 
+static int chrysler_fwd_hook(int bus_num, CAN_FIFOMailBox_TypeDef *to_fwd) {
+  int32_t addr = to_fwd->RIR >> 21;
+  // forward CAN 0 -> 2 so stock LKAS camera sees messages
+  if (bus_num == 0 && addr != 0x2d9 && addr != 0x2a6 && addr != 0x292) {
+    return 2;
+  }
+  return -1;  // do not forward
+}
+
```
rbiasini: This is good. there should be a detection of camera messages on BUS 0 though, in order to avoid forwarding when the giraffe is set to "stock"

adhintz: Implemented and tested that it works in a car.

board/safety/safety_chrysler.h
```diff
@@ -125,12 +125,29 @@ static int chrysler_tx_hook(CAN_FIFOMailBox_TypeDef *to_send) {
   return true;
 }
 
+static void chrysler_init(int16_t param) {
+  chrysler_camera_detected = 0;
```
rbiasini: this needs to be set to one camera msgs are on bus 0, right?

adhintz: Yeah, it already is up on line 50:
```
  // check if stock camera ECU is still online
  if (bus == 0 && addr == 0x292) {	  if (bus == 0 && addr == 0x292) {
    chrysler_camera_detected = 1;	    chrysler_camera_detected = 1;
```

rbiasini: ah, right!

PR #178, opened 2019-04-03
tools: Add flash_panda

dougsland:
Instead of ask users to copy and past a few lines
from wiki lets provide a ready to go tool.

### DIFF
--- a/tools/flash_panda
+++ b/tools/flash_panda
@@ -1,0 +1,25 @@
+#!/usr/bin/env python
+import os
+import sys
+
+from panda import Panda
+
+PANDA_DIR = '/data/openpilot/panda'
+
+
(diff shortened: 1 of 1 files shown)


legonigel:
Panda's are flashed automatically when the EON is restarted. When pandad is started, it checks if the connected Panda is running the latest firmware and flashes it if it is not. A panda can also be flashed in one command by navigating to `panda/board` and running make. I don't think this script is necessary.

dougsland:
@legonigel thanks, going to give a try about the make. I wish we could have something in /bin so no need to navigate to dirs.

dougsland:
@legonigel Additionally, when the panda is flashing red, even restarting EON never helped.

PR #183, opened 2019-04-17
fix usb flashing

dekerr:
libusb1 versions >1.6.6 do not work


legonigel:
Was fixed in the requirements, but not setup. Thanks!

PR #184, opened 2019-04-22
generate dependencies and clean

dekerr:
Main.o and bootstub.o were missing headers (like recent power_saving.h). Generate them instead to guarantee makefile re-makes if any dependency is modified.

### DIFF
--- a/board/build.mk
+++ b/board/build.mk
@@ -50,7 +50,6 @@ include ../common/version.mk
 obj/cert.h: ../crypto/getcertheader.py
 	../crypto/getcertheader.py ../certs/debug.pub ../certs/release.pub > $@
 
-
 obj/%.$(PROJ_NAME).o: %.c obj/gitversion.h obj/cert.h $(DEPDIR)/%.d
 	$(CC) $(DEPFLAGS) $(CFLAGS) -o $@ -c $<
 	$(POSTCOMPILE)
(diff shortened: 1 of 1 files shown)


board/build.mk
```diff
@@ -59,12 +65,18 @@ obj/$(PROJ_NAME).bin: obj/$(STARTUP_FILE).o obj/main.$(PROJ_NAME).o
...
-	@BINSIZE=$$(du -b "obj/$(PROJ_NAME).bin" | cut -f 1) ; if [ $$BINSIZE -ge 32768 ]; then echo "ERROR obj/$(PROJ_NAME).bin is too big!"; exit 1; fi;
-
+	@BINSIZE=$$(du -b "obj/$(PROJ_NAME).bin" | cut -f 1) ; \
+	if [ $$BINSIZE -ge 32768 ]; then echo "ERROR obj/$(PROJ_NAME).bin is too big!"; exit 1; fi;
 
 obj/bootstub.$(PROJ_NAME).bin: obj/$(STARTUP_FILE).o obj/bootstub.$(PROJ_NAME).o obj/sha.$(PROJ_NAME).o obj/rsa.$(PROJ_NAME).o
 	$(CC) $(CFLAGS) -o obj/bootstub.$(PROJ_NAME).elf $^
 	$(OBJCOPY) -v -O binary obj/bootstub.$(PROJ_NAME).elf $@
 
+$(DEPDIR)/%.d: ;
```
legonigel: Shouldn't the dependency file be created here, not in the main compile step so that make can create them at include time?

legonigel:
If I understand this correctly, clang will make a dependency file that is imported by make to add all the dependencies? If so, this is great. I had development problems with not recompiling header changes, but didn't realize this type of fix exists.

dekerr:
Yeah, that's the gist of it.

legonigel:
Shouldn't the dependency file be created inside `$(DEPDIR)/%.d:`, not in the main compile step so that make can create them at include time?

dekerr:
$(DEPDIR)/%.d: ; is just here as an empty recipe in case a dependency file doesn't exist. 
I'm not sure how files could be created elsewhere since they are generated by compiler flags

PR #189, opened 2019-05-10
Subaru Safety: added different message addresses for non-Global platform

rbiasini:
@bugsy924 , can you please review?


bugsy924:
I trimmed the lowest bit (when compared to the dbc) to keep driver torque within a reasonable limit. The outback/legacy torque sensor is more sensitive than on global so the safety limits would keep tripping on disengagement. This pause (because of violated safety limit) in sending ES_LKAS when torque is high is enough to cause the steering controller to lock out.

I would love to test your revision, but eyesight on this outback seems to be stuck in some lock out mode

bugsy924:
outback is all functional again
loaded panda with this and it works perfectly with openpilot

rbiasini:
Superseeded by https://github.com/commaai/panda/pull/259

PR #194, opened 2019-05-21
panda code refactor pre MISRA

geohot:


geohot:
Tests passing in CI. Tested a replay locally and tested that the pedal has the same behavior (20k -> 3k of code!).

Power saving doesn't trigger until the car has gone through a cycle, but the blinking blue LED means power draw is around 100mW. Will get the EON down below 900, and we should be good on the killing car issue.

Look this over, let's merge it, then we can test in cars (or maybe test in one car first?).

@rbiasini @andyh2 @pd0wm

board/main.c
```diff
@@ -505,6 +490,98 @@ void __attribute__ ((noinline)) enable_fpu() {
...
+  if (TIM3->SR != 0) {
+    can_live = pending_can_live;
+
+    //puth(usart1_dma); puts(" "); puth(DMA2_Stream5->M0AR); puts(" "); puth(DMA2_Stream5->NDTR); puts("\n");
+
+    uint32_t current = adc_get(ADCCHAN_CURRENT);
+
+    switch (usb_power_mode) {
+      case USB_POWER_CLIENT:
+        if ((tcnt-marker) >= CLICKS) {
```
rbiasini: minor thing and not related to this refactor: I would compute `(tcnt-marker) >= CLICKS` once, before L508 and collapse the if conditions
```
if (a) {
  if (b) {
```
into 
```
if (a && b) {
```

board/main.c
```diff
@@ -577,136 +644,65 @@ int main() {
   // default to silent mode to prevent issues with Ford
   // hardcode a specific safety mode if you want to force the panda to be in a specific mode
   safety_set_mode(SAFETY_NOOUTPUT, 0);
```
rbiasini: for VIN query we'll need ELM327 mode, but I guess that can be the default mode set by boardd, while ok for panda to boot in NOOUTPUT.

andyh2: this default is sane

andyh2:
LGTM, let’s get it in our cars. signaling power efficiency via blue LED is nice touch.

board/drivers/clock.h
```diff
@@ -0,0 +1,40 @@
+void clock_init() {
+  // enable external oscillator
+  RCC->CR |= RCC_CR_HSEON;
+  while ((RCC->CR & RCC_CR_HSERDY) == 0);
+
+  // divide shit
```
LemonAndroid: language

board/drivers/spi.h
```diff
@@ -23,8 +28,8 @@ void spi_init() {
 
   // setup interrupt on falling edge of SPI enable (on PA4)
   SYSCFG->EXTICR[2] = SYSCFG_EXTICR2_EXTI4_PA;
-  EXTI->IMR = (1 << 4);
-  EXTI->FTSR = (1 << 4);
+  EXTI->IMR |= (1 << 4);
+  EXTI->FTSR |= (1 << 4);
```
LemonAndroid: pls explain

edsammy: https://www.st.com/content/ccc/resource/technical/document/reference_manual/group0/81/ea/88/1f/97/9e/4a/d0/DM00305666/files/DM00305666.pdf/jcr:content/translations/en.DM00305666.pdf

PR #196, opened 2019-05-25
hex conversion refactor

dekerr:


geohot:
I don't think this is more readable

PR #197, opened 2019-05-25
update memcpy

dekerr:
copy 4 bytes at a time if possible
disregards misra (11.5 advisory):
> Conversion of a pointer to void into a pointer to object may result
>in a pointer that is not correctly aligned, resulting in undefined
>behaviour. It should be avoided where possible but may be necessary,
>for example when dealing with memory allocation functions. If
>conversion from a pointer to object into a pointer to void is used,
>care should be taken to ensure that any pointers produced do not
>give rise to the undefined behaviour discussed under Rule 11.3.


dekerr:
I think this is good to go now

rbiasini:
yeah, this new version is pretty quick, but break quite a few Misra things. Let's keep it open until Misra refactor is done.

dekerr:
all mem functions are going to break 11.5 since they have pointer void casting. No way around it. This doesn't break any more than the original memcpy.

geohot:
Hmm, can the processor handle misalignment okay?

dekerr:
armv7e can access unaligned memory. Implementation is based on this post:

[https://cboard.cprogramming.com/c-programming/154333-fast-memcpy-alternative-32-bit-embedded-processor-posted-just-fyi-fwiw.html#post1149163](url)

PR #200, opened 2019-06-04
Toyota gas cancellation

rbiasini:
- Added cancellation on gas pressed for Toyota
- minor nomenclature change
- fix bug in regression test (honda hook was called instead of toyota hook in gas interceptor test)
- added a complete set of tests to cover for gas interceptor
- remove unused TOYOTA_NOLIMITS safety mode

### DIFF
--- a/board/safety/safety_toyota.h
+++ b/board/safety/safety_toyota.h
@@ -54,12 +54,15 @@ static void toyota_rx_hook(CAN_FIFOMailBox_TypeDef *to_push) {
   if ((to_push->RIR>>21) == 0x1D2) {
     // 5th bit is CRUISE_ACTIVE
     int cruise_engaged = to_push->RDLR & 0x20;
-    if (cruise_engaged && !toyota_cruise_engaged_last) {
-      controls_allowed = 1;
-    } else if (!cruise_engaged) {
+    // 4th bit is GAS_RELEASED
+    int gas = !(to_push->RDLR & 0x10);
+    if (!cruise_engaged || (gas && !toyota_gas_prev && !gas_interceptor_detected)) {
       controls_allowed = 0;
+    } else if (cruise_engaged && !toyota_cruise_engaged_last) {
+      controls_allowed = 1;
     }
     toyota_cruise_engaged_last = cruise_engaged;
+    toyota_gas_prev = gas;
   }
 
   // exit controls on rising edge of gas press if interceptor (0x201)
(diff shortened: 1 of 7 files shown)


board/safety/safety_toyota.h
```diff
@@ -94,7 +109,7 @@ static int toyota_tx_hook(CAN_FIFOMailBox_TypeDef *to_send) {
     if ((to_send->RIR>>21) == 0x343) {
       int desired_accel = ((to_send->RDLR & 0xFF) << 8) | ((to_send->RDLR >> 8) & 0xFF);
       desired_accel = to_signed(desired_accel, 16);
-      if (controls_allowed && toyota_actuation_limits) {
+      if (controls_allowed) {
```
geohot: Why was the && here removed? Does toyota_actuation_limits toggle still work? If not, let's remove the safety mode entirely.

board/safety/safety_toyota.h
```diff
@@ -46,19 +48,32 @@ static void toyota_rx_hook(CAN_FIFOMailBox_TypeDef *to_push) {
...
+
+  // exit controls on rising edge of gas press if interceptor (0x201)
+  if ((to_push->RIR>>21) == 0x201) {
+    gas_interceptor_detected = 1;
+    int gas_interceptor = ((to_push->RDLR & 0xFF) << 8) | ((to_push->RDLR & 0xFF00) >> 8);
+    if ((gas_interceptor > TOYOTA_GAS_INTERCEPTOR_THRESHOLD) &&
+        (toyota_gas_interceptor_prev <= TOYOTA_GAS_INTERCEPTOR_THRESHOLD)) {
+      controls_allowed = 0;
+    }
+    toyota_gas_interceptor_prev = gas_interceptor;
```
geohot: gas_interceptor is global, but honda_gas_interceptor_prev and toyota_gas_interceptor_prev are specific to the file. Why?

board/safety/safety_toyota.h
```diff
@@ -189,7 +202,6 @@ const safety_hooks toyota_hooks = {
 
 static void toyota_nolimits_init(int16_t param) {
```
geohot: Remove this function and nolimits_hooks?

geohot:
A few comments inline. Remove nolimits entirely? I like the added tests.

board/safety/safety_toyota.h
rbiasini: We added it for convenience when testing, but I actually never used it. Yes, let me remove the safety mode.

board/safety/safety_toyota.h
rbiasini: yep

board/safety/safety_toyota.h
rbiasini: gas_interceptor is sent in the health packet so should be global. I wanted to limit the amount of global vars, so kept gas_interceptor_prev file specific. I can make them global too, it will simplify both safety code and tests.

geohot: If they have the exact same meaning I think they can be shared in a "safety.h global"

rbiasini: well, they are global anyway, just a matter of where they appear. The interceptor itself is abstracted from the specific car, so I'll just move gas_interceptor_prev to safety.h

rbiasini:
Tested in car the gas disable logic for Toyota and passing regression tests. Good to merge

PR #201, opened 2019-06-05
use __auto_type in max/min

dekerr:
from doc:

> Using __auto_type instead of typeof has two advantages:
>
> - Each argument to the macro appears only once in the expansion of the macro. This prevents the size of the macro expansion growing exponentially when calls to such macros are nested inside arguments of such macros.
>
> - If the argument to the macro has variably modified type, it is evaluated only once when using __auto_type, but twice if typeof is used.

### DIFF
--- a/board/config.h
+++ b/board/config.h
@@ -25,13 +25,13 @@
 #define COMPILE_TIME_ASSERT(pred) switch(0){case 0:case pred:;}
 
 #define min(a,b) \
- ({ __typeof__ (a) _a = (a); \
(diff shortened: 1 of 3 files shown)


rbiasini:
looks good (http://www.rowleydownload.co.uk/arm/documentation/gnu/gcc/Typeof.html), but it generates a lot more Misra violations. Do you know why?

dekerr:
appears to be due to the way nested calls are unwrapped since it shadows inner variables. Not sure if this is a shortcoming of the checker or unique to auto_type.

PR #205, opened 2019-06-06
replay test

quillford:
replay drives to test for safety violations. fails if any messages are blocked while controls are allowed


tests/safety/test.c
```diff
@@ -63,6 +63,18 @@ void set_timer(int t){
...
+void rx(CAN_FIFOMailBox_TypeDef *to_send) {
+  safety_rx_hook(to_send);
+}
+
+int tx(CAN_FIFOMailBox_TypeDef *to_push) {
+  return safety_tx_hook(to_push);
+}
+
+int set_mode(uint16_t  mode, int16_t param) {
+  return safety_set_mode(mode, param);
```
rbiasini: simplify `tests/safety` and use this instead of initting each car liek we do now

tests/safety_replay/replay_drive.py
```diff
@@ -0,0 +1,92 @@
...
+    if msg.which() == 'sendcan':
+      for canmsg in msg.sendcan:
+        to_send = libpandasafety_py.ffi.new('CAN_FIFOMailBox_TypeDef *')
+        to_send[0].RIR = canmsg.address << 21
+        to_send[0].RDTR = (canmsg.src & 0xF) << 4
+        to_send[0].RDHR = struct.unpack('<I', canmsg.dat.ljust(8, '\x00')[4:])[0]
+        to_send[0].RDLR = struct.unpack('<I', canmsg.dat.ljust(8, '\x00')[:4])[0]
+
+        # nanoseconds to microseconds
+        safety.set_timer(int(t/1000.))
```
rbiasini: no need to convert to float and then back to int

.circleci/config.yml
```diff
@@ -76,6 +76,23 @@ jobs:
...
+          name: Build image
+          command: "docker build -t panda_safety_replay -f tests/safety_replay/Dockerfile ."
+      - run:
+          name: Replay drives
+          command: |
+            mkdir /tmp/safety_replay
+            docker run -v /tmp/safety_replay:/tmp/safety_replay panda_safety_replay /bin/bash -c "cd /openpilot/panda/tests/safety_replay; PYTHONPATH=/openpilot ./test_safety_replay.py > /tmp/safety_replay/output.txt"
+      - store_artifacts:
+          name: Store safety_replay test output
+          path: /tmp/safety_replay/output.txt
```
rbiasini: output is small, so we can just print it in the terminal. No need to store in an artifact

PR #207, opened 2019-06-07
MISRA 16_x

dekerr:
fix misra 16 violations


rbiasini:
16_x alrady fixed in master. Thanks though.

PR #212, opened 2019-06-13
Toyota gas pressed: reconcile panda code with openpilot

rbiasini:


rbiasini:
tested in cars, works.

PR #214, opened 2019-06-13
Allow controls if ASCM is installed, but only if steering cmds are filtered

andrewcopenpilot:
This change removes overly aggressive ASCM detection, that once tripped required the panda to be power cycled before openpilot can be used again.

The ASCM sends a steering commands at 50Hz, When LKA is disabled the ASCM sends a null steering heart beat, So checking for this message originating from Bus 0 should reliably detect an unfiltered ASCM and force controls_allowed = 0.

This should provide two improvements to the GM community
1. This change allows GM Giraffe users the ability to swap between the GM Giraffe and the ASCM without power cycling the panda. (This is regularly done to recover from radar calibration faults)
2. Will allow the community to experiment with leaving the ASCM in place while filtering the steering commands with a secondary forwarding panda.

Tests: Verified initial code using panda dev board replaying logs from drives. Tested with ASCM in place filtering out steering commands (lateral only control). Tested with the ASCM in place and unfiltered to verify OP would not be allowed to engage. Tested filtering the ASCM filtered again to ensure OP could control the car again without power cycling the panda.

### DIFF
--- a/board/safety/safety_gm.h
+++ b/board/safety/safety_gm.h
@@ -22,8 +22,6 @@ const int GM_MAX_BRAKE = 350;
 int gm_brake_prev = 0;
 int gm_gas_prev = 0;
 int gm_speed = 0;
-// silence everything if stock car control ECUs are still online
-int gm_ascm_detected = 0;
 int gm_ignition_started = 0;
 int gm_rt_torque_last = 0;
 int gm_desired_torque_last = 0;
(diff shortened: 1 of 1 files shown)


beejayceeya:
This change would have saved me about 3 hours when I flipped from OP to stock and then back!

MiDe84:
This would alleviate a lot of head aches with the panda. I know I have ran into this before when I was new to openpilot. Thanks for working on a fix for this.

vntarasov:
I don't understand how this is supposed to work. Say one message from ASCM sets "controls_allowed" to 0, but then nothing prevents controls_allowed to go back to 1 on set and resume buttons. For fraction of a second until the next heartbeat sets controls_allowed back to 0, openpilot might send a conflicting message with undefined behavior on can bus with multiple senders of same address.

Also, avoiding panda reset when switching between giraffe and ASCM is achieved by https://github.com/commaai/panda/pull/144.

PR #219, opened 2019-06-17
Misra 15 7

rbiasini:
`An "if... else if" statement shall always be terminated with an "else".`

Avoided blank `else` statements.

### DIFF
--- a/board/safety/safety_gm.h
+++ b/board/safety/safety_gm.h
@@ -69,7 +69,8 @@ static void gm_rx_hook(CAN_FIFOMailBox_TypeDef *to_push) {
     // res/set - enable, cancel button - disable
     if ((buttons == 2) || (buttons == 3)) {
       controls_allowed = 1;
-    } else if (buttons == 6) {
+    }
+    if (buttons == 6) {
       controls_allowed = 0;
     }
   }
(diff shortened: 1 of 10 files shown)


board/safety/safety_gm.h
```diff
@@ -69,7 +69,8 @@ static void gm_rx_hook(CAN_FIFOMailBox_TypeDef *to_push) {
     // res/set - enable, cancel button - disable
     if ((buttons == 2) || (buttons == 3)) {
```
geohot: switch buttons?

board/safety/safety_honda.h
```diff
@@ -31,7 +31,8 @@ static void honda_rx_hook(CAN_FIFOMailBox_TypeDef *to_push) {
     int buttons = (to_push->RDLR & 0xE0) >> 5;
     if ((buttons == 4) || (buttons == 3)) {
```
geohot: switch buttons

board/safety/safety_toyota.h
```diff
@@ -126,12 +125,13 @@ static int toyota_tx_hook(CAN_FIFOMailBox_TypeDef *to_send) {
...
-      if (controls_allowed && long_controls_allowed) {
-        bool violation = max_limit_check(desired_accel, TOYOTA_MAX_ACCEL, TOYOTA_MIN_ACCEL);
-        if (violation) {
+      if (!controls_allowed || !long_controls_allowed) {
+        if (desired_accel != 0) {
           tx = 0;
         }
-      } else if (desired_accel != 0) {
+      }
+      bool violation = max_limit_check(desired_accel, TOYOTA_MAX_ACCEL, TOYOTA_MIN_ACCEL);
```
geohot: I like this pattern of if statements at the highest scope possible.

board/safety/safety_gm.h
rbiasini: you mean changing the var name?

rbiasini: oh, switch - case

PR #224, opened 2019-06-25
Fixed Misra 13.3: A full expression containing an increment (++) or d…

rbiasini:
…ecrement (--) operator should have no other potential side effects other than that caused by the increment or decrement operator


geohot:
Ugh 7 lines

PR #225, opened 2019-06-25
Misra 14.4: The controlling expression of an if statement and the controlling expression of an iteration‑statement shall have essentially Boolean type

rbiasini:


board/drivers/can.h
```diff
@@ -57,8 +57,8 @@ int can_overflow_cnt = 0;
 
 // ********************* interrupt safe queue *********************
 
-int can_pop(can_ring *q, CAN_FIFOMailBox_TypeDef *elem) {
-  int ret = 0;
+bool can_pop(can_ring *q, CAN_FIFOMailBox_TypeDef *elem) {
+  bool ret = 0;
```
geohot: false?

board/drivers/uart.h
```diff
@@ -14,7 +14,7 @@ typedef struct uart_ring {
 
 void uart_init(USART_TypeDef *u, int baud);
 
-int getc(uart_ring *q, char *elem);
+bool getc(uart_ring *q, char *elem);
 int putc(uart_ring *q, char elem);
```
geohot: bool?

board/main.c
```diff
@@ -219,7 +219,7 @@ int usb_cb_control_msg(USB_Setup_TypeDef *setup, uint8_t *resp, int hardwired) {
       // so it's blocked over wifi
       switch (setup->b.wValue.w) {
         case 0:
-          if (hardwired) {
+          if (hardwired != 0) {
```
geohot: Change hardwired to bool

board/drivers/can.h
rbiasini: zero automatically cats into false and non-zero into true and misra does not complain. It will be quite a bit of work to be consistent with bool assignments everywhere (I started it, then I gave up :) ).
I can change it for this function though.

board/main.c
rbiasini: yes, it will be a bit of refactor (hardwired is an input param in few functions), but it will be cleaner.

board/drivers/uart.h
rbiasini: seems like it returns 1 when it succeeds and zero when it does not.
```
int getc(uart_ring *q, char *elem) {
  int ret = 0;

  enter_critical_section();
  if (q->w_ptr_rx != q->r_ptr_rx) {
    if (elem != NULL) *elem = q->elems_rx[q->r_ptr_rx];
    q->r_ptr_rx = (q->r_ptr_rx + 1) % FIFO_SIZE;
    ret = 1;
  }
  exit_critical_section();

  return ret;
}

```

board/main.c
rbiasini: 97a11115bbec51be0b6d56b3ac3f8461d64633b7

PR #226, opened 2019-06-26
Misra 15.5: A function should have a single point of exit at the end

rbiasini:


board/drivers/can.h
```diff
@@ -130,29 +130,24 @@ void can_set_speed(uint8_t can_number) {
   CAN_TypeDef *CAN = CANIF_FROM_CAN_NUM(can_number);
   uint8_t bus_number = BUS_NUM_FROM_CAN_NUM(can_number);
 
-  while (true) {
```
rbiasini: does not seem like the while was needed.

geohot: yea, that's fine. can init may fail, but looping is probably the wrong choice anyway

board/main.c
```diff
@@ -153,11 +153,13 @@ int usb_cb_ep1_in(uint8_t *usbdata, int len, bool hardwired) {
...
 void usb_cb_ep2_out(uint8_t *usbdata, int len, bool hardwired) {
   UNUSED(hardwired);
-  if (len == 0) return;
-  uart_ring *ur = get_ring_by_number(usbdata[0]);
-  if (!ur) return;
-  if ((usbdata[0] < 2) || safety_tx_lin_hook(usbdata[0]-2, usbdata+1, len-1)) {
-    for (int i = 1; i < len; i++) while (!putc(ur, usbdata[i]));
+  if (len != 0) {
+    uart_ring *ur = get_ring_by_number(usbdata[0]);
+    if (ur != NULL) {
```
geohot: merge 156 and 158 to avoid deep indent?

board/power_saving.h
```diff
@@ -4,54 +4,56 @@
 int power_save_status = POWER_SAVE_STATUS_DISABLED;
 
 void power_save_enable(void) {
```
geohot: These should really be refactored into set_power_save_state(int state)

board/main.c
rbiasini: yep, I'll move `uart_ring *ur = get_ring_by_number(usbdata[0]);` up

board/power_saving.h
rbiasini: see 96fb376

rbiasini: well, let me refactor more. Reading the code, there is a lot of repeated stuff

rbiasini: See https://github.com/commaai/panda/pull/226/commits/ec3d0386c2f6e546cc2e5ee8cc855086e8afdbf6

PR #229, opened 2019-06-27
Misra 16_4: every switch statement shall have a default label

rbiasini:
Every switch statement shall have a default label


geohot:
What is 16.4?

rbiasini:
yeah, forgot to add the rule this time. See ^^

PR #230, opened 2019-06-28
Misra 15_7: if … else if constructs should be terminated with an else clause

rbiasini:


board/drivers/uart.h
```diff
@@ -289,6 +289,8 @@ void uart_init(USART_TypeDef *u, int baud) {
     NVIC_EnableIRQ(USART3_IRQn);
   } else if (u == UART5) {
     NVIC_EnableIRQ(UART5_IRQn);
+  } else {
+      // in this case, should we still execute what comes before the switch?
```
rbiasini: @geohot , unsure about this

geohot: Meh, probably not.

geohot:
Good to merge I think, the UART behavior stays the same.

PR #235, opened 2019-07-03
memxxx function changes to be Misra compliant

rbiasini:
11.8: A cast shall not remove any const or volatile qualification from the type pointed to by a pointer
17.8: A function parameter should not be modified

No impact on performance. Tested with:

```
#include <stdint.h>
#include "libc.h"

int main(void) {  
  uint8_t a1[60] = {1};
  uint8_t b1[60] = {2};
  for (int i; i < 1000000; i++) {
    //printf("%d\n", memcmp(a1, b1, 61));
    memset(a1, 1, 60);
    memcpy(a1, b1, 60);
    memcmp(a1, b1, 60);
    //printf("%d\n", memcmp(a1, b1, 61));
  }
  return 1;
}
```


board/libc.h
```diff
@@ -6,32 +6,36 @@ void delay(int a) {
...
-  unsigned int i;
-  for (i = 0; i < n; i++) {
-    *((uint8_t*)str) = c;
-    ++str;
+  uint8_t *s = str;
+  for (unsigned int i = 0; i < n; i++) {
+    *s = c;
+    s++;
   }
   return str;
```
rbiasini: To keep it the same, I should return `s` instead of `str`, but this is a more canonical return for memset I think.

geohot: Fine to change

PR #237, opened 2019-07-03
Misra 17.7: the value returned by a function having non-void return shall be used

rbiasini:


geohot:
This is not the intention of this. Will discuss

rbiasini:
ok, i'll merge the changes for the obvious ones (memset etc...), will change behavior for can_push return.

PR #244, opened 2019-07-04
Added Nissan safety

avolmensky:
Still a WIP.

Feedback would be appreciated. Nissan is angle type steering.


board/safety/safety_nissan.h
```diff
@@ -0,0 +1,159 @@
...
+  }
+  
+  if (bus == 1) {
+    if (addr == 0x1b6) {
+      int cruise_engaged = (to_push->RDHR >> 6) & 1;
+      if (cruise_engaged && !nissan_cruise_engaged_last) {
+        controls_allowed = 1;
+      } else if (!cruise_engaged) {
+        controls_allowed = 0;
+      }
```
rbiasini: missing else (misra)

board/safety/safety_nissan.h
```diff
@@ -0,0 +1,159 @@
...
+      int lowest_rt_angle = nissan_rt_angle_last - ((nissan_rt_angle_last > 0) ? rt_delta_angle_down : rt_delta_angle_up);
+
+      // Limit maximum steering angle at current speed
+      float maximum_angle = interpolate(NISSAN_LOOKUP_MAX_ANGLE, nissan_speed);
+
+      if (highest_rt_angle > maximum_angle) {
+        highest_rt_angle = maximum_angle;
+      } else if (lowest_rt_angle < -maximum_angle) {
+        lowest_rt_angle = -maximum_angle;
+      }
```
rbiasini: else missing

rbiasini:
Thanks. I'll give you more detailed feedback early next week. Seems like you are on the right path though. I think we should abstract some of the angle checks functions and re-utilize them on both Tesla and Nissan, since the logic is the same (I can do the refactor).

geohot:
Super happy to see this PR :)

avolmensky:
@geohot Bugsy will do the opendbc PR. He did pretty most of the DBC stuff

board/safety/safety_nissan.h
```diff
@@ -0,0 +1,159 @@
...
+  {5., .8, .15}};
+
+const struct lookup_t NISSAN_LOOKUP_ANGLE_RATE_DOWN = {
+  {2., 7., 17.},
+  {5., 3.5, .4}};
+
+// TODO: Need to work out the max angle at certain speeds
+const struct lookup_t NISSAN_LOOKUP_MAX_ANGLE = {
+    {2., 29., 38.},
+    {500., 500., 500.}};
```
rbiasini: need max angles. Maybe get them from `safety_tesla.h`?

board/safety/safety_nissan.h
```diff
@@ -0,0 +1,159 @@
...
+
+static void nissan_rx_hook(CAN_FIFOMailBox_TypeDef *to_push) {
+  int bus = GET_BUS(to_push);
+  int addr = GET_ADDR(to_push);
+
+  if (bus == 0) {
+    if (addr == 0x2) {
+      // Current steering angle
+      // Factor -0.1, little endian
+      nissan_angle_meas_new = to_signed(to_push->RDLR & 0xFFFF, 16) * -0.1;
```
rbiasini: we now have the convenient macros `GET_BYTE`. Would you mind using that instead?

board/safety.h
```diff
@@ -55,6 +56,7 @@ typedef struct {
 #define SAFETY_TESLA 8U
 #define SAFETY_CHRYSLER 9U
 #define SAFETY_SUBARU 10U
+#define SAFETY_NISSAN 11U
```
rbiasini: Nissan has now 14 as a reserved spot.

avolmensky:
This is getting pretty close to being completed. Open to comments/suggestions.

geohot:
Can you get this to pass the MISRA check?

board/safety/safety_nissan.h
```diff
@@ -0,0 +1,188 @@
...
+        if (cruise_engaged && !nissan_cruise_engaged_last) {
+          controls_allowed = 1;
+        }
+        if (!cruise_engaged) {
+          controls_allowed = 0;
+        }
+        nissan_cruise_engaged_last = cruise_engaged;
+      }
+    }
+  }
```
rbiasini: Missing explicit disable checks on:
- gas rising edge 
- brake pedal pressed when speed > 0

Even though these checks are missing/incomplete in few other safety files, it would be nice to merge this one with those checks already included.

rbiasini:
@avolmensky are you planning to add the regression tests too? See `panda/tests/safety/test_xxx.py`

avolmensky:
@rbiasini I'll write regression tests in the next few days

board/safety/safety_nissan.h
```diff
@@ -0,0 +1,212 @@
...
+}
+
+
+static int nissan_fwd_hook(int bus_num, CAN_FIFOMailBox_TypeDef *to_fwd) {
+  int bus_fwd = -1;
+  int addr = GET_ADDR(to_fwd);
+
+  if (bus_num == 0) {
+    // 0x20b is CruiseThrottle
+    int block_msg = (addr == 0x20b);
```
rbiasini: why do you need to block the cruise throttle from bus 0 to 2?

board/safety/safety_nissan.h
```diff
@@ -0,0 +1,212 @@
...
+  {540., 120., 23.}};
+
+const int NISSAN_DEG_TO_CAN = 100; 
+
+const AddrBus NISSAN_TX_MSGS[] = {{0x169, 0}, {0x20b, 2}};
+
+AddrCheckStruct nissan_rx_checks[] = {
+  {.addr = {0x2}, .bus = 0, .expected_timestep = 100000U},
+  {.addr = {0x29a}, .bus = 0, .expected_timestep = 50000U},
+  {.addr = {0x20b}, .bus = 0, .expected_timestep = 50000U},
```
rbiasini: seems like 0x20b is unused in rx?

board/safety/safety_nissan.h
```diff
@@ -0,0 +1,212 @@
...
+        if (!cruise_engaged) {
+          controls_allowed = 0;
+        }
+        nissan_cruise_engaged_last = cruise_engaged;
+      }
+
+      // exit controls on rising edge of brake press if speed > 0
+      if (addr == 0x454) {
+        int brake = (GET_BYTE(to_push, 2) & 0x80);
+        if ((brake > 0) && (nissan_brake_prev == 0) && (nissan_speed > 0.)) {
```
rbiasini: bug. It should be
```if ((brake > 0) && (nissan_brake_prev == 0) || (nissan_speed > 0.))```
as you want o disengage no matter what of speed > 0.

the safety test should highlight this bug.

File: board/safety/safety_nissan.h [hunk 10]
```diff
@@ -0,0 +1,212 @@
...
+    int desired_angle = ((GET_BYTE(to_send, 0) << 10) | (GET_BYTE(to_send, 1) << 2) | ((GET_BYTE(to_send, 2) >> 6) & 0x3));
+    bool lka_active = (GET_BYTE(to_send, 6) >> 4) & 1;
+
+    // offeset 1310 * NISSAN_DEG_TO_CAN
+    desired_angle =  desired_angle - 131000;
+    bool violation = 0;
+
+    if (controls_allowed && lka_active) {
+      // add 1 to not false trigger the violation
+      float delta_angle_float;
```
rbiasini: Not for now, but this code is the same as the one used in other safety files (toyota_ipas and maybe Tesla?), so it should be abstracted. I'll look into this.

board/safety/safety_nissan.h
avolmensky: This is probably a bad message name. This message contains cruise control button status and gas pedal position.

This message is generated on the EON, it basically passes through all button presses unless OP wants to cancel ACC/LKAS then it will set the cancel button signal to true.

rbiasini: as discussed, instead of blocking the message, it's better to just spam this message and put a safety tx check that the only allowed button is "cancel"

rbiasini:
@avolmensky , great job. It looks pretty complete to me. The openpilot side of the port might end up requiring changes to the safety code, but no reason to not merge this for now.

PR #248, opened 2019-07-09
fix bootstub flashing

dek3rr:
dfu.recover() causes this failure. 
main bootstub flash defaults to soft flasher because it can't validate length with 0x8004000 erased.
soft flasher then gets stuck green flashing because 0x8004000 is erased and therefore nowhere to go.

### DIFF
--- a/board/bootstub.c
+++ b/board/bootstub.c
@@ -50,8 +50,6 @@ void fail(void) {
 // know where to sig check
 extern void *_app_start[];
 
-// FIXME: sometimes your panda will fail flashing and will quickly blink a single Green LED
-// BOUNTY: $200 coupon on shop.comma.ai or $100 check.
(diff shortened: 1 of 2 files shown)


pd0wm:
Are you sure that is what's going on?

I was able to replicate the flashing issue by putting CAN traffic on bus 1. I think the DFU mode gets confused when there is activity on those pins on the MCU. The solution for that problem would be to turn off the CAN transceivers before going into DFU mode.

dekerr:
I'm confident that this is at least part of the problem. I can't comment on CAN traffic right now, but everything I said earlier held true in my testing. DFU shouldn't be erasing anything that it's not immediately writing to and that avoids the soft flasher which never gets interrupted from its blinking loop.

PR #249, opened 2019-07-10
Propagate can_push errors

rbiasini:
Misra 17.7: The value returned by a function having non-void return type shall be used

### DIFF
--- a/board/drivers/can.h
+++ b/board/drivers/can.h
@@ -356,7 +356,7 @@ void can_send(CAN_FIFOMailBox_TypeDef *to_push, uint8_t bus_number) {
         // TODO: why uint8 bro? only int8?
         gmlan_send_errs += !bitbang_gmlan(to_push);
       } else {
-        can_fwd_errs += can_push(can_queues[bus_number], to_push);
+        can_fwd_errs += !can_push(can_queues[bus_number], to_push);
         process_can(CAN_NUM_FROM_BUS_NUM(bus_number));
       }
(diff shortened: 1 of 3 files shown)


board/drivers/can.h
```diff
@@ -351,9 +354,9 @@ void can_send(CAN_FIFOMailBox_TypeDef *to_push, uint8_t bus_number) {
       to_push->RDTR &= 0xF;
       if ((bus_number == 3U) && (can_num_lookup[3] == 0xFFU)) {
         // TODO: why uint8 bro? only int8?
-        bitbang_gmlan(to_push);
+        gmlan_send_errs += !bitbang_gmlan(to_push);
       } else {
-        can_push(can_queues[bus_number], to_push);
+        can_fwd_errs += can_push(can_queues[bus_number], to_push);
```
geohot: forgot a not

board/main.c
```diff
@@ -101,11 +101,12 @@ int get_health_pkt(void *dat) {
   struct __attribute__((packed)) {
     uint32_t voltage_pkt;
     uint32_t current_pkt;
-    uint8_t started_pkt;
-    uint8_t controls_allowed_pkt;
-    uint8_t gas_interceptor_detected_pkt;
-    uint8_t started_signal_detected_pkt;
-    uint8_t started_alt_pkt;
+    bool started_pkt;
```
geohot: Don't make these bools

PR #251, opened 2019-07-12
Update core to CMSIS 5.6 release 

dekerr:
CMSIS 5 is largely refactored and cleaned up to be more generic, more readable, and prevent a few compiler bugs. CMSIS 4 is now legacy code. 
It also conforms more closely to MISRA C: 2012 and other standards outlined here:
https://arm-software.github.io/CMSIS_5/General/html/index.html

repository available here:
https://github.com/ARM-software/CMSIS_5/tree/master


rbiasini:
I'll copy this changes in an internal branch, so we can let it run the HIL tests.

rbiasini:
tests passed, also verified files match 1:1 with CMSIS_5 repo. Thanks.

PR #254, opened 2019-07-16
Black

rbiasini:


board/board.h
```diff
@@ -0,0 +1,87 @@
...
+// ///// Board definition ///// //
+#ifdef PANDA
+  #include "boards/white.h"
+  #include "boards/grey.h"
+  #include "boards/black.h"
+#else
+  #include "boards/pedal.h"
+#endif
+
+// ///// Board detection ///// //
```
rbiasini: @robbederks , no need to make  GREY val 1. This is fine, I'll change boardd. 
Also, there should be a comment here to say that this enum shall match the one in cereal/log.capnp.

Since the pedal is not a considered a panda, I recommend the following names:
```
#define HW_TYPE_UNSUPPORTED 0
#define HW_TYPE_WHITE_PANDA 1
#define HW_TYPE_GREY_PANDA 2
#define HW_TYPE_BLACK_PANDA 3
#define HW_TYPE_PEDAL 4
```

robbederks: Ok, great, I'll change it right away!

geohot:
Can we fix tests before review?

board/drivers/harness.h
```diff
@@ -0,0 +1,141 @@
...
+  uint8_t pin_relay_normal;
+  uint8_t pin_relay_flipped;
+  uint8_t adc_channel_SBU1;
+  uint8_t adc_channel_SBU2;
+};
+
+// this function will be the API for tici
+void set_intercept_relay(bool intercept) {
+  if (car_harness_status != HARNESS_STATUS_NC) {
+    if(intercept)
```
geohot: Spacing and braces

board/drivers/harness.h
```diff
@@ -0,0 +1,141 @@
...
+  }
+  #endif
+
+  return ret;
+}
+
+void harness_init(void) {
+  // chilling for power to be stable (we have interrupts)
+  current_board->set_led(LED_RED, 1);
+  delay(5000000);
```
geohot: Do we still need this delay? And do we need to try three times? A lot of this junk was because the LIN transceiver was really bad.

board/main.c
```diff
@@ -482,6 +512,12 @@ int usb_cb_control_msg(USB_Setup_TypeDef *setup, uint8_t *resp, bool hardwired)
         }
         break;
       }
+    // **** 0xf3: Set harness relay to (not) intercept (black panda only).
+    case 0xf3:
```
geohot: Do we need this? Why can't we just use the safety mode?

geohot:
Overall looks very good!

tests/black_loopback_test.py
```diff
@@ -0,0 +1,146 @@
...
+import random
+import argparse
+
+from hexdump import hexdump
+from itertools import permutations
+
+sys.path.append(os.path.join(os.path.dirname(os.path.realpath(__file__)), ".."))
+from panda import Panda
+
+def set_intercept(p, intercept):
```
geohot: Use the API for this?

board/power_saving.h
```diff
@@ -10,33 +10,33 @@ void set_power_save_state(int state) {
     bool enable = false;
     if (state == POWER_SAVE_STATUS_ENABLED) {
       puts("enable power savings\n");
-      if (is_grey_panda) {
+      if ((hw_type == HW_TYPE_GREY_PANDA) || (hw_type == HW_TYPE_BLACK_PANDA)) {
```
rbiasini: probably worth defining
` 
has_gps_receiver = (hw_type == HW_TYPE_GREY_PANDA) || (hw_type == HW_TYPE_BLACK_PANDA)
`

rbiasini: indeed, `is_grey_panda` should be totally removed from the code.

board/main.c
```diff
@@ -710,23 +687,24 @@ int main(void) {
...
   }
   // only enter power save after the first cycle
-  /*if (is_gpio_started()) {
+  /*if (current_board->check_ignition()) {
     set_power_save_state(POWER_SAVE_STATUS_ENABLED);
   }*/
-  // interrupt on started line
-  started_interrupt_init();
+
+  if(hw_type != HW_TYPE_BLACK_PANDA){
```
rbiasini: spaces

board/drivers/can.h
```diff
@@ -326,9 +366,9 @@ void can_rx(uint8_t can_number) {
 
     safety_rx_hook(&to_push);
 
-    set_led(LED_BLUE, 1);
+    current_board->set_led(LED_BLUE, true);
     can_send_errs += !can_push(&can_rx_q, &to_push);
-
+
```
rbiasini: spaces

board/board.h
```diff
@@ -0,0 +1,57 @@
...
+      hw_type = HW_TYPE_WHITE_PANDA;
+      current_board = &board_white;
+    } else if(detect_with_pull(GPIOA, 13, PULL_DOWN)) { // Rev AB deprecated, so no pullup means black. In REV C, A13 is pulled up to 5V with a 10K
+      hw_type = HW_TYPE_GREY_PANDA;
+      current_board = &board_grey;
+    } else {
+      hw_type = HW_TYPE_BLACK_PANDA;
+      current_board = &board_black;
+    }
+  #else
```
rbiasini: can be simplified in:
```
#elseif PEDAL
  hw_type = HW_TYPE_PEDAL;
  current_board = &board_pedal;
#else 
    hw_type = HW_TYPE_UNKNOWN;
    puts("Hardware type is UNKNOWN!\n");
 #endif

File: board/main.c [hunk 10]
```diff
@@ -258,40 +277,51 @@ int usb_cb_control_msg(USB_Setup_TypeDef *setup, uint8_t *resp, bool hardwired)
...
-    // **** 0xdb: set GMLAN multiplexing mode
+    // **** 0xdb: set GMLAN (white/grey) or OBD CAN (black) multiplexing mode
     case 0xdb:
-      if (setup->b.wValue.w == 1U) {
-        // GMLAN ON
-        if (setup->b.wIndex.w == 1U) {
-          can_set_gmlan(1);
-        } else if (setup->b.wIndex.w == 2U) {
-          can_set_gmlan(2);
+      if(hw_type != HW_TYPE_BLACK_PANDA){
```
rbiasini: minor: I would switch the if and else so that:
```
if (hw_type == HW_TYPE_BLACK_PANDA) {
...
} else {
...
}
```

board/board.h
rbiasini: Also, if the type is UNKNOWN, does it make sense to keep executing code? i.e. are we sure we won't have unexpected consequences?

board/main.c
rbiasini: currently, the safety mode persists until panda is disconnected from EON. We'll have to refactor things a bit.

board/board.h
robbederks: If the type is UNKNOWN, it stops executing in the main loop :)
```
// check for non-supported board types
  if(hw_type == HW_TYPE_UNKNOWN){
    puts("Unsupported board type\n");
    while (1) { /* hang */ }
  }
```

board/drivers/harness.h
robbederks: Fixed!

board/drivers/harness.h
robbederks: True. Simplified the code a bit, now just waits for about a sec after power-on and then does the orientation detection once. Seems to be stable...

board/main.c
robbederks: Removed this part and set the intercept to be on when not in NOOUTPUT or ELM mode.

tests/black_loopback_test.py
robbederks: Removed it. Not needed anymore with new intercept logic.

board/power_saving.h
robbederks: I agree. Added it as a function to board.h

board/main.c
robbederks: Fixed!

board/drivers/can.h
robbederks: Not sure what you mean here. Might already be fixed?

board/main.c
robbederks: Fixed!

board/board.h
rbiasini: Nice!

board/drivers/can.h
rbiasini: just 4 trailing spaces on the empty line.

geohot:
I don't like the heartbeat logic. If EON fails, the safety mode should stay as before to cause the dashboard warning lights.

rbiasini:
@geohot that's why it waits 5 sec before switching. Otherwise, how do you go back to stock?

PR #264, opened 2019-08-31
Improved GMLAN support. Minor CAN bus bug fixes and improvements

mjuhanne:
* Allow user to set GMLAN Single Wire CAN operation modes (Normal, High Voltage Wake Up, High Speed, Sleep)
* Change SJW=2 for GMLAN Single Wire CAN
* Add non-blocking CAN read function
* Log can_send_errs / gmland_send_errs properly
* Fix lost_rx errors counted as tx_errors
* More verbose logging in case of CAN tx/arbitration lost error

### DIFF
--- a/board/drivers/can.h
+++ b/board/drivers/can.h
@@ -132,7 +132,7 @@ uint32_t can_speed[] = {5000, 5000, 5000, 333};
 #define CAN_NAME_FROM_CANIF(CAN) ((CAN)==CAN1 ? "CAN1" : ((CAN) == CAN2 ? "CAN2" : "CAN3"))
 #define BUS_NUM_FROM_CAN_NUM(num) (bus_lookup[num])
 #define CAN_NUM_FROM_BUS_NUM(num) (can_num_lookup[num])
-#define ID_FROM_IR(rir) ((rir) & 0x4U ? (rir) >> 3U : (rir) >> 21U)
+#define ID_FROM_IR(rir) (((rir) & 0x4U) ? ((rir) >> 3U) : ((rir) >> 21U))
 
 void process_can(uint8_t can_number);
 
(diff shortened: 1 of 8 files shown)


mjuhanne:
Ok, is this a false positive warning by MISRA? 
```
Failed! found Misra violations in panda code:
[board/drivers/can.h:377] (style) misra violation (use --rule-texts=<file> to get proper output) (Undefined) [misra-c2012-15.7]
```
_**Misra c 2012 15.7: All if . . else if constructs shall be terminated with an else statement**_

But there is no "else if" statement there...

board/drivers/can.h
```diff
@@ -387,16 +463,21 @@ void CAN3_RX0_IRQHandler(void) { can_rx(2); }
...
+  bool res;
   if (safety_tx_hook(to_push) != 0) {
     if (bus_number < BUS_MAX) {
       // add CAN packet to send queue
       // bus number isn't passed through
       to_push->RDTR &= 0xF;
       if ((bus_number == 3U) && (can_num_lookup[3] == 0xFFU)) {
         // TODO: why uint8 bro? only int8?
-        gmlan_send_errs += !bitbang_gmlan(to_push);
+        res = bitbang_gmlan(to_push);
```
geohot: What was wrong with += before?

geohot:
The MISRA thing is not a false positive. If you want to use an if statement, you need to handle the else case as well.

board/drivers/can.h
rbiasini: Bool added to a int. Misra does not like it. It's already fixed in master, just need to resolve conflicts.

board/drivers/can.h
```diff
@@ -326,6 +346,60 @@ void process_can(uint8_t can_number) {
         CAN->sTxMailBox[0].TDTR = to_send.RDTR;
         CAN->sTxMailBox[0].TIR = to_send.RIR;
       }
+    } else {
+      if ((CAN->TSR & CAN_TSR_TERR0) == CAN_TSR_TERR0) {
```
rbiasini: This is the same as "else if", which is why misra complains.

geohot:
If you still want this, please rebase and resubmit

PR #266, opened 2019-09-04
GM: Add passive safety mode

vntarasov:
Redo of https://github.com/commaai/openpilot/pull/487: adding a separate, passive version of GM safety model, to allow EONs to be used as dashcams on stock GM cars (with ASCM present), or in chffrplus passive mode.

To be used with https://github.com/vntarasov/openpilot/tree/gm-passive2, which is to be updated with a panda repo sync.


rbiasini:
ok, makes sense. I think a clean way to handle this safety mode is to add safetyModelPassive in `interface.py` for each car. For GM 
```
safetyModel = car.CarParams.SafetyModel.gm
safetyModelPassive = car.CarParams.SafetyModel.gmPassive
```
While for most of the other cars it will look like:
```
safetyModel = car.CarParams.SafetyModel.<carBrand>
safetyModelPassive = car.CarParams.SafetyModel.elm327
```

board/safety/safety_gm_passive.h
```diff
@@ -0,0 +1,22 @@
...
+
+  if ((addr == 0x1F1) && (bus_number == 0)) {
+    bool ign = (GET_BYTE(to_push, 0) & 0x20) != 0;
+    gm_ignition_started = ign;
+  }
+}
+
+const safety_hooks gm_passive_hooks = {
+  .init = gm_init,
+  .rx = gm_passive_rx_hook,
```
rbiasini: I think it's ok to just use `gm_rx_hook` here. As long as nooutput_tx_hook and default_fwd_hook are set, there is really no risk.

vntarasov: @rbiasini, actually `gm_rx_hook` modifies `contorls_allowed`, and that goes into both health packet, and controls green LED of the panda. I thought it would be misleading to have a wrong value for that, hence copy-pasted code for determining the ignition.

rbiasini:
makes sense

vntarasov:
@rbiasini  So... can panda change be merged, or more changes are needed?

rbiasini:
pushed changes to openpilot as well, so that this safety model is used. Will be in next release.

PR #267, opened 2019-09-05
fix canflash for pedal

BogGyver:


rbiasini:
Yeah, as discussed in Discord, this is correct. Thanks.

PR #268, opened 2019-09-08
honda bosch longitudinal safety

gregjhogan:
TODO:
- [x] brake limit should be -350 when moving and -400 when stopped (or can we use -350 for standstill?) -350 seems fine
- [x] standstill bit dangerous to send when moving? seems fine, probably smooths out braking
- [x] openpilot should set `long_controls_allowed = false` when running stock bosch? (no longer exists)
- [x] always allow sending ACC messages? no
- [x] tests
- [x] split code out into new safety mode

### DIFF
--- a/board/safety/safety_honda.h
+++ b/board/safety/safety_honda.h
@@ -9,6 +9,8 @@
 const CanMsg HONDA_N_TX_MSGS[] = {{0xE4, 0, 5}, {0x194, 0, 4}, {0x1FA, 0, 8}, {0x200, 0, 6}, {0x30C, 0, 8}, {0x33D, 0, 5}};
 const CanMsg HONDA_BG_TX_MSGS[] = {{0xE4, 2, 5}, {0xE5, 2, 8}, {0x296, 0, 4}, {0x33D, 2, 5}};  // Bosch Giraffe
 const CanMsg HONDA_BH_TX_MSGS[] = {{0xE4, 0, 5}, {0xE5, 0, 8}, {0x296, 1, 4}, {0x33D, 0, 5}};  // Bosch Harness
+const CanMsg HONDA_BG_LONG_TX_MSGS[] = {{0xE4, 0, 5}, {0x1DF, 0, 8}, {0x1EF, 0, 8}, {0x1FA, 0, 8}, {0x30C, 0, 8}, {0x33D, 0, 5}, {0x39F, 0, 8}, {0x18DAB0F1, 0, 8}};  // Bosch Giraffe w/ gas and brakes
+const CanMsg HONDA_BH_LONG_TX_MSGS[] = {{0xE4, 1, 5}, {0x1DF, 1, 8}, {0x1EF, 1, 8}, {0x1FA, 1, 8}, {0x30C, 1, 8}, {0x33D, 1, 5}, {0x39F, 1, 8}, {0x18DAB0F1, 1, 8}};  // Bosch Harness w/ gas and brakes
 
 // Roughly calculated using the offsets in openpilot +5%:
 // In openpilot: ((gas1_norm + gas2_norm)/2) > 15
@@ -18,12 +20,16 @@ const CanMsg HONDA_BH_TX_MSGS[] = {{0xE4, 0, 5}, {0xE5, 0, 8}, {0x296, 1, 4}, {0
 // In this safety: ((gas1 + (gas2/2))/2) > THRESHOLD
 const int HONDA_GAS_INTERCEPTOR_THRESHOLD = 344;
 #define HONDA_GET_INTERCEPTOR(msg) (((GET_BYTE((msg), 0) << 8) + GET_BYTE((msg), 1) + ((GET_BYTE((msg), 2) << 8) + GET_BYTE((msg), 3)) / 2 ) / 2) // avg between 2 tracks
+const int HONDA_BOSCH_NO_GAS_VALUE = -30000; // value sent when not requesting gas
+const int HONDA_BOSCH_GAS_MAX = 2000;
+const int HONDA_BOSCH_ACCEL_MIN = -350; // max braking == -3.5m/s2
 
 // Nidec and Bosch giraffe have pt on bus 0
 AddrCheckStruct honda_rx_checks[] = {
   {.msg = {{0x1A6, 0, 8},  {0x296, 0, 4}}, .check_checksum = true, .max_counter = 3U, .expected_timestep = 40000U},
   {.msg = {{0x158, 0, 8}}, .check_checksum = true, .max_counter = 3U, .expected_timestep = 10000U},
   {.msg = {{0x17C, 0, 8}}, .check_checksum = true, .max_counter = 3U, .expected_timestep = 10000U},
+  {.msg = {{0x1BE, 0, 3}}, .check_checksum = true, .max_counter = 3U, .expected_timestep = 10000U},
 };
 const int HONDA_RX_CHECKS_LEN = sizeof(honda_rx_checks) / sizeof(honda_rx_checks[0]);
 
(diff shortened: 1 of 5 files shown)


geohot:
How do you enable this? Do you need to unplug the radar?

gregjhogan:
There are several options:
- unplug radar
- enter diagnostic session and send communication control command to disable tx/rx, all communication, suppress response
  (address `0x18DAB0F1` send `\x02\x10\x03` then `\x03\x28\x83\x03`)
  something is still sent indicating a fault state, but since non-fault is dominant you can override by sending everything is OK (if you send it fast enough)
- enter programming session

board/safety/safety_honda.h
```diff
@@ -144,14 +162,24 @@ static int honda_tx_hook(CAN_FIFOMailBox_TypeDef *to_send) {
...
         tx = 0;
       }
     }
   }
+  // GAS: safety check (bosch)
+  if (addr == 0x1DF) {
+    int gas = (GET_BYTE(to_send, 0) << 8) | GET_BYTE(to_send, 1);
+    gas = to_signed(gas, 16);
+    if (!current_controls_allowed || !long_controls_allowed) {
+      if (gas != -30000) {
```
rbiasini: I would parametrize these values

board/safety/safety_honda.h
```diff
@@ -133,6 +135,22 @@ static int honda_tx_hook(CAN_FIFOMailBox_TypeDef *to_send) {
       tx = 0;
     }
   }
+  // BRAKE: safety check (bosch)
+  if (addr == 0x1DF) {
```
rbiasini: specify bus, depending on panda type

board/safety/safety_honda.h
```diff
@@ -144,14 +162,24 @@ static int honda_tx_hook(CAN_FIFOMailBox_TypeDef *to_send) {
...
+  // GAS: safety check (nidec)
   if (addr == 0x200) {
     if (!current_controls_allowed || !long_controls_allowed) {
       if (GET_BYTE(to_send, 0) || GET_BYTE(to_send, 1)) {
         tx = 0;
       }
     }
   }
+  // GAS: safety check (bosch)
+  if (addr == 0x1DF) {
```
rbiasini: can be merged with brake check. Is the same message

board/safety/safety_honda.h
gregjhogan: fixed

board/safety/safety_honda.h
gregjhogan: fixed

board/safety/safety_honda.h
gregjhogan: fixed

board/safety/safety_honda.h
```diff
@@ -7,9 +7,13 @@
 //      brake rising edge
 //      brake > 0mph
 const AddrBus HONDA_N_TX_MSGS[] = {{0xE4, 0}, {0x194, 0}, {0x1FA, 0}, {0x200, 0}, {0x30C, 0}, {0x33D, 0}};
-const AddrBus HONDA_BG_TX_MSGS[] = {{0xE4, 2}, {0x296, 0}, {0x33D, 2}};  // Bosch Giraffe
-const AddrBus HONDA_BH_TX_MSGS[] = {{0xE4, 0}, {0x296, 1}, {0x33D, 0}};  // Bosch Harness
+// TODO: only allow TX messages on bus 0 if OP longitudinal control?
+const AddrBus HONDA_BG_TX_MSGS[] = {{0xE4, 0}, {0xE4, 2}, {0x1DF, 0}, {0x1EF, 0}, {0x296, 0}, {0x30C, 0}, {0x33D, 0}, {0x33D, 2}, {0x39F, 0}};  // Bosch Giraffe
+// TODO: only allow TX messages on bus 1 if OP longitudinal control?
+const AddrBus HONDA_BH_TX_MSGS[] = {{0xE4, 0}, {0xE4, 1}, {0x1DF, 1}, {0x1EF, 1}, {0x296, 1}, {0x30C, 1}, {0x33D, 0}, {0x33D, 1}, {0x39F, 1}};  // Bosch Harness
```
gregjhogan: @rbiasini is it OK to always allow steering/gas/brake messages on powertrain bus?

gregjhogan:
I think this is ready to merge now.  Drove this branch with OP master without any issues.

rbiasini:
I know we discussed this the other night: I think it's fine (from a safety standpoint) to allow a message to be sent on both buses, but since this use case (long control for bosch) is for now way less frequent than the LaC only, I recommend the following:

* Not worth supporting the giraffe config IMO. It will probably be deprecated at a certain point, not worth adding features on top of it.
* create a separate safety mode instead of overloading hondaBoschHarness. It could be called hondaBoschNoRadar or similar, and set a variable (`honda_bosch_no_radar=true` ?) so that in the tx_hook you can decide which checks apply. 
* Have a separate safety test class, which inherits from honda_bosch_harness and runs extra checks for gas and brakes.

Sorry, I would have recommended these changes during the first review, but the safety architecture changed a bit recently and this is the way I would do it now.

board/safety/safety_honda.h
```diff
@@ -7,9 +7,12 @@
 //      brake rising edge
 //      brake > 0mph
 const AddrBus HONDA_N_TX_MSGS[] = {{0xE4, 0}, {0x194, 0}, {0x1FA, 0}, {0x200, 0}, {0x30C, 0}, {0x33D, 0}};
-const AddrBus HONDA_BG_TX_MSGS[] = {{0xE4, 2}, {0x296, 0}, {0x33D, 2}};  // Bosch Giraffe
-const AddrBus HONDA_BH_TX_MSGS[] = {{0xE4, 0}, {0x296, 1}, {0x33D, 0}};  // Bosch Harness
+const AddrBus HONDA_BG_TX_MSGS[] = {{0xE4, 0}, {0xE4, 2}, {0x1DF, 0}, {0x1EF, 0}, {0x296, 0}, {0x30C, 0}, {0x33D, 0}, {0x33D, 2}, {0x39F, 0}};  // Bosch Giraffe
+const AddrBus HONDA_BH_TX_MSGS[] = {{0xE4, 0}, {0xE4, 1}, {0x1DF, 1}, {0x1EF, 1}, {0x296, 1}, {0x30C, 1}, {0x33D, 0}, {0x33D, 1}, {0x39F, 1}};  // Bosch Harness
```
rbiasini: hesitant to allow gas and brake messages to be sent for the normal case where only lateral control is being executed. Better to define a separate TX_MSGS list like:
```
HONDA_BG_TX_MSGS[] = {{0xE4, 1}, {0x1DF, 1}, ...};  // Bosch Harness with radar disconnected
```
And use it in its own safety mode.

gregjhogan:
updated to enable bosch longitudinal control messages based on a param passed when setting safety mode, should be good to merge once build is green

@pd0wm please review - TX whitelist is now broke out (gas/brake only allowed when explicitly enabled)
@quillford please review - re-organized tests in an attempt to minimize code duplication

also noticed that there was no checksum check on alt brake message, so I fixed that, too

board/safety.h
```diff
@@ -249,6 +249,10 @@ int to_signed(int d, int bits) {
   return d_signed;
 }
 
+bool get_param_flag(int16_t param, uint16_t mask) {
```
quillford, line 252: I don't think this really needs a function, but if we keep it, it should probably be more generic like `get_flag` so it can be used for the unsafe modes too.

board/safety/safety_honda.h
```diff
@@ -211,8 +226,9 @@ static int honda_tx_hook(CAN_FIFOMailBox_TypeDef *to_send) {
     pedal_pressed = pedal_pressed || gas_pressed_prev || (gas_interceptor_prev > HONDA_GAS_INTERCEPTOR_THRESHOLD);
   }
   bool current_controls_allowed = controls_allowed && !(pedal_pressed);
+  int bus_pt = (honda_hw == HONDA_BH_HW)? 1 : 0;
 
-  // BRAKE: safety check
+  // BRAKE: safety check (nidec)
   if ((addr == 0x1FA) && (bus == 0)) {
```
quillford, line 232: bus == bus_pt?

board/safety.h
gregjhogan: due to misra checks this requires a cast from `int16_t` to `uint16_t` (see latest commit) and the fact that it uses int16 in general makes it somewhat specific to safety mode param.  Should I move it and re-name it to `get_int16_flag()` or something?

board/safety/safety_honda.h
gregjhogan: yeah, I can update

board/safety/safety_honda.h
```diff
@@ -228,6 +244,31 @@ static int honda_tx_hook(CAN_FIFOMailBox_TypeDef *to_send) {
     }
   }
 
+  // BRAKE/GAS: safety check (bosch)
+  if ((addr == 0x1DF) && (bus == bus_pt)) {
+    int accel = (GET_BYTE(to_send, 3) << 3) + ((GET_BYTE(to_send, 4) >> 5) & 0x7);
```
adeebshihadeh, line 249: or instead of plus

board/safety.h
adeebshihadeh: hmm, then maybe just make it a macro?

adeebshihadeh:
LGTM, nice catch on the missing checksum check

board/safety.h
gregjhogan: good idea, the macro can abstract away the type

board/safety/safety_honda.h
gregjhogan: done

gregjhogan:
changes complete, good to merge?

gregjhogan:
@pd0wm rebase complete and ready to merge if no one sees any additional issues

gregjhogan:
removing 0x1BE brake message RX check because adding it means that message is required (breaks cars that do not use the lat brake message)

PR #283, opened 2019-10-01
UART instability fix with high interrupt load

robbederks:
The bug can be triggered reliably by creating a high interrupt load on the panda (e.g. spamming the CAN buses or adding an artificial delay in one of the interrupts).
This fix clears the UART errors reliably. It may still lose a char when it happens, but does not lock up the UART RX anymore.


robbederks:
Do not merge yet. Still a bug in this code (LED is not blinking anymore).

robbederks:
Catching the UART errors did not fix it. I made a mistake such that the interrupts were called continuously, restarting the locked up DMA.
Rewrote the DMA code now such that it does not rely on the interrupts any more to reset it's operation (dump directly in the ring buffer). This also reduces the interrupt load on the STM.

PR #284, opened 2019-10-02
Stop charge

rbiasini:


board/main.c
```diff
@@ -747,6 +738,13 @@ int main(void) {
   uint64_t cnt = 0;
 
   for (cnt=0;;cnt++) {
+    #ifdef EON
+    // set CDP mode on rising edge of ignition to start charging EON again in case we turned charging off
+    if (current_board->check_ignition() && !ignition_prev) {
+      current_board->set_usb_power_mode(USB_POWER_CDP);
+    }
+    ignition_prev = current_board->check_ignition();
+    #endif
```
robbederks: Wouldn't it be a lot cleaner (and more efficient) to put this in the `started_interrupt_handler`?
And then just do:
```
if (current_board->check_ignition()) {
  current_board->set_usb_power_mode(USB_POWER_CDP);
}
```

rbiasini: yep

PR #285, opened 2019-10-02
match safety enum in cereal

rbiasini:


board/safety.h
```diff
@@ -45,42 +45,43 @@ typedef struct {
...
-#define SAFETY_MAZDA 12U
-#define SAFETY_GM_ASCM 0x1334U
-#define SAFETY_TOYOTA_IPAS 0x1335U
-#define SAFETY_ALLOUTPUT 0x1337U
-#define SAFETY_ELM327 0xE327U
+#define SAFETY_TESLA 10U
+#define SAFETY_SUBARU 11U
+#define SAFETY_GM_PASSIVE 12U
+#define SAFETY_MAZDA 13U
+#define SAFETY_TOYOTA_IPAS 16U
```
jyoung8607: Might we reserve the cereal-matching VW and Nissan safety numbers while you're in the neighborhood?

#define SAFETY_NISSAN 14U
#define SAFETY_VOLKSWAGEN 15U

A real Panda safety PR will be coming in the next few days, once I fix the last issue with VW torque rate limiting. Just tired of getting bumped off our safety number, this is our third bump. :)

rbiasini: oh, no worries, it will not happen again. That's why 14 and 15 are not in there. 
The enumeration now matches what's in cereal. 
Just add:
```
#define SAFETY_VOLKSWAGEN 15U
```

whithin your PR

PR #287, opened 2019-10-03
Made setting of NOOUTPUT on no heartbeat more efficient

robbederks:


board/safety.h
```diff
@@ -16,6 +16,7 @@
 #include "safety/safety_mazda.h"
 #include "safety/safety_elm327.h"
 
+uint16_t current_safety_mode = 0U;
```
rbiasini: would be better to use SAFETY_NOOUTPUT

rbiasini:
merged master to see how jenkins test goes

PR #289, opened 2019-10-04
Fix usb_power_mode missing initialization

rbiasini:


board/boards/black.h
```diff
@@ -183,6 +183,9 @@ void black_init(void) {
 
   // init multiplexer
   can_set_obd(car_harness_status, false);
+
+  // init usb power mdoe
```
geohot: typo in comment

board/boards/white.h
```diff
@@ -243,8 +243,9 @@ void white_init(void) {
 
   // on PC, set USB power mode to CLIENT
 #ifndef EON
-  white_set_usb_power_mode(USB_POWER_CLIENT);
+  usb_power_mode = USB_POWER_CLIENT;
```
geohot: So I wouldn't do it like this. I'd init to NONE and call white_set_usb_power_mode for CLIENT on PC, and CDP on EON.

white_set_usb_power_mode sets usb_power_mode, it shouldn't be set elsewhere!

rbiasini: ok, that forces `set_usb_power_mode` to be called. 
Change done. Panda and cereal need to be bumped at the same time in master after this change.

geohot:
lgtm

PR #291, opened 2019-10-05
Fixed USB power mode on black

robbederks:


board/bootstub.c
```diff
@@ -68,8 +69,6 @@ int main(void) {
   detect_configuration();
   detect_board_type();
 
-  current_board->set_usb_power_mode(USB_POWER_CLIENT);
```
rbiasini: Not sure why this was here.... Sure we don't want it for white/grey?

robbederks: I'll check... Flashed fine in standalone mode, but can also test quickly when it is powered with 12V

robbederks: Seems to work fine on both white and black in all scenarios. We just don't want it to do this any more on black because `CLIENT` now means that the switch is off, while it was on in `CLIENT` mode before

robbederks: I'll let you double check first before merging in the branch, don't need it right away, and master still works fine on all pandas that are non-black

rbiasini:
yeah, this is needed. 
-Could not find unexpected behavior on white/black pandas with this change
- could replicate black panda standalone not working

I have bumped the version, so pandas get auto updated.

PR #293, opened 2019-10-09
Panda safety code for Volkswagen, Audi, SEAT, and Škoda

jyoung8607:
Third of five expected PRs for adding Volkswagen MQB platform support. I acknowledge with gratitude the work @cbenhagen did to get this started back in the day, particularly around regression tests (see PR #157) but the Panda codebase has moved on considerably since then, and this one supersedes.

This version uses the standard templated torque limiting code, resolves some issues that were preventing it from working on-the-road, and adds GRA_ACC_01 filtering so OP can take control of the cruise-control buttons. I don't have the resources to verify MISRA compliance. I will happily take any feedback you have on that front, or any other.


rbiasini:
High quality PR, good job. 
I can help you out with fixing the regression tests, I should be able to push commits directly to your branch (or maybe you have to grant me the permission?)

jyoung8607:
> High quality PR, good job.

Thank you!

> I can help you out with fixing the regression tests, I should be able to push commits directly to your branch (or maybe you have to grant me the permission?)

For my personal education I gave MISRA their £15 of flesh, and together with some hints from the community on Discord, I think I have it beaten into shape. That said, I have made you contributor on that repo and am happy to take any revisions or feedback on how it was done.

board/safety/safety_volkswagen.h
```diff
@@ -0,0 +1,169 @@
...
+    }
+
+    if (violation) {
+      tx = 0;
+    }
+
+  }
+
+  // 1 allows the message through
+  return tx;
```
jyoung8607: The violation->tx thing is a little redundant/overkill in our case, because our only no-transmit condition is a torque violation at this time. But, "it's the way everyone else does it" for convenience of review and audit, and we're likely to grow other conditions later.

board/safety/safety_volkswagen.h
```diff
@@ -0,0 +1,169 @@
...
+  if ((bus == 0) && (addr == MSG_KLEMMEN_STATUS_01)) {
+    vw_ignition_started = (GET_BYTE(to_push, 2) & 0x2) >> 1;
+  }
+
+  // Update driver input torque samples from EPS_01.Driver_Strain for absolute torque, and EPS_01.Driver_Strain_VZ
+  // for the direction.
+  if ((bus == 0) && (addr == MSG_EPS_01)) {
+    int torque_driver_new = GET_BYTE(to_push, 5) | ((GET_BYTE(to_push, 6) & 0x1F) << 8);
+    int sign = (GET_BYTE(to_push, 6) & 0x80) >> 7;
+    if (sign == 1) {
```
rbiasini: so weird they did not make it a 14 bits signed signal. the 2 bits gap between the message and the sign is very unusual.

board/safety/safety_volkswagen.h
```diff
@@ -0,0 +1,169 @@
+const int VW_MAX_STEER = 300;               // 3.0 nm
+const int VW_MAX_RT_DELTA = 130;            // max delta torque allowed for real time checks
+const uint32_t VW_RT_INTERVAL = 250000;     // 250ms between real time checks
+const int VW_MAX_RATE_UP = 10;
```
rbiasini: you might get better performance (for the safety) by having the rate up lower (in absolute value) than the rate down. From a safety perspective, there is really no constraint on the rate down number, since having an EPS fault would cause an immediate loss of torque anyway. 

Is the steer command sent at 100Hz?

board/safety/safety_volkswagen.h
rbiasini: yeah, I prefer keeping the same format, even if it's a little redundant

board/safety/safety_volkswagen.h
```diff
@@ -0,0 +1,169 @@
...
+    if (violation || !controls_allowed) {
+      vw_desired_torque_last = 0;
+      vw_rt_torque_last = 0;
+      vw_ts_last = ts;
+    }
+
+    if (violation) {
+      tx = 0;
+    }
+
```
rbiasini: extra line

board/safety/safety_volkswagen.h
```diff
@@ -0,0 +1,169 @@
...
+#define MSG_LDW_02              0x397
+#define MSG_KLEMMEN_STATUS_01   0x3C0
+
+static void volkswagen_init(int16_t param) {
+  UNUSED(param); // May use param in the future to indicate MQB vs PQ35/PQ46/NMS vs MLB, or wiring configuration.
+  controls_allowed = 0;
+  vw_ignition_started = 0;
+}
+
+static int volkswagen_ign_hook(void) {
```
rbiasini: you can keep the comment (perhaps move it before the safety_hooks definitions), but I would delete this function until it's different from the default_hook, which you should use instead.

board/safety/safety_volkswagen.h
```diff
@@ -0,0 +1,169 @@
+const int VW_MAX_STEER = 300;               // 3.0 nm
+const int VW_MAX_RT_DELTA = 130;            // max delta torque allowed for real time checks
```
rbiasini: where does this come from? It should be:
`VW_MAX_RATE_UP * message_freq * VW_RT_INTERVAL / 1000000 * 1.5`

where the final extra 50% is to allow for tolerable timing delays from controlsd. what's message_freq?

board/safety/safety_volkswagen.h
jyoung8607: Germans, what are you gonna do. Anyway, they are somewhat consistent about putting the sign of directional values in the high-bit of the MSB byte. Examples include the driver input torque and steering angle in EPS_01 and the heading control assist torque in HCA_01.

board/safety/safety_volkswagen.h
jyoung8607: The rack will tolerate a range of send rates. Stock sends at 50Hz when steering and 10Hz when not steering. We have tested sending at 100Hz and it works fine, but I did not perceive any real benefit and reverted to 50Hz.

The rack seems to tolerate literally anything for up/down rate; during development I let OP send through unlimited raw values without OP or Panda limiting in place. However, we had some really tough tuning problems, nasty high amplitude oscillations at low speed and so forth. I recently got information from "a source" that the _effective_ rate may be shaped to a 5nm/sec rate of change. It's not clear if that applies to both up and down.

board/safety/safety_volkswagen.h
jyoung8607: Removed.

board/safety/safety_volkswagen.h
jyoung8607: Change accepted.

board/safety/safety_volkswagen.h
rbiasini: Ok, we can push tuning for later and keep this "as is" for now.

board/safety/safety_volkswagen.h
rbiasini: so, if it's 50Hz, then this number should be:
10 * 50 * 250000 / 1000000 * 1.5 = 188

tests/safety/test_volkswagen.py
```diff
@@ -0,0 +1,193 @@
+#!/usr/bin/env python3
```
rbiasini: tests are good. played with them and any change to params caused a test failure, as expected.

board/safety/safety_volkswagen.h
jyoung8607: Send rate is 50Hz. The current value of 130 (125 rounded up) was essentially that minus the 50% pad. I've committed a change to 190 (187.5 rounded up).

board/safety/safety_volkswagen.h
```diff
@@ -0,0 +1,162 @@
...
+  int addr = GET_ADDR(to_fwd);
+  int bus_fwd = -1;
+
+  // NOTE: Will need refactoring for other bus layouts, such as no-forwarding at camera or J533 running-gear CAN
+
+  switch(bus_num) {
+    case 0:
+      if(addr == MSG_GRA_ACC_01) {
+        // OP intercepts, filters, and updates the cruise-control button messages before they reach the ACC radar.
+        bus_fwd = -1;
```
rbiasini: mmm... this is the only thing I'm not sure about. For example, what happens if you mistakenly send a set press to the radar? Does the radar start commanding acceleration?

board/safety/safety_volkswagen.h
```diff
@@ -0,0 +1,162 @@
+const int VW_MAX_STEER = 300;               // 3.0 nm
+const int VW_MAX_RT_DELTA = 190;            // max delta torque allowed for real time checks
```
rbiasini: can we do 188 precisely? :) otherwise we are going to be confused later on where it came from

jyoung8607: How about 188 and some free commentary from the peanut gallery?

board/safety/safety_volkswagen.h
jyoung8607: A fair question, and the answer would be yes. That said, I'm not sure if we have more exposure than any other port that's set up to spam cancel to comply with Comma strict safety, or spam resume to get stock ACC moving again in traffic. The only thing special about this port is the message is CRC-and-counter protected, so we have to stick ourselves in the middle of the message stream rather than just blat things out on the wire. Thoughts welcome.

rbiasini: well, other ports have a check to make sure that only the cancel button can be sent.
I think we need that here as well. See https://github.com/commaai/panda/blob/master/board/safety/safety_honda.h#L156

Anyway, happy to merge this PR as is. It's good enough.

jyoung8607: > I'm also curious about why this is needed at all: I'm still not convinced but I'll know more when I see the port to openpilot.

A preview: https://github.com/jyoung8607/openpilot/blob/vw-064-devel/selfdrive/car/volkswagen/carcontroller.py#L98-L142

The short version: I want to send cancel and resume at a minimum, and I expect I would piss off the radar if it started getting duplicate/out-of-order counter values, although I admit I never tried. I also want to actually STOP set/resume from getting through if OP is not engageable, and for that I definitely have to get in the middle.

> I think we need something like that here as well. See https://github.com/commaai/panda/blob/master/board/safety/safety_honda.h#L156
> Anyway, happy to merge this PR as is. It's good enough.
 
Hmm, I like that. It's valid to restrict OP from sending resume unless stock ACC is already engaged, which it will be in stop-and-go traffic where OP needs to do so and it wouldn't be surprising to the user for the car to move. If that condition isn't met, I could potentially have Panda safety require any set/resume button press to be received from the car on 0 before it can be transmitted to the radar on 2.

Per our conversation, I'll take that as a TODO item, and try it out sometime after the current PR is merged.

PR #294, opened 2019-10-09
GPS UART fix until boardd is refactored

robbederks:
This makes sure that each time the last available characters are available to be sent when the usb command is used.

This results in the message length coming in to boardd only being 1 or 2 chars, not ruining the logic there.

### DIFF
--- a/board/main.c
+++ b/board/main.c
@@ -399,6 +399,12 @@ int usb_cb_control_msg(USB_Setup_TypeDef *setup, uint8_t *resp, bool hardwired) 
       if (!ur) {
         break;
       }
+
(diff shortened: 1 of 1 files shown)


board/main.c
```diff
@@ -399,6 +399,12 @@ int usb_cb_control_msg(USB_Setup_TypeDef *setup, uint8_t *resp, bool hardwired)
       if (!ur) {
         break;
       }
+
+      // TODO: Remove this again and fix boardd code to hande the message bursts instead of single chars
+      if(ur == &uart_ring_esp_gps){
```
rbiasini: being pedantic :)
```
if (ur == &uart_ring_esp_gps) {
```

PR #303, opened 2019-10-21
make ignition logic common for all cars

rbiasini:
some cars have no ignition line and we have a chicken and egg problem: boardd sets the safety mode once `started` is detected, but the started logic depends on the safety mode.

This PR standardizes the `started` detection logic across all cars, by always looking at ignition from specific CAN messages.

Risk:
- False positive detection of ignition due to messages unrelated to ignition being present in unknown cars. 

Mitigation:
- currently the ignition based messages (as a combination of message addr and length) are only present in the cars where can based ignition is needed. 
- If in the future we add cars that share messages used to detect ignition, we can always add additionally rules to disambiguate them (we just have to find unique patterns in a 64 bits message)
- most cars won't keep CAN line on after car is turned off, so we can always switch to ignition False after a timeout period.


board/power_saving.h
```diff
@@ -25,22 +25,23 @@ void set_power_save_state(int state) {
       enable = true;
     }
 
-    // Switch CAN transcievers
-    current_board->enable_can_transcievers(enable);
+    // Switch CAN transcievers except for BUS 1, which must remain on to detect CAN based ignition
+    current_board->enable_can_transciever(2U, enable);
+    current_board->enable_can_transciever(3U, enable);
```
robbederks: This doesn't include the 4th CAN transceiver on black panda / uno. Maybe we want to do this in a more generic way? Not entirely sure how yet though.

rbiasini: `enable_can_transcievers` can maybe have a second input that is an uint8_t array of buses to leave untouched?

PR #389, opened 2019-11-21
Power saving refactor

rbiasini:
What was wrong:
- can based ignition would not trigger the interrupt on ignition line that changes the power save mode and set usb power mode to CDP.
- easy to trigger race conditions where EON changes the safety mode and re-enables some of the CAN lines after the ignition off interrupt is triggered. 

What is done:
- Move power save mode logic to boardd (openpilot). ON when car is ON, OFF when car is ON.
- Only set_power_save_state call in panda is when EON is disconnected: we already enter SILENT safety mode and now we also enter power saving mode, to ensure that if EON is disconnected while car is ON, panda won't remain in high consumption mode.
- enter CDP mode when (check_started) is True
- added usb command to change power save state
- added power save state to the health packet
- disable ignition line interrupt

### DIFF
--- a/board/boards/white.h
+++ b/board/boards/white.h
@@ -302,13 +302,6 @@ void white_grey_common_init(void) {
   // Set normal CAN mode
   white_set_can_mode(CAN_MODE_NORMAL);
 
-  // Setup ignition interrupts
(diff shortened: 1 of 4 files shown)


rbiasini:
@robbederks what do you think?

PR #391, opened 2019-11-21
Add uptime counter to the health packet

robbederks:


rbiasini:
boardd in openpilot also needs to change

PR #392, opened 2019-11-22
Fix can_logger.py to run correctly on python3

claydugo:
Fix opening a csv file to python3 format, use built in .hex() function instead of binascii, and use f-strings instead of str() methods.


robbederks:
Thanks!

derhaegr:
Format strings are only supported in Python 3.6 and greater, 3.5 is default for ubuntu 16.04

PR #399, opened 2019-12-06
Command to get signature

pd0wm:
I want to reflash the panda based on signature mismatch instead of version numbers we can forget to increase. This PR adds panda support to read the 128 byte RSA signature at the end of the code section


robbederks:
Looks good! Think the sig check is a good idea

PR #403, opened 2019-12-14
WIP: Checksum checks

rbiasini:
In addition to openpilot, this PR adds safety checks on safety relevant messages to panda as well:
- checksum checks
- counter checks
- dead/lagging messages

TODO:
- [x] checksum and counter checks only implemented for Honda and Toyota for now.
- ~[ ] Remove hack to ignore bus check in Honda. Still thinking about a clean way to deal with Nidec, Bosch Giraffe and Bosch Harness.~
- [x] Add regression tests to stress these checks


board/safety.h
```diff
@@ -67,6 +67,71 @@ bool msg_allowed(int addr, int bus, const AddrBus addr_list[], int len) {
...
+  int addr = GET_ADDR(to_push);
+
+  int index = -1;
+  for (int i = 0; i < len; i++) {
+    for (uint8_t j = 0U; j < addr_list[i].addr_len; j++) {
+      // TODO: remove rule to skip bus check if expected bus is negative. See safety_honda.h
+      if ((addr == addr_list[i].addr[j]) && ((bus == addr_list[i].bus) || (addr_list[i].bus < 0))) {
+        index = i;
+        i = len;  // break also second loop
+        break;
```
pd0wm: why not just return early here? Or is that not allowed in misra?

board/safety.h
```diff
@@ -67,6 +67,71 @@ bool msg_allowed(int addr, int bus, const AddrBus addr_list[], int len) {
...
+    }
+  }
+  return index;
+}
+
+void check_lagging_addrs(const safety_hooks *hooks) {
+  uint32_t ts = TIM2->CNT;
+  for (int i=0; i < hooks->addr_check_len; i++) {
+    uint32_t elapsed_time = get_ts_elapsed(ts, hooks->addr_check[i].last_timestamp);
+    // lag threshold is max of: 1s and MAX_MISSED_MSGS * expected timestep.
```
pd0wm: The check is called at 1Hz, so the worst case time to trigger is 2 seconds. Should we document this?

board/safety/safety_honda.h
```diff
@@ -6,6 +6,13 @@
 //      accel rising edge
 //      brake rising edge
 //      brake > 0mph
+
+// TODO: -1 bus check to indicate that bus check is skipped. Temp solution
+AddrCheckStruct HONDA_RX_CHECKS[] = {{{0x1A6, 0x296}, 2U, -1, true, 3U, 40000U, true, 0, 0U, 0U, true},
```
pd0wm: I would replace the addresses by defines/constants. The code is now riddled with these magic numbers.

board/safety/safety_toyota.h
```diff
@@ -36,13 +38,36 @@ int toyota_gas_prev = 0;
...
 
 
+static bool toyota_addr_check(CAN_FIFOMailBox_TypeDef *to_push) {
+  int index = get_addr_check_index(to_push, TOYOTA_RX_CHECKS, TOYOTA_RX_CHECKS_LEN);
+  update_addr_timestamp(TOYOTA_RX_CHECKS, index);
+
+  // checksum check
+  if (index != -1) {
+    if (TOYOTA_RX_CHECKS[index].check_checksum) {
+      int checksum_byte_pos = GET_LEN(to_push) - 1;
```
pd0wm: extract checksum computation to separate function.

board/safety/safety_honda.h
```diff
@@ -18,21 +25,56 @@ bool honda_bosch_hardware = false;
...
 
   int addr = GET_ADDR(to_push);
   int len = GET_LEN(to_push);
   int bus = GET_BUS(to_push);
 
+  bool valid = honda_addr_check(to_push);
+
   // sample speed
-  if (addr == 0x158) {
+  if ((addr == 0x158) && valid) {
```
pd0wm: Why check for valid everywhere. Can we just skip the whole function if the message is not valid?

board/main.c
```diff
@@ -742,6 +742,9 @@ void TIM1_BRK_TIM9_IRQ_Handler(void) {
     uptime_cnt += 1U;
     safety_mode_cnt += 1U;
     ignition_can_cnt += 1U;
+
+    // check for lagging safety messages
+    check_lagging_addrs(current_hooks);
```
rbiasini: change name into safety tick

board/safety/safety_chrysler.h
```diff
@@ -137,4 +137,6 @@ const safety_hooks chrysler_hooks = {
   .tx = chrysler_tx_hook,
   .tx_lin = nooutput_tx_lin_hook,
   .fwd = chrysler_fwd_hook,
+  .addr_check = default_addr_check,
```
rbiasini: remove this and default to NULL

board/safety.h
rbiasini: 15.1 prohibits goto, but it's also advisory. I'll just suppress that rule. It's fine as long as we follow 15.2 and 15.3

board/safety/safety_honda.h
rbiasini: I'll clean up in another PR. what's done in safety_volkswagen is better indeed.

board/safety/safety_honda.h
```diff
@@ -18,8 +27,45 @@ bool honda_bosch_hardware = false;
...
+      update_counter(HONDA_RX_CHECKS, index, counter);
+    }
+  }
+  return is_msg_valid(HONDA_RX_CHECKS, index);
+}
+
 static void honda_rx_hook(CAN_FIFOMailBox_TypeDef *to_push) {
 
+  if (!honda_addr_check(to_push)) {
+    goto Invalid;
```
geohot: I don't think you need a goto here, just put the body in the if

board/safety/safety_toyota.h
```diff
@@ -18,11 +18,15 @@ const int TOYOTA_MIN_ACCEL = -3000;       // 3.0 m/s2
...
+  {.addr = {0x1D2}, .addr_len = 1U, .bus = 0, .check_checksum = true, .max_counter = 0U, .expected_timestep = 30000U},
+};
 const AddrBus TOYOTA_TX_MSGS[] = {{0x283, 0}, {0x2E6, 0}, {0x2E7, 0}, {0x33E, 0}, {0x344, 0}, {0x365, 0}, {0x366, 0}, {0x4CB, 0},  // DSU bus 0
-                                         {0x128, 1}, {0x141, 1}, {0x160, 1}, {0x161, 1}, {0x470, 1},  // DSU bus 1
-                                         {0x2E4, 0}, {0x411, 0}, {0x412, 0}, {0x343, 0}, {0x1D2, 0}, // LKAS + ACC
-                                         {0x200, 0}};  // interceptor
+                                  {0x128, 1}, {0x141, 1}, {0x160, 1}, {0x161, 1}, {0x470, 1},  // DSU bus 1
+                                  {0x2E4, 0}, {0x411, 0}, {0x412, 0}, {0x343, 0}, {0x1D2, 0}, // LKAS + ACC
+                                  {0x200, 0}};  // interceptor
+const int TOYOTA_RX_CHECKS_LEN = sizeof(TOYOTA_RX_CHECKS) / sizeof(TOYOTA_RX_CHECKS[0]);
```
geohot: Move this to line 24

board/safety/safety_toyota.h
```diff
@@ -36,8 +40,33 @@ int toyota_gas_prev = 0;
...
+    }
+  }
+
+  return is_msg_valid(TOYOTA_RX_CHECKS, index);
+}
+
 static void toyota_rx_hook(CAN_FIFOMailBox_TypeDef *to_push) {
 
+  if (!toyota_addr_check(to_push)) {
+    goto Invalid;
```
geohot: Same comment as above, no goto

PR #418, opened 2020-01-02
update J2534 driver for new health data structure

gregjhogan:
HDS for pre-2017 vehicles won't open because it can't read the car battery voltage due to https://github.com/commaai/panda/pull/391 changing the order of the health packet type def


gregjhogan:
installer: 
[panda J2534 driver install.zip](https://github.com/commaai/panda/files/4017364/panda.J2534.driver.install.zip)

PR #421, opened 2020-01-12
TX message guaranteed delivery

gregjhogan:
It is easy to overflow the TX buffer when sending a large number of messages rapidly. When sending UDS messages you don't want anything to get silently dropped. This change causes the client to pause (leave NAK active) until there is enough room in the TX buffer to process the next USB data packet. Not sure if this is desirable in all scenarios, or even if this is the right way to handle the situation, but it does seem to work.

### DIFF
--- a/board/drivers/can.h
+++ b/board/drivers/can.h
@@ -28,7 +28,7 @@ void can_set_forwarding(int from, int to);
 
 void can_init(uint8_t can_number);
 void can_init_all(void);
-bool can_tx_check_min_slots_free(int min);
+bool can_tx_check_min_slots_free(uint32_t min);
 void can_send(CAN_FIFOMailBox_TypeDef *to_push, uint8_t bus_number, bool skip_tx_hook);
 bool can_pop(can_ring *q, CAN_FIFOMailBox_TypeDef *elem);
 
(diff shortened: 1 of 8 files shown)


robbederks:
Pretty sure this change also means that we need to add a timeout to the openpilot boardd code to prevent it from freezing if the CAN buffers are not cleared.
Another option would be to disable this NAK'ing for openpilot.
Any thoughts @pd0wm?

pd0wm:
For openpilot the most recent messages are more important, so it makes sense to drop older ones if the buffer overflows. I would like the to detect and log when this happens though!

gregjhogan:
I was thinking we should set the timeout low (instead of 0) in [boardd](https://github.com/commaai/openpilot/blob/9c16ad8e184171f8d07bfc764e89a1067f4d1086/selfdrive/boardd/boardd.cc#L562) and then skip past messages that time out.

pd0wm:
If I understand correctly, that will result in the same behavior as we have right now?

gregjhogan:
very close to the same since the ring buffer is pretty small (once a message is in the buffer it will always get sent); could also add an immediate NAK setting for OP but probably not necessary

PR #425, opened 2020-01-18
Fix pending_can_live reset

neon-dev:
Didn't look intentional, according to the comment above changed line.
Only thing is, I couldn't figure out where the `can_live` variable is ultimately used. Maybe it has no purpose anymore and therefore `can_live` and `pending_can_live` could be removed?


ohasanliUW:
Operator & is absolutely fine on that line.
Every 16th pass means last 4 bits of the counter will be reset to 0. Checking last 4 bits by bitwise operation is a common practice.
`counter % 16` also can be used but not `counter % 0xF` since 0xF is actually 15.

neon-dev:
Oh boy, you're right. I was confusing 0xF with 16, sorry. Since 15 is first four bits set to 1, old check was fine.

PR #427, opened 2020-01-19
Hyundai-Kia-Genesis (HKG)

xx979xx:
this PR has to accompany by [openpilot](https://github.com/commaai/openpilot/pull/970)
driver torque signal is context here [openddbc](https://github.com/commaai/opendbc/pull/211)


geohot:
If a car doesn't have adaptive cruise control, we won't upstream it. Is that what the SCC check is?

board/safety/safety_hyundai.h
```diff
@@ -29,14 +30,17 @@ static int hyundai_rx_hook(CAN_FIFOMailBox_TypeDef *to_push) {
     int bus = GET_BUS(to_push);
     int addr = GET_ADDR(to_push);
 
-    if (addr == 897) {
-      int torque_driver_new = ((GET_BYTES_04(to_push) >> 11) & 0xfff) - 2048;
+    if (addr == 593) {
```
geohot: What is the advantage of using this signal over the previous one?

TK211X: `SG_ CR_Mdps_DrvTq : 11|12@1+ (0.0078125,-2048) [-2048|2046] "" SPAS`

Is a SPAS dependent message. If a proper SPAS signal is not active, some MDPS units do not output DrvTq by default. However, the torque values output in 593 (MDPS12) are always alive and are produced by MDPS specifically for LKA logic.

geohot: My concern is that 593 includes computer torque, while 897 is only user torque. Unless we absolutely have to, I'd rather not change this.

TK211X: What

TK211X: Ok so basically you have two options right.

1. Create SPAS messages so MDPS11 creates DrvTq for all vehicles.

2. Use the OEM default message created for LKAS and ESC to derive the exact driver torque input.

TK211X: image

xx979xx: here is the confirmation:
since 593 signal is 11 bits long vs 879 12 bits, I have to scale up the first by factor 0.79
comparison after scale
it is reverse, I first discover this when I tried LCA, you have to nag right to initiate left lane change. surely this is a huge flaw

xx979xx: when 879 signal reversed, it is identical only for SPAS supported cars. other will have none
comparison after scale and flip

geohot: This doesn't answer my original question, this is a polarity issue fixed by a minus sign.

We are looking for the signal that is only user torque, and does not also include computer applied torque.

geohot: Also, our Santa Fe did not have parking assist, but 593 was present.

geohot:
Aside from the comment inline, assuming we do want to got with this change, we will have to fix the safety replay test.

@pd0wm Do you know how to do this? Can we put instructions somewhere?

board/safety/safety_hyundai.h
xx979xx: I do not know to be honest, my point is that both signal are identical. if first one is user torque then the new one is. 
however, here is a cabana showing LKAS steer req, Mdps new signal and steering angle:
driversteer
**you can clearly see that Mdps_StrColTq did not move as result of LKAS steer req, even though the steering angle decreased significantly**

geohot: Ok, I'm sold on the signal not being user torque. Have you confirmed this works as intended in panda? This is supposed to allow for easy override of computer applied torque.

I'm still not following the polarity thing. The lane change has nothing to do with panda, I could see how it might be wrong on the openpilot side, but this also flips the polarity in panda?

xx979xx: here is the issues that fixed by this change and change in OP code:
- LKAS fault when some users override steering, caused by panda blocking msgs as result of driver limit check violation
- OP wont recognize some users steer input to make DM happy or make lane change
I will upload the same comparison for old signal
do not deceive by name, some user will get no driver steering signal

geohot: Do we know which cars have each signal?

And that fault is by design. Not triggering it is not a fix! openpilot should be changed to not trigger it, it's not the fault that should be changed. See driver_limit_check in board/safety.h, it's designed to reduce/stop applying torque when the user does, and is a part of the safety model.

TK211X: DRVTQ:

Any vehicle MDPS that is provisioned to have SPAS for the model either regionally or globally will output DrvTq via MDPS11 or MDPS legacy message..... and / or for any other vehicle ECU that requires it possibly for newer vehicles. There is no way to actually know what the additional trigger would be without doing lots of testing. However, the one sure fire way to have it output is to produce a valid SPAS message. So if you are adamant about using DRvTq you can simply make a idle SPAS message to make the MDPS flag and produce the required messages to support it (DrvTq)

STRCOLTQ:

Any vehicle that has LKAS or VSM will output STRCOLTQ via MDPS12. Practically every HKG that we can possibly support has VSM and as such the MDPS produces MDPS12 that contains the messages for it's operation. In addition, STRCOLTQ appears to be more precise as opposed to DRVTQ.


SPAS and VSM came out in like 2010-2011

LKAS came out in like 2014

We can hypothesize that the legacy MDPS messages supported SPAS and VSM. When LKAS came out they fragmented the new MDPS messages (MDPS11 & MDPS12) with separate messages (DRVTQ & STRCOLTQ) to support different arrangements of operation between the types of VSM and SPAS modules. HKMC tends to not over complicate things and make operation mostly universal on their platform so it would make sense they did this.

George you have two options. Doesn't matter what you prefer honestly we can make it work either way.

xx979xx: here is The Last Straw:
I had this user today who has some panda related issue,  I checked `MDPS11 CR_Mdps_DrvTq`, his car has its polarity corrected [cabana link](https://my.comma.ai/cabana/?route=7653b2bce7bcfdaa%7C2020-01-26--10-20-16&exp=1611571041&sig=JV7aVrVwSEm%2Fj6nHXIs2N00%2BAfXCoRdFL7DqAdBHp6E%3D&max=6&url=https%3A%2F%2Fchffrprivate-vzn.azureedge.net%2Fchffrprivate3%2Fv2%2F7653b2bce7bcfdaa%2F819a09170039184e45b4f710ef602aef_2020-01-26--10-20-16)
so, this signal is inconsistent (some do not have it and  the reset some of them have it reversed )
his original issue is that LCAN has msgs with the same addresses as SCC11 and SCC12. the fix is to disconnect L-CAN from CAN2 as it is useless ( it has cam data sent to radar!)

geohot: If we want to make this switch, we'll need to verify the torque limiting functionality works as intended on all cars.

It's hard to follow this, "his original issue is that LCAN has msgs with the same addresses as SCC11 and SCC12.", like why is this ever happening? If the relay is closed, it should only connect CAN busses that are connected in the stock car. I can't see where new messages would come from.

TK211X:
Lol. You officially support elantra and DrvTq has never worked for my car only StrColTq. I never needed OpenPilot to have it work cause I always maxed out the time limit locally in my EON code. Now that we have driver monitoring at night I don’t feel like restarting OpenPilot, just nudge the wheel, this is why it coming out of the woodwork right now. It’s not like we just felt like doing it got explained multiple times.

StrColTq = Universal

DrvTq = LoL who knows

KU7 coded it to DrvTq initially cause it’s literally in the name and he saw it active for him. Not much consideration was put in when he used DrvTq for the lols.


I’m not replying to this specific thing again.

board/safety/safety_hyundai.h
TK211X: Lol. You officially support elantra and DrvTq has never worked for my car only StrColTq. I never needed OpenPilot to have it work cause I always maxed out the time limit locally in my EON code. Now that we have driver monitoring at night I don’t feel like restarting OpenPilot, just nudge the wheel, this is why it coming out of the woodwork right now. It’s not like we just felt like doing for no reason. It got explained multiple times

StrColTq = Universal

DrvTq = LoL who knows

KU7 coded it to DrvTq initially cause it’s literally in the name and he saw it active for him. Not much consideration was put in when he used DrvTq for the lols.


I’m not replying to this specific thing again.

xx979xx: @geohot fair enough for me
regarding the LCAN issue,  `hyundai_rx_hook` goes like this:
```
    if (addr == 1057) {
      // 2 bits: 13-14
      int cruise_engaged = (GET_BYTES_04(to_push) >> 13) & 0x3;
      if (cruise_engaged && !hyundai_cruise_engaged_last) {
        controls_allowed = 1;
      }
```
so, if the same address received from any Bus, the condition is met.
sure, it can be fix by check for bus also

xx979xx: today another user has L-CAN sending messages with LKAS12 address, which caused OP to switch to dashcam mode.
I hope comma will not connect L-can to CAN2 in HKG harness, it just cause issue for nothing

geohot: Oh the hook can certainly be fixed to require the correct bus.

geohot: Where are we on this? Can we use the old message if we also check the bus?

xx979xx: yes sure we can,
but L-CAN issue will continue to confuse Op fingerprinting and cam_ecu check

geohot:
@rbiasini Does this make sense?

@pd0wm How do we update the safety replay?

pd0wm:
For the safety replay we just need a recorded drive with the changes to the safety code. Then we can sync it to the CI bucket and change it in the test.

PR #431, opened 2020-01-28
Added the option to turn on IR in power savings mode

robbederks:
Needed changes for issue commaai/openpilot#1005.

Needs changes in boardd too to make sure that they turn off when ignition goes low


pd0wm:
https://github.com/commaai/openpilot/commit/3a65206c8d0400fb22239c806429842236afac2b

PR #432, opened 2020-02-07
Panda side lateral support for Chevy Bolt EV, Chevy Equinox

JasonJShuler:
The Bolt EV uses much of the same messages as the existing GM vehicles, but as it does not have an ASCM the development of a Giraffe is much simpler - I have been able to implement AEB and auto-brights passthrough. (The full implementation also passes through stock LKAS when OP is disengaged)

There are two fundamental differences with the Volt.

1. The Bolt EV's PSCM does not tolerate gaps in the rolling counter, when active or inactive. If the sequence is not 0,1,2,3,0,1,2,3... the PSCM faults in such a way that the car must be shut off, the door opened and closed, and you must wait for about 4 minutes. A single message out of sequence will cause such a fault.

The Volt is much more tolerant - the existing safety code would block frames with a safety violation, or on disengagement. This is resolved by changing the safety code to always send a frame, but it sends an inactive 0 frame (with the correct rolling counter) when it would have previously blocked.

This resolved the engage / disengage issues.

2. The Power Steering Control Module on the Bolt is _extremely_ sensitive to the timing of the LKAS frames while active. They must be 50hz (every 20ms). If there is even a 2ms rush or drag, the PSCM faults in the same semi-permanent way. (note that the stock LKAS when inactive is sent every 100ms, and the PSCM is much more tolerant of delays)
Under ideal circumstances, when the EON generated the LKAS frames from Python I was able to get 45-minutes (max) before a fault would occur. When I enabled forwarding, this was reduced to a 5 minute average.

To resolve this, I am using one of the hardware timers on the STM32 (timer 7) to publish LKAS messages with reliable timing. The safety code buffers the LKAS frames from OP and the stock camera, but the Panda maintains the rolling counter. In this way we can seamlessly switch between stock and OP LKAS.

The message pump is only started when LKAS frames have been received. It tracks the age of the buffered frames - switching to inactive mode if the buffered frame is older than 40ms, disabling the pump if the buffer is older than 1 second.

Additionally, forwarding is only performed after LKAS frames have been received on CAN2. This should prevent issues with CAN2 being on the object or chassis bus. (alternatively we could setup a separate safety model for non-ASCM GM vehicles)

I have personally tested this code over many hours of error-free operation. A couple others have started using my fork and Giraffe with great success.

There is information about the port here: https://shulerent.com/openpilot-2019-chevy-bolt-port/


geohot:
On the surface, I'm very reluctant to merge this. The job of the panda is to enforce the safety model, not a place to put functionality. It shouldn't be sending messages, it should just be gating messages from the EON. If this is really needed, I would prefer something that looks more like the EON scheduling a message for the future. Though I've never seen a car be this strict about CAN timings, so I'm wondering if it's possible it's something else. If not:

If we need 2ms precision, we should be improving the Python code to get there. It's definitely possible, and openpilot overall would be improved for everyone. For us, it's about long term value of the project, not adding certain car specific hacks. I understand fork maintainers might have a different perspective, particularly working on it as a volunteer.

How about this, the "CAN Precision Project" has value to the company. Come up with a current understanding of where we are, where we need to be, and propose a bounty if you can make it happen. We'd upstream the tools to instrument this as well, and we have a testing closet which we can run 10 EONs on the code checking for violations.

For openpilot to be 1.0, we have to do this precision project anyway.

JasonJShuler:
Arguably even if we were to schedule messages, something on the Panda would be required to run on an interval to fulfill the scheduled messages.

I'm not sure it is the delayed message that causes the fault. What I would see is
20ms
20ms
20ms
30ms
10ms
------fault

It is possible that messages sent faster than 50hz are missed, so it could actually be the rolling counter issue in disguise... but the solution is the same. Reliable 50hz LKAS messages. It could even be configured to only ever publish an active frame once - thus the Panda would not technically be "generating" messages.

The other side of this is the use of AEB. In order for OP to issue AEB commands, it will need to generate its own rolling counter. I would argue that we must be prepared at all times for the stock AEB to override anything OP is doing, but the stock rolling counter is unlikely to be in sync with OP. The only way I can imagine to do this (due to the dual frame timing) is for the panda to maintain the rolling counter and apply it to either OP or stock messages (much the same as what this POC performs with LKAS)

So while there might be some tomfoolery that could avoid the LKAS fault without a message pump, there is really no getting around the pump for AEB.

geohot:
The scheduling would be generic though, and not specific to a car. Due to panda code needing to meet MISRA + the testing requirements, it is a lot more expensive to code and maintain.

Re: the AEB counter. AEB doesn't have an ASIL rating on working. Therefore, I would put the forwarding for this in openpilot, and have openpilot always generate the brake message. It can read the stock and pass it through.

Eventually, we will have our own AEB in openpilot once we can show that it is without a doubt better, and that would all be running on the EON. So this is a first step toward that. It's way better than the current GM which disables it.

JasonJShuler:
How much would such a solution be worth?

JasonJShuler:
I've never been very good at associating dollar value to effort but this project is very important to me and I want to help and I also need money

geohot:
I mean, the scheduling isn't worth much, that's more code to maintain. The real value is in fixing the timing issues with controlsd.

JasonJShuler:
I have a theory regarding the timing issue. It seems on my casual inspection of the Panda code that there is a _lot_ or work performed withing ISRs. If not all of the important work.

All the TX code is run within the ISR for each packet received. It would appear that all the RX code (and forward) code functions likewise.

I was raised with the belief that no (or very little) work should be done within ISRs because they "block" everything else. What I suspect we are encountering is simply the result of the Panda processing a burst of RX traffic, and the processing of that when there is a sufficiently persistent burst causes the TX code to be blocked. This explains why when forwarding is enabled the timing gets bad with much greater frequency.

My suggestion - if possible - would be to modify the critical and high-traffic interrupt handlers to queue only, and have a main code loop that processes the requests asynchronously. This main code loop would be able to apply prioritization (such as the CAN protocol message number prioritization) and prioritizing Transmit processing. Or processing TX and RX events in the order received.

The current implementation allows a busy CAN bus to saturate the Panda with ISR handlers and block reliable timing of TX messages.

Thoughts?
Is this work pursuing or has this approach already been considered and ruled out?

Thanks!

robbederks:
I'm pretty sure doing critical work in the ISRs is not uncommon or bad practice in embedded firmware, but we might consider giving them different priorities (ISRs with a higher priority will interrupt other ISRs currently running, otherwise they are queued up) in the future for good practice.
In the HIL latency tests, the latency of sending and receiving messages is consistently sub 300us, more than fast enough.

Most gains can definitely be had on the openpilot side!

JasonJShuler:
I'm going to close this PR because I am reimplementing the panda code in a new repository.

PR #435, opened 2020-02-17
Harness orientation designation fixes

hivoltag3:
This was causing me mental anguish

```
// flip CAN0 and CAN2 if we are flipped
  if (car_harness_status == HARNESS_STATUS_NORMAL) {
    can_flip_buses(0, 2);
  }
```


robbederks:
Wouldn't it be easier to just flip the two names?
There are also some compile errors due to some mistakes in the changes

hivoltag3:
Essentially this is flipping the names and renaming the relay pins to be more consistent and descriptive.  IMO it is more easily readable and understandable this way.

robbederks:
Thank you for fixing it. I'll clone this into a local branch such that the internal HIL tests can run to ensure functionality, and then it can be merged.

dzid26:
Just in case someone goes deep on this. - It took me a while to grasp this [line](https://github.com/hivoltag3/panda/blob/de82ff1d57cd2d2bd372dd4c8840dae8e5aad048/board/drivers/harness.h#L65):
```
// orientation flipped (PANDA_SBU1->HARNESS_SBU1(relay), PANDA_SBU2->HARNESS_SBU2(ign))
```
I don't know if this was initial intention of this PR, but it actually makes sense because SBU1 connects to SBU2 inside the USB-C cable. 
image
RXs<->TXs are also rerouted by the cable, by the way.

PR #442, opened 2020-02-18
Chrysler: disengage on gas press 

rbiasini:


rbiasini:
@adhintz , looks right?

rbiasini:
yes, scaling would be better since most of other ports have it scaled. Is the unit a percent? If so, I can just change the dbc factor to 0.01.

PR #443, opened 2020-02-19
Upgrades to VW MQB safety

jyoung8607:
Updated Volkswagen MQB safety for your consideration. Now with 38% more safety!

- Now insensitive to installed location (works for either camera or CAN gateway)
- Refactoring in support of upcoming Volkswagen PQ platform (commaai/cereal#19)
- German Engineered™ CRC and counter validation for all applicable messages (#411)
- Added standstill check and brake in motion safety check (#441)
- Seven new safety unit tests

The MQB code above passes automated safety testing (although not tried with replay) and has real-world mileage with several cars. The upcoming PQ platform safety is already written and already passing its own test suite, and you're welcome to look at it if you wonder why I refactored the way I did (see [VW unified safety](https://github.com/jyoung8607/openpilot/blob/vw-community-private-pq/panda/board/safety/safety_volkswagen.h) and [separate PQ regression test](https://github.com/jyoung8607/openpilot/blob/vw-community-private-pq/panda/tests/safety/test_volkswagen_pq.py)) but I've held it out of this PR until it has real-world mileage on it and the corresponding port is closer to upstream quality.

I will add a couple more comments inline with code after opening.

Issue #441, opened 2020-02-18
ISO26262: check for brake pressed in Volkswagen safety


rbiasini:

rbiasini:
https://github.com/commaai/panda/pull/445

Issue #411, opened 2019-12-21
ISO26262: add checksum/counter checks for Volkswagen


rbiasini:

rbiasini:
https://github.com/commaai/panda/pull/445

PR #445, opened 2020-02-19
Volkswagen safety updates: Phase 2

jyoung8607:
Second of roughly three PRs to review the overall change set proposed in #443.

- German Engineered™ CRC and counter validation for all applicable messages (#411)
- Added standstill check and brake in motion safety check (#441)
- Six new safety unit tests

### DIFF
--- a/board/safety.h
+++ b/board/safety.h
@@ -57,6 +57,23 @@ int safety_fwd_hook(int bus_num, CAN_FIFOMailBox_TypeDef *to_fwd) {
   return current_hooks->fwd(bus_num, to_fwd);
 }
 
+// Given a CRC-8 poly, generate a static lookup table to use with a fast CRC-8
+// algorithm. Called at init time for safety modes using CRC-8.
+void gen_crc_lookup_table(uint8_t poly, uint8_t crc_lut[]) {
+  int i, j;
+
+  for (i = 0; i < 256; i++) {
+    uint8_t crc = i;
+    for (j = 0; j < 8; j++) {
+      if ((crc & 0x80U) != 0U)
+        crc = (uint8_t)((crc << 1) ^ poly);
+      else
+        crc <<= 1;
+    }
+    crc_lut[i] = crc;
+  }
+}
+
 bool msg_allowed(int addr, int bus, const AddrBus addr_list[], int len) {
   bool allowed = false;
   for (int i = 0; i < len; i++) {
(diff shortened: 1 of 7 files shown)


board/safety/safety_volkswagen.h
```diff
@@ -83,6 +143,16 @@ static int volkswagen_mqb_rx_hook(CAN_FIFOMailBox_TypeDef *to_push) {
       volkswagen_gas_prev = gas;
     }
 
+    // Exit controls on rising edge of brake press
+    // Signal: ESP_05.ESP_Fahrer_bremst
+    if ((bus == 0) && (addr == MSG_ESP_05)) {
+      bool brake_pressed = (GET_BYTE(to_push, 3) & 0x4) >> 2;
+      if (brake_pressed && (!(volkswagen_brake_pressed_prev) || volkswagen_moving)) {
```
rbiasini: this should be:
- rising edge of brake press always cancels
- brake pressed with moving car also cancels

The case not covered here is: car is stopped, system is engaged and user presses the brake.

rbiasini: edit: my bad, did not read it properly. It's fine

board/safety.h
```diff
@@ -57,6 +57,23 @@ int safety_fwd_hook(int bus_num, CAN_FIFOMailBox_TypeDef *to_fwd) {
   return current_hooks->fwd(bus_num, to_fwd);
 }
 
+// Given a CRC-8 poly, generate a static lookup table to use with a fast CRC-8
+// algorithm. Called at init time for safety modes using CRC-8.
+void gen_crc_lookup_table(uint8_t poly, uint8_t crc_lut[]) {
+  int i, j;
+
+  for (i = 0; i < 256; i++) {
```
rbiasini: `for (int i = 0; i < 256; i++) {`

`i, j` can have smaller scope.

jyoung8607: Done

board/safety/safety_volkswagen.h
```diff
@@ -43,17 +93,27 @@ static void volkswagen_mqb_init(int16_t param) {
...
     int bus = GET_BUS(to_push);
     int addr = GET_ADDR(to_push);
 
+    // Update in-motion state by sampling front wheel speeds
+    // Signal: ESP_19.ESP_VL_Radgeschw_02 (front left)
+    // Signal: ESP_19.ESP_VR_Radgeschw_02 (front right)
+    if ((bus == 0) && (addr == MSG_ESP_19)) {
+      int wheel_speed_fl = GET_BYTE(to_push, 4) | (GET_BYTE(to_push, 5) << 8);
+      int wheel_speed_fr = GET_BYTE(to_push, 6) | (GET_BYTE(to_push, 7) << 8);
+      volkswagen_moving = (wheel_speed_fl > 0) || (wheel_speed_fr > 0);
```
rbiasini: this is not exactly the same as what we have in OP. In OP we check for standstill (speed < 0.1m/s). Here we should be less aggressive and perhaps just check that the average of the front wheel speeds > 0.3m/s to assign the moving bool to true.

jyoung8607: I will do that. It'll be a little bit because I'll want to buff the unit test to look on both sides of the grace threshold, and I have to leave here shortly.

PR #446, opened 2020-02-20
Subaru: disengage on gas press

rbiasini:


rbiasini:
@bugsy924 , looks right?

PR #448, opened 2020-02-24
abstract crc function

rbiasini:
wip: will add crc for hyundai

### DIFF
--- a/board/crc.h
+++ b/board/crc.h
@@ -1,0 +1,16 @@
+uint8_t crc_checksum(uint8_t *dat, int len, const uint8_t poly) {
+  uint8_t crc = 0xFF;
+  int i, j;
+  for (i = len - 1; i >= 0; i--) {
+    crc ^= dat[i];
+    for (j = 0; j < 8; j++) {
+      if ((crc & 0x80U) != 0U) {
+        crc = (uint8_t)((crc << 1) ^ poly);
+      }
+      else {
+        crc <<= 1;
+      }
+    }
+  }
+  return crc;
+}
--- a/board/pedal/main.c
+++ b/board/pedal/main.c
@@ -19,6 +19,7 @@
 #include "drivers/timer.h"
 
 #include "gpio.h"
+#include "crc.h"
 
 #define CAN CAN1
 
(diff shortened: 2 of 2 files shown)


rbiasini:
@jyoung8607 , seems like we can also use this abstraction in VW

jyoung8607:
> @jyoung8607 , seems like we can also use this abstraction in VW

A form of it, yes. IMO the proper approach is a merger:

* The pedal version can be trivially converted to the [faster LUT-based CRC-8 algorithm](http://www.sunshine2k.de/articles/coding/crc/understanding_crc.html#ch44) instead of feeding it the poly. The VW port [does this already](https://github.com/commaai/panda/blob/1b49d3e830f6d894f9daf820f956f315fd951428/board/safety/safety_volkswagen.h#L62). Similar consolidation could be done in [OpenDBC's implementation](https://github.com/commaai/opendbc/blob/master/can/common.cc).

* The [generic LUT generator](https://github.com/commaai/panda/blob/1b49d3e830f6d894f9daf820f956f315fd951428/board/safety.h#L62) should probably move from `safety.h` to the new `crc.h` regardless.

* The VW Panda safety (also HKG, Mazda, and Nissan) could then use the central CRC function, provided the IV and final XOR values are made parameters. Both will vary between ports, and VW (not sure about the others) will have to call it twice with different IVs and XORs to roll in the magic pad byte (or call it after making a separate copy of the data with the magic pad byte included).

rbiasini:
Yes, I agree. The table should be pre-computed.

jyoung8607:
Untested. Should be straightforward to transform for any CRC8 port by tweaking poly, init, and final. VW will need a tiny bit of specialsauce, but I can handle that if you want, once the global stuff is in.

**Pedal:**
```
declare:
  const uint8_t pedal_crc8_poly = 0xD5;
  const uint8_t pedal_crc8_init = 0xFF;
  const uint8_t pedal_crc8_final = 0x00;
  uint8_t pedal_crc8_lut[256];

init:
  gen_crc_lookup_table(pedal_crc8_poly, pedal_crc8_lut);

to use:
  crc_checksum(dat, CAN_GAS_SIZE - 1, pedal_crc8_lut, pedal_crc8_init, pedal_crc8_final) == dat[5]
```
**crc.h:**
```
uint8_t crc_checksum(uint8_t *dat, int len, uint8_t crc8_lut, uint8_t init, uint8_t final) {
  uint8_t crc = init;
  for (int i = 0; i < len; i++) {
    crc ^= dat[i];
    crc = crc8_lut[crc];
  }
  return crc ^ final;
}
```

rbiasini:
@jyoung8607 ok! I'll start merging this as is. Then I'll globalize the pre-computation as you mentioned, but I need to be careful because pedal has no regression tests.

PR #458, opened 2020-03-03
Mazda: Lane Info msgs need to be managed by OP

Jafaral:
The EPS drops torque messages if lane lines are not marked
visible by the camera. Drop stock laneinfo msgs  and
give control to OP to manage them.

Signed-off-by: Jafar Al-Gharaibeh <to.jafar@gmail.com>

### DIFF
--- a/board/safety/safety_mazda.h
+++ b/board/safety/safety_mazda.h
@@ -129,7 +129,7 @@ static int mazda_fwd_hook(int bus, CAN_FIFOMailBox_TypeDef *to_fwd) {
       bus_fwd = MAZDA_CAM;
     }
     else if (bus == MAZDA_CAM) {
-      if (!(addr == MAZDA_LKAS)) {
+      if (!((addr == MAZDA_LKAS) || (addr == MAZDA_LANEINFO))) {
         bus_fwd = MAZDA_MAIN;
       }
     }


robbederks:
Is this needed for all Mazdas? And if so, how did this work in the past?

Jafaral:
I had this in originally,  and I had the logic to control when to forward and when not to. I just looked at the history , it was dropped in [cfcce8f06bc4b94aafa5daed1e003c8d844192de]. For a while I didn't notice this because LKAS doesn't work below 28/32mph  and I drive on streets with good lane marking most of the time.  I did more testing the last few days and confirmed this is needed in all current generation Mazdas.

rbiasini:
would be nice to add a comment in the code

Jafaral:
@rbiasini a comment about Mazda not honoring the torque msgs if not confirmed by lane lines marking msgs ?

rbiasini:
yes, it's quite an exception

Jafaral:
@rbiasini it sure is. I actually thought it didn't matter initially. The laneinfo msgs were added from the start just as a nice to have feature to integrate with the dashboard. In the last few months since the code was dropped I was losing torque sometimes I couldn't explain it. I blamed on OP changes that I may have not caught up with Mazda. Last week a new Mazda OP user reported  this  issue so I looked at the data and review drives from three different Mazda cars. All behaved the same way, i.e EPS ignores torque msgs if the  the lines are not visible. It does make sense  and looks like Mazda chose to implement this feature.

I will update the PR with a comment.

Jafaral:
@rbiasini, I remembered some test  I did a while back. The torque msg itself has a "line not visible" bit . I did some more testing today,  it is that bit that controls whether the torque msgs are honored or not. Instead of just passing it as is I now just always unset that bit. See :
https://github.com/Jafaral/openpilot/commit/b1f150a035be7101f01c1c8ae26d5bca81ca55cd

That was enough to get the torque working regardless of the state of laneinfo msg. 

This PR can be closed.

rbiasini:
Thanks @Jafaral , seems like a cleaner explanation

PR #460, opened 2020-03-04
remove toyota ipas safety code and tests

rbiasini:


pd0wm:
-420 lines. I like it!

PR #461, opened 2020-03-05
remove unnecesary nissan angle check

rbiasini:


rbiasini:
na, this is needed so that the angle command starts from actual angle when OP is engaged

PR #462, opened 2020-03-06
Volkswagen safety updates: Phase 3

jyoung8607:
Third and hopefully last PR to review the overall change set proposed in #443.

- Bump torque limit from 2.5Nm to 3.0Nm [as previously discussed](https://github.com/commaai/panda/pull/443#discussion_r381085994)
- Get ACC state from TSK_06 (always on powertrain) rather than ACC_06 (could be on either)
- Matching update to regression tests
- Matching update to safety replay drive data

OP will need a matching update, to be filed shortly. Holding this PR in draft till that is done.


tests/safety_replay/test_safety_replay.py
```diff
@@ -18,7 +18,7 @@
   ("02ec6bea180a4d36|2019-04-17--11-21-35.bz2", Panda.SAFETY_HYUNDAI, 0), # HYUNDAI.SANTA_FE
   ("6fb4948a7ebe670e|2019-11-12--00-35-53.bz2", Panda.SAFETY_CHRYSLER, 0), # CHRYSLER.PACIFICA_2018_HYBRID
   ("791340bc01ed993d|2019-04-08--10-26-00.bz2", Panda.SAFETY_SUBARU, 0), # SUBARU.IMPREZA
-  ("b0c9d2329ad1606b|2019-11-17--17-06-13.bz2", Panda.SAFETY_VOLKSWAGEN_MQB, 0), # VOLKSWAGEN.GOLF
+  ("76b83eb0245de90e|2020-03-05--19-16-05.bz2", Panda.SAFETY_VOLKSWAGEN_MQB, 0), # VOLKSWAGEN.GOLF
```
rbiasini: which segment of this route has an engagement event?

jyoung8607: As examples, segment zero has an engagement, segment 4 has both a disengage and engagement.

jyoung8607:
This PR should pass CI on its own, or so I hope (not sure why it appears to bomb in Subaru) but commaai/openpilot#1147 should be merged before (or together) with the Panda ref update.

tests/safety_replay/test_safety_replay.py
rbiasini: ok, let me upload that segment and I bet this test will pass.

PR #468, opened 2020-03-10
Filtering LKAS HUD messages

avolmensky:


rbiasini:
Mmm... Safety test should be changed to not fail. Seems like the TX messages aren't checked by the test?

avolmensky:
@rbiasini added the fwd_hook and tx_msgs tests

rbiasini:
@avolmensky , ops, just added to master as well. You just need to fix the conflicts and we are good.

PR #474, opened 2020-03-28
Safety model for Volkswagen PQ35/PQ46/NMS

jyoung8607:
Cars this will eventually support, provided they have a Gen 3 EPS rack that supports HCA:

PQ35: https://en.wikipedia.org/wiki/Volkswagen_Group_A_platform#PQ35_(A5)
PQ46: https://en.wikipedia.org/wiki/Volkswagen_Group_B_platform#PQ46_(A6) (includes NMS)

This includes *most* older Volkswagen, lower-end Audi (A3/TT, Q3), SEAT and Škoda with EPS that aren't on the common MQB platform. Their networking concept is a first cousin of MQB, similar enough that we can share one reasonably clean OP port, but different enough to need a different Panda safety ID with different regression tests.

Only one of the four messages watched by Panda has a checksum and counter, which we are validating per ISO26262. The other three are watched for arrival rate only.

This code has many hours of road time from four different PQ tester vehicles. Full tests are included. The replay drive for safety_replay will be d12cd943127f267b|2020-03-27--15-57-18 segment 0, tested offline but naturally we'll need Comma's help to upload to the CI bucket.


VERSION
```diff
@@ -1 +1 @@
-v1.7.3
```
pd0wm: No newline in the VERSION string. You don't have to update it anymore. We now flash based on firmware hash. We should remove the version altogether.

jyoung8607: Fixed!

board/safety/safety_volkswagen.h
```diff
@@ -94,10 +132,21 @@ static void volkswagen_mqb_init(int16_t param) {
...
 
+static void volkswagen_pq_init(int16_t param) {
+  UNUSED(param);
+
+  controls_allowed = false;
+  relay_malfunction = false;
+  volkswagen_torque_msg = MSG_HCA_1;
+  volkswagen_lane_msg = MSG_LDW_1;
+
+  return;
```
pd0wm, line 143: Don't need return here

jyoung8607: Removed!

PR #476, opened 2020-03-31
first unsafe_mode PR

arne182:
If this works, more will follow


geohot:
Can you explain in a comment what this does?

arne182:
For safe, no disengage on brake, we rather check main on. So that the user has the chance to switch off the system with main off. With this first modification it opens up the ability to brake and not loose lateral control

geohot:
No disengage on brake will not be merged. The two shipped examples of an openpilot like system, Tesla Autopilot and GM Super Cruise, both disengage on brake.

PR #481, opened 2020-04-01
Unsafe mode gas disengage for VW MQB and PQ

jyoung8607:
Implement `UNSAFE_DISABLE_DISENGAGE_ON_GAS` for both Volkswagen MQB and PQ, with tests. Also fix a test compile warning (not VW specific) that's been annoying me for a while.

Regarding the other unsafe_mode flags:

`UNSAFE_DISABLE_STOCK_AEB`, N/A because Volkswagen never needs to disable stock FCW/AEB
`UNSAFE_ENABLE_WEAK_STEERING_WHILE_NOT_ENGAGED`, might do this at some point
`UNSAFE_RAISE_LONGITUDINAL_LIMITS_TO_ISO_MAX`, N/A until we take long control


geohot:
Thank you for writing tests.

PR #485, opened 2020-04-07
WIP: disengage on open door or seatbelt

quillford:
* [ ] all models
* [ ] tests

### DIFF
--- a/board/safety/safety_subaru.h
+++ b/board/safety/safety_subaru.h
@@ -128,6 +128,16 @@ static int subaru_rx_hook(CAN_FIFOMailBox_TypeDef *to_push) {
         (((addr == 0x122) && subaru_global) || ((addr == 0x164) && !subaru_global))) {
       relay_malfunction = true;
     }
+
+    // ensure all doors are closed
+    if ((addr == 0x3ac) && ((GET_BYTE(to_push, 4) & 0xF) != 0)) {
+      controls_allowed = 0;
+    }
+
+    // check if seatbelt is unlatched
+    if ((addr == 0x390) && (((GET_BYTE(to_push, 6) >> 2) & 1) == 1)) {
+      controls_allowed = 0;
+    }
   }
   return valid;
 }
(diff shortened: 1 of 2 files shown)


jyoung8607:
I'm not sure this is the right approach, or at least, it's not a complete approach. Today OP will notice things like no seatbelt, door open, wrong gear and go into SOFT_DISABLE, wherein the user is shown a clearly relevant urgent warning with audio, and OP disengages 3 seconds later. If this code were merged into Panda as it stands, we'd jump right to Controls Mismatch (a loss of clarity) with OP losing control immediately (a loss of safety).

Envision if you will, family of five cruising down the highway, OP engaged, chill. Curious youngster in back tugs on a door handle (or front passenger accidentally bumps the gearshift, whatever). Driver is confronted with a sudden and unexpected event. Their full attention is momentarily drawn to it. Other vehicle alarms may be going off in addition to OP's alarm. That is not a good moment to exacerbate the driver's workload.

I think Panda safety in this area is a good concept, but that Panda should approach it like OP does today. Track OP's NO_ENTRY events by gating the `controls_allowed` false->true transition on acceptable seatbelt/door/etc states. But if `controls_allowed` is already true when one of those events occurred, I think we need a timer based exit rather than immediate.

geohot:
I support the timer based exit, and if the issue is rectified within the timer, it doesn't disengage.

Hard disengages should only ever happen if we actually can't keep driving, they probably cause more safety issues themselves.

We are going to look into the standard and see what it says regarding this.

geohot:
After working through our internal safety analysis, we've decided to not merge this. The concern is around a driver exiting a stopped car, and a delay in this case would make the problem worse. Immediate disengage is also bad. The most basic thing we want is that the driver is alerted, which is already in openpilot.

Ideally, we also want to engage the EPB in these scenarios, so the car can't move even if it loses power. We are going to investigate what the stock systems do and look into the best way to do this.

Also, as a consequence of this discussion, we want to change openpilot to only disengage on driver door, not all the doors. We'd merge that PR :)

PR #488, opened 2020-04-08
toyota: use universal gas pressed bit

### LINKED ISSUE #404: ISO26262: Toyota gas message not present in Hybrid models
Use a common gas pressed indicator instead

quillford:
While refactoring the safety tests to use the can packer, I noticed that some toyota models (like prius) aren't included in the gas press check since they didn't have the `0x2c1` signal. This new signal is in the `PCM_CRUISE` message, so this should work for all toyota.

Ideally, there should be a test that tests each model and checks that panda safety and CarState agree to catch things like this.

Closes #404


rbiasini:
Nice, then this can be closed: https://github.com/commaai/panda/issues/404

pd0wm:
Nice!

I think we should also merge in a change to openpilot at the same time to avoid differences in controls allowed state. Can you make a PR to openpilot so that is also uses this message for `carState.gasPressed`?

PR #489, opened 2020-04-08
WIP: Use CANPacker for safety tests

quillford:
* easier to read/write (abstracted the low level can interface like the actual panda safety does)
* all safety tests now share a common set of tests, ensuring a minimum level of test coverage
* removes a lot of redundant test code

It might make sense to move the functions that make the particular can msgs to `selfdrive/car/<brand>/<brand>can.py`.

This should also make it pretty easy to test each panda safety against each supported model.


board/safety/safety_toyota.h
```diff
@@ -249,6 +249,7 @@ static int toyota_tx_hook(CAN_FIFOMailBox_TypeDef *to_send) {
 static void toyota_init(int16_t param) {
   controls_allowed = 0;
   relay_malfunction = 0;
+  gas_interceptor_detected = 0;
```
quillford, line 252: Is there a reason this wasn't done before?

pd0wm: It's initialized in safety declarations. But good catch, this could be an issue if you unplug a pedal without restarting the panda.

jyoung8607:
> volkswagen mqb

I'm a fan of this work! 👍 Something like this has been sorely needed IMO. You've probably noticed Volkswagen PQ can't be refactored the same way yet, because the OP port isn't upstreamed. 

Realistically I'm several weeks out from upstream, because there's a couple problems left to solve before it'll meet Comma's standards, but those problems aren't related to the safety or CAN packer code. Sometime in the next couple days, I'll open PRs to ship just the DBC and CAN message packer functions so you can bring this work all the way to the finish line.

quillford:
Thanks, @jyoung8607. I was actually about to ping you for a DBC.

I'll split this into a bunch of smaller PRs for each car brand, so it's easier to review.

PR #491, opened 2020-04-09
Safety Test Refactor: Toyota + support code

quillford:
This is the first of several PRs that break up #489 by car brand. This includes a refactored Toyota safety test and the support code. 

goals of the refactor:
* easier to read/write (abstracted the low level can interface like the actual panda safety does)
* all safety tests should share a common set of tests, ensuring a minimum level of test coverage
* remove redundant code

This should also make it pretty easy to test each panda safety mode against each supported model.


tests/safety/common.py
```diff
@@ -8,17 +10,133 @@ class UNSAFE_MODE:
...
+  FWD_BLACKLISTED_ADDRS = {} # {bus: [addr]}
+  FWD_BUS_LOOKUP = {}
+
+  def _rx(self, msg):
+    return self.safety.safety_rx_hook(msg)
+
+  def _tx(self, msg):
+    return self.safety.safety_tx_hook(msg)
+
+  def _brake_msg(self, brake): raise NotImplementedError
```
pd0wm, line 63: You can also `use abc.abstractmethod`

tests/safety/common.py
```diff
@@ -8,17 +10,133 @@ class UNSAFE_MODE:
...
+  def _speed_msg(self, speed): raise NotImplementedError
+
+  # TODO: is there a better way to do this?
+  # make linter happy
+  def assertTrue(self, c): raise Exception
```
pd0wm, line 69: Why not inherit PandaSafetyTest from TestCase?

quillford: If you do this the tests run in this class, but I found a fix for that.

PR #500, opened 2020-04-14
Safety Test Refactor: More test cases

adeebshihadeh:
Changes so far:

* Abstracted interceptor safety tests
* tests defaults like `gas_pressed_prev` both on panda init and after safety mode init

TODO
* [ ] TorqueSteeringSafetyTest
* [ ] AngleSteeringSafetyTest
* [ ] no TX on wrong safety mode (#470) 
* [ ] RX checksum check
* [ ] RX counter check


pd0wm:
Can we add a test that breaks the checksum, and make sure that get's caught by the safety code?

quillford:
@pd0wm

quillford:
I’m planning on adding checksum support to the can packer for each brand. There’s also an opendbc PR for manually setting the checksum so we can set invalid checksums without touching the bytes or knowing anything about the msg layout.

PR #507, opened 2020-04-15
Pedal gas pressed safety limits

robbederks:
* Made pedal calculation brand-specific
* Fixed Toyota pedal calculations
* Re-calculated limits based on OP limits + 5%


pd0wm:
LGTM. Let's test in the rav4 before we merge.

Issue #449, opened 2020-02-24
Some panda's don't go into dfu mode when trying to run openpilot/devel


pd0wm:
We've gotten multiple reports of pandad getting stuck in "waiting for DSU" when trying to reflash the bootstub. This seems very repeatable for those people. Haven't reproduced yet.

jyoung8607:
Per @robbederks / f07a6ee we need to bounce through the bootstub first to reach DFU mode from release builds of Panda firmware. I haven't examined the boardd startup process to see why this isn't happening successfully during normal startup (or if this is something that should work by design) but making the same fix to `make recover`, which uses different code in `build.mk` to enter DFU, I was eventually able to get an affected user fixed.

This did take two tries to get into DFU, or there may be another state transition required here that I don't see or understand that was necessary to reach DFU mode.

```
root@localhost:/data/openpilot/panda/board$ make recover
arm-none-eabi-gcc -MT obj/bootstub.panda.o -MMD -MP -MF generated_dependencies/bootstub.Td -g -Wall -Wextra -Wstrict-prototypes -Werror -mlittle-endian -mthumb -mcpu=cortex-m4 -mhard-float -DSTM32F4 -DSTM32F413xx -mfpu=fpv4-sp-d16 -fsingle-precision-constant -I inc -I ../ -nostdlib -fno-builtin -std=gnu11 -Os -Tstm32_flash.ld "-DEON" "-DALLOW_DEBUG" -o obj/bootstub.panda.o -c bootstub.c
arm-none-eabi-gcc -g -Wall -Wextra -Wstrict-prototypes -Werror -mlittle-endian -mthumb -mcpu=cortex-m4 -mhard-float -DSTM32F4 -DSTM32F413xx -mfpu=fpv4-sp-d16 -fsingle-precision-constant -I inc -I ../ -nostdlib -fno-builtin -std=gnu11 -Os -Tstm32_flash.ld "-DEON" "-DALLOW_DEBUG" -o obj/sha.panda.o -c ../crypto/sha.c
[...]

PR #509, opened 2020-04-16
Match Panda DFU entry fix in "make recover" process

jyoung8607:
Title says it all. This helped at least one user get unstuck from Panda refusing to accept DEBUG signed firmware by flashing it via DFU from the command line. Original update process fix taken from f07a6ee.

This fix may help users experiencing #449.


robbederks:
Nice!
Could  you change it to a one-liner? i.e.:
`from python import Panda; Panda().reset(enter_bootstub=True); Panda().reset(enter_bootloader=True)`

PR #511, opened 2020-04-18
Fix definition naming. Was this a typo?

csouers:
AFAIK, RDLR and RDHR are 4 bytes each.


robbederks:
Depends on how you interpret the range: 04 probably referred to a range [0:4[, not [0:4]
In most programming languages, slicing is done like in the code here already (i.e. in python: https://stackoverflow.com/a/509295)

PR #515, opened 2020-04-24
Block 0xe5 (Honda Bosch) at the panda/uno. Only allow static values.

csouers:


robbederks:
Could you make the misra tests pass, and also fix the failing safety tests too?

csouers:
Done. I don't have a specific test there for e5 as my attempts would fail for some unknown reason. I'm happy to look into this more, if needed.

csouers:
PS. This panda safety change requires TWO other PRs to be merged in to not break honda bosch. 
https://github.com/commaai/openpilot/pull/1420
and https://github.com/commaai/opendbc/pull/250

board/safety/safety_honda.h
```diff
@@ -239,6 +239,14 @@ static int honda_tx_hook(CAN_FIFOMailBox_TypeDef *to_send) {
     }
   }
 
+  // Bosch supplemental control check
+  if (addr == 0xE5) {
+    int e5_state = (GET_BYTE(to_send, 0) << 24U) | (GET_BYTE(to_send, 1) << 16U) | (GET_BYTE(to_send, 2) << 8U) | GET_BYTE(to_send, 3);
```
rbiasini, line 244: use `GET_BYTES_04` macro instead?

csouers: Tried that. Doesn't allow the message through.
`242   // Bosch supplemental control check
243   if (addr == 0xE5) {
244     int e5_state = GET_BYTES_04(to_send);
245     if (e5_state != 0x04008010) {
246       tx = 0;
247     }
248   }`

rbiasini: It's an endianness issue. Try this:
```
if (addr == 0xE5) {
  if (GET_BYTES_04(to_send) != 0x10800004) {
    tx = 0;
  }
}
```

csouers: Came to the same conclusion. This works.

pd0wm:
Can you also add a check for bytes 4, 5 and 6? Since we don't know what they are it might be safer to force those to zero as well.

csouers:
> Can you also add a check for bytes 4, 5 and 6? Since we don't know what they are it might be safer to force those to zero as well.

Yes. I'll also put up a PR for opendbc adding those as well and fixup the the openpilot PR too.

PR #524, opened 2020-05-03
FCA Fix: When initializing all the CAN busses, make sure they are also cleared

j-vanetten:
FCA vehicles work great the first time the panda relay is activated after the panda is reset. However, then subsequent ignitions cause ignition aborts, powers steering failures, and LKAS faults.  This is due to messages remaining in the busses from the previous ignition and get sent the next time a vehicle is started.  

This solution always clears the busses when the mode changes.


robbederks:
Looks good! Will run through our internal CI and then merge

robbederks:
Merged the CI PR!

PR #526, opened 2020-05-04
Fix CAN Ignition for Black Panda and Uno

andrewcopenpilot:
Both Black Panda and Uno transceiver enabled code needed to be updated to match ODB-C orientation. Otherwise it only works in one orientation.

Uno needed to be updated to leave the transceiver mapped to BUS 0 enabled while sleeping to listen for CAN based ignition.

### DIFF
--- a/board/boards/black.h
+++ b/board/boards/black.h
@@ -2,7 +2,7 @@
 // Black Panda + Harness //
 // ///////////////////// //
 
-uint8_t transciever_lookup[] = {1U, 2U, 3U, 4U}; // Map transciever to bus in default orientation
+uint8_t black_transciever_lookup[] = {1U, 2U, 3U, 4U}; // Map transciever to bus in default orientation
 
 void black_enable_can_transciever(uint8_t transciever, bool enabled) {
   switch (transciever){
@@ -27,7 +27,7 @@ void black_enable_can_transciever(uint8_t transciever, bool enabled) {
 void black_enable_can_transcievers(bool enabled) {
   uint8_t first_bus = enabled ? 0U : 1U;  // leave transciever for bus 0 enabled to detect CAN ignition
   for(uint8_t bus=first_bus; bus<=3U; bus++) {
-    uno_enable_can_transciever(transciever_lookup[bus], enabled);
+    black_enable_can_transciever(black_transciever_lookup[bus], enabled);
   }
 }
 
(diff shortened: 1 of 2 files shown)


robbederks:
Great catch! Definitely a bug.

I would however rather see this fixed in the `uno_enable_can_transcievers` function to not make the other function confusing to use.
Thanks!

andrewcopenpilot:
@robbederks what do you think of this setup?

robbederks:
Seems a little bloated.
Wouldn't something like this suffice?
```
void uno_enable_can_transcievers(bool enabled) {
  for(uint8_t i=0U; i<=4U; i++){
    // Leave main CAN always on for CAN-based ignition detection
    if((car_harness_status == HARNESS_STATUS_NORMAL) ? (i == 0U) : (i == 2U)){
      uno_enable_can_transciever(i, true);
    } else {
      uno_enable_can_transciever(i, enabled);
    }
  }
}
```

andrewcopenpilot:
@robbederks I think what you suggested would work, in previous iterations I'd tried to lose the ternary operator as I think they are harder to read. Now that I've thought about it more I don't think its an issue for it to falp for a few cycles on calling enable_can_transcievers.  What do you think about this?

```void uno_enable_can_transcievers(bool enabled) {
  for(uint8_t i=1U; i<=4U; i++){
    uno_enable_can_transciever(i, enabled);
  }
  
  // Leave main CAN always on for CAN-based ignition detection
  if (car_harness_status == HARNESS_STATUS_FLIPPED) {
    uno_enable_can_transciever(3U, true);
    return;
  }
  uno_enable_can_transciever(1U, true);
}```

**Comment by robbederks** [comment]
I'd rather not have the GPIO flip for a few cycles just to increase the readability. Also, iirc, no early returns are permitted by the MISRA checking ;)

For me the ternary operator is entirely fine and clean imo. We use them all over the place anyways

**Comment by andrewcopenpilot** [comment]
@robbederks, fair enough, I'm kinda new to embedded development. I'm not really sure of what the implications are of flapping a GPIO pin. If we don't want to do that I don't have a better idea than what you suggested.

The one tweak we might make is to swap to HARNESS_STATUS_FLIPPED. Otherwise if this code was ever combined into a common library for all panda types, a NULL harness state would result in leaving transceiver 3 on. Something like below:

```
void uno_enable_can_transcievers(bool enabled) {
  for(uint8_t i=0U; i<=4U; i++){
    // Leave main CAN always on for CAN-based ignition detection
    if((car_harness_status == HARNESS_STATUS_FLIPPED) ? (i == 2U) : (i == 0U)){
      uno_enable_can_transciever(i, true);
    } else {
      uno_enable_can_transciever(i, enabled);
    }
  }
}
```

If that looks good to you, I'll test and update the PR.

robbederks:
Looks great! Thanks!

andrewcopenpilot:
@robbederks tested this most recent commit on black panda and Uno in both orientations and all checked out.

robbederks:
Cool, thx. Will get our internal hardware tests to run it on a local PR, and will merge if it checks out!

PR #529, opened 2020-05-08
Safety: message length check on RX and TX

adeebshihadeh:
* [x] chrylser
* [x] gm
* [x] honda
* [x] hyundai
* [x] nissan
* [x] subaru
* [x] toyota
* [x] vw mqb
* [x] vw pq


adeebshihadeh:
@jyoung8607 do you have the message lengths for the VW PQ? It'd be nice if we had a dbc for it upstreamed, even if it's not complete, especially since we have to maintain this panda code.

jyoung8607:
> @jyoung8607 do you have the message lengths for the VW PQ?

Yep, commaai/opendbc#259 for the latest. Let me know if I can help in any other way. I think we may have discussed upstreaming the openpilot CAN packer components as well, for your earlier test refactoring, I can do that if desired.

adeebshihadeh:
@jyoung8607 I'd really like to get all the checksums in the CAN packer and have the VW PQ tests refactored. You can do the packer stuff in a separate PR on opendbc.

PR #535, opened 2020-05-14
GitHub Actions

adeebshihadeh:
* [x] build
* [x] safety
* [x] safety replay
* [x] language
* [x] python linters
* [x] misra
* [x] update shield in README


adeebshihadeh:
CircleCI is causing CI to report failure because its config is deleted in this branch. Circle should stop running when this is merged and its config is gone from master.

PR #541, opened 2020-05-24
Refactor addr check

adeebshihadeh:
Needed to cleanly add the gas signal to the RX checks for hyundai EVs, which doesn't have a counter

Changes:
* check_checksum, max_counter, and expected_timestep are now properties of the individual can message, not all the allowed messages for an RX check
* before when multiple msgs were allowed, either could update the RX check struct, however now it only uses the first one it sees for the subsequent checks

### DIFF
--- a/board/safety.h
+++ b/board/safety.h
@@ -100,28 +100,38 @@ int get_addr_check_index(CAN_FIFOMailBox_TypeDef *to_push, AddrCheckStruct addr_
 
   int index = -1;
   for (int i = 0; i < len; i++) {
-    for (uint8_t j = 0U; addr_list[i].msg[j].addr != 0; j++) {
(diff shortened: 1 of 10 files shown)


board/safety.h
```diff
@@ -100,15 +100,25 @@ int get_addr_check_index(CAN_FIFOMailBox_TypeDef *to_push, AddrCheckStruct addr_
...
+        if ((addr == addr_list[i].msg[j].addr) && (bus == addr_list[i].msg[j].bus) &&
+              (length == addr_list[i].msg[j].len)) {
+          addr_list[i].index = j;
+          addr_list[i].msg_seen = true;
+          break;
+        }
       }
     }
+
+    if ((addr == addr_list[i].msg[addr_list[i].index].addr) &&
```
pd0wm, line 115: pull out `addr_list[i].index` into a variable

PR #543, opened 2020-05-27
document tx message addresses better

AskAlice:


pd0wm:
This PR is also adding TX messages for ACC. We can't add those without proper safety checks. For the safety we need to do something similar to honda bosch: https://github.com/commaai/panda/pull/268/files. Note how we added a safety mode param to explicitly allow longitudinal control

AskAlice:
fixed

PR #544, opened 2020-05-27
Fix length of 0x20b in NISSAN_TX_MSGS

avolmensky:
ACC wasn't being cancelled when gas was pressed


adeebshihadeh:
Hmm, safety replay should have caught this, but maybe the segment we use for the test doesn't have a gas press cancellation event. Can you PM me a replacement route on discord that you'd be ok with making public that covers all the TX msgs?

PR #545, opened 2020-05-29
Add pre commit checks + CI

pd0wm:


Dockerfile.panda
```diff
@@ -46,7 +47,8 @@ RUN pyenv install 3.7.3 && \
     pyenv global 3.7.3 && \
     pyenv rehash && \
     pip install --no-cache-dir --upgrade pip==18.0 && \
-    pip install --no-cache-dir -r /tmp/requirements.txt
+    pip install --no-cache-dir -r /tmp/requirements.txt && \
+    pip install pre-commit==2.4.0 pylint==2.5.2
```
adeebshihadeh, line 51: --no-cache-dir, probably doesn't make a big different but good practice

adeebshihadeh: Actually, why don't they just get added to the requirements?

pd0wm: I’ll add them to the requirements!

PR #557, opened 2020-06-15
Add Subaru Outback route to replay tests

martinl:
Subaru Outback 2017, route id: 5ab784f361e19b78|2020-06-08--16-30-41


adeebshihadeh:
Had to fix the file name, added in #558

PR #563, opened 2020-06-27
J2534 iso14230

gregjhogan:
HDS tested and working for both ISO14230 and ISO9141 (both fall back to fast init)


gregjhogan:
new installer: [panda J2534 driver install.zip](https://github.com/commaai/panda/files/4844574/panda.J2534.driver.install.zip)

gregjhogan:
depends on #565

PR #565, opened 2020-06-29
k-line 5 baud init

gregjhogan:
looks good on logic analyzer (matches tactrix), but haven't found a place where HDS uses this and actually gets a response


robbederks:
Actually, maybe instead of doing while, can't you put it in some kind of timer interrupt? Because you're actually doing the wait in an interrupt itself, so main and lower priority interrupts won't work anymore?

board/main.c
```diff
@@ -628,6 +630,15 @@ int usb_cb_control_msg(USB_Setup_TypeDef *setup, uint8_t *resp, bool hardwired)
         heartbeat_counter = 0U;
         break;
       }
+    // **** 0xf4: k-line/l-line 5 baud initialization
+    case 0xf4:
+      k_wakeup = (setup->b.wValue.w == 0U) || (setup->b.wValue.w == 2U);
```
robbederks, line 635: Would maybe add a check here to make sure we are on a white/grey panda. On black/uno, GPIOC12 is used for other purposes, so that might screw with it

PR #566, opened 2020-06-30
USB enumeration detection fix

robbederks:
On some devices, the suspend condition sometimes false-triggers as enumerated when the host is not connected due to noise on the USB lines. Trying to fix by checking if we actually rx the SOF packets (1/ms)


robbederks:
Now it also checks that it received an end-of-periodic-frame interrupt. This only seems to fire if the packet it saw on the bus was valid, and thus a periodic frame has started

PR #567, opened 2020-06-30
WIP: Improve board detection stability

robbederks:
Uno sometimes identifies as Dos. Suspecting that the detection line is not pulled high long enough to generate a logical 1.
Needs to be tested.


robbederks:
Seems to work in empirical testing, although on the old version it was also very hard to get it into the wrong state.
Can't hurt anyways, will be watching the logs to see if the amount of mis-detections goes down

PR #568, opened 2020-07-05
README update

adeebshihadeh:
preview here https://github.com/commaai/panda/blob/update-readme/README.md


adeebshihadeh:
@robbederks might want to add something at the top about the newer pandas after white

robbederks:
Kk, will add something tomorrow

PR #575, opened 2020-07-23
Change Subaru safety test to use generated dbc

martinl:
Use generated Subaru global dbc in safety test, related to https://github.com/commaai/opendbc/pull/277


adeebshihadeh:
This also needs to update the pinned commit from openpilot which is how opendbc gets checked out.

PR #579, opened 2020-08-03
Volvo safety model V40 and EUCD

danielzmod:
This is the current safety model i wrote for V40 (C1MCA platform) and EUCD-platform vehicles. I have tried to copy the style and reuse functions from Nissan safety model to have the same style.


danielzmod:
Hi! I haven't received any feedback yet. Is there something that needs changing?

adeebshihadeh:
See the openpilot PR for comments

PR #580, opened 2020-08-04
Clock source driver

robbederks:


robbederks:
Can be merged when https://github.com/commaai/openpilot/pull/2036 is merged

PR #582, opened 2020-08-07
update gas pressed bit to TCS13 DriverOverride

xps-genesis:
It has been found that a common gas pressed bit is required to satisfy all three powertrains configuration on Hyundai.
TCS13 - DriverOverride signal shows driver gas pressed and it is available on ICE, HEV and EV.
EMS16 - CF_Ems_AclAct signals shows driver gas pressed on ICE manual transmission. TCS13 is unavailable on manual vehicles. Keeping this check here for future support for Manual trans Hyundai vehicles.


adeebshihadeh:
Superseded #583. The necessary changes are more involved than this.

PR #584, opened 2020-08-08
socketcan: usb bulk mode for rx

gregjhogan:
There are some good tools that support linux SocketCAN and because our driver was using interrupt mode it can't keep up heavy traffic on the CAN bus.  This should fix that issue and some other miscellaneous cleanup.

### DIFF
--- a/drivers/linux/Makefile
+++ b/drivers/linux/Makefile
@@ -1,8 +1,7 @@
 VERSION=0.0.1
 obj-m+=panda.o
 
-link:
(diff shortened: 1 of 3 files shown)


adeebshihadeh:
Good to merge?

gregjhogan:
it was, but looks like there are conflicts that need to be fixed now

adeebshihadeh:
The conflict is pretty small. Can you fix and merge or just close for now?

PR #585, opened 2020-08-08
Updating X-Trail/Rogue gas_pressed threshold

avolmensky:
@adeebshihadeh Updating gas pressed threshold in panda as requested in commaai/openpilot#1995


adeebshihadeh:
Thanks, @avolmensky!

PR #598, opened 2020-09-21
fix "du: illegal option -- b" error on macos

vell001:
http://manpages.ubuntu.com/manpages/precise/en/man1/du.1.html


adeebshihadeh:
Shouldn't be an issue after ad9ecefe65ce4ea7bcaf3bfe692ee2a3fc6b0b63

PR #609, opened 2020-11-03
Update safety_toyota.h

birdman6450:
Tss2 limit, new values do not fault. Allows code to be edited on shanes SA. My indi tune with this is the best so far and i need to gather 5 routes to submit full tune PR. This is the first step. Does not effect vehicles just allows edits for routes for PR. Improved handing for corolla tss2 tune.

### DIFF
--- a/board/safety/safety_toyota.h
+++ b/board/safety/safety_toyota.h
@@ -3,8 +3,8 @@ const int TOYOTA_MAX_TORQUE = 1500;       // max torque cmd allowed ever
 
 // rate based torque limit + stay within actually applied
 // packet is sent at 100hz, so this limit is 1000/sec
-const int TOYOTA_MAX_RATE_UP = 10;        // ramp up slow
-const int TOYOTA_MAX_RATE_DOWN = 25;      // ramp down fast
+const int TOYOTA_MAX_RATE_UP = 15;        // ramp up slow
+const int TOYOTA_MAX_RATE_DOWN = 44;      // ramp down fast
 const int TOYOTA_MAX_TORQUE_ERROR = 350;  // max torque cmd in excess of torque motor
 
 // real time torque limit to prevent controls spamming


geohot:
Explain to me what safety testing you have done to confirm this is okay. A good tune is not safety.

Please read: https://medium.com/@comma_ai/safer-control-of-steering-362f3526c9ab

jamcar23:
@geohot do you / comma have a list of actionable objectives the community can take to prove something is safe? In the article you mention "Safety is when you have thought through every possible scenario, tested the worst possible thing that can happen, and it was still safe." but I don't it's possible or realistic to think through and tests literally all possible outcomes. 

I'd imagine comma has a specific set of situations you tested for when these values where found the first time. Can comma release info on those so that the community can benefit? It would not only allow us to prove changes like this are still on the path to safety but it would also allow the community to make safer car ports. 

Also, I imagine we would need to upload logs at some point for comma to review before accepting something like this, correct?  Would making this change in an otherwise acceptable fork be a bannable offense? If so, what's the process to upload logs when the change is bannable but may (eventually) be accepted in a PR?

birdman6450:
> Explain to me what safety testing you have done to confirm this is okay. A good tune is not safety. Not faulting the car is not safety.
> 
> Please read: https://medium.com/@comma_ai/safer-control-of-steering-362f3526c9ab

No nshta pertinent documents violations. No safety.md violations. No safer-control-of-steering paper violations. Does not exceed manufacturers limitations. Test driven in all situations repeatedly and a significantly higher probability of accurate ALC is executed. Yes, merged gated behind, "unsafe mode" route data collection and analysis can be submitted. Via extensive route data and statistics the complete tune can then be thoroughly scrutinized for total LDW , cmd vs. eps unit accuracy, and total lateral overrides. If any safety guidlines are in violation please advise.

geohot:
One of the things to do is injection testing. Have a driver using openpilot normally then have someone in the back seat with a joystick inject the max torque allowed in the panda profile. This is the same as a model failure, which we consider to be QM. If the driver has any trouble keeping control of the car or deviates from the lane, it's too much.

https://github.com/commaai/openpilot/blob/master/SAFETY.md

Also, in general, less torque and less ramp up rate is safer. Show exactly when and why you need more. Many times the problem is tuning, not the limits.

jamcar23:
Thanks for reaching out again! 

Proving this is safe with injection testing is a great objective for us to work towards.

As far as graphs go, are you talking about cabana graphs? If so, does that mean we won't get banned for making this change in a fork?

geohot:
We should clarify what gets you banned:

1. Not disengaging on brake.
2. Disabling or seriously nerfing driver monitoring.

These are two very clear rules you should never have any reason to violate.

Otherwise, use your best judgement. If one person is testing stuff that doesn't violate the above two rules, they won't be banned. If we see 50 people suddenly using higher torques, that might be a problem as well, but for an individual/small group, this is how we learn and grow.

The safety model isn't set in stone, it will need to grow with time. And if people understand what they are doing and can experiment safely with it, more power to them. The concern is around people copying and pasting a command they don't understand the implications of.

jamcar23:
Thanks for clarifying the safety model. I was under the impression that changing any panda code that didn't have an already defined unsafe flag would get you banned. Glad to know that's not the case

geohot:
Not for experimentation, no. If you are shipping a fork to 100 people though, that's a different story.

SCshredder17:
With a driver response time of less than 1 second (.75 seconds) maintaining control of the vehicle during an injection test is calculated to be reasonable. If the vehicle departs from the lane we can tune the steer up to have a max torque of 1 second instead of .75 seconds (steer up 15) or 1.5 seconds (steer up 10). Data from injection testing pending.

adeebshihadeh:
Closing due to inactivity. Feel free to re-open once testing is done.

PR #615, opened 2020-12-16
Honda Bosch: Support extended LKAS_HUD IDs

csouers:
Honda has started using some extended IDs on 2021+ Bosch Hondas on the private CAN. They have also modified how the checksum is being calculated for the new messages (based on the two IDs that we're dealing with; 0x33DA and 0x33DB). The method is the same as before, but they do additional math if the ID is extended.

I know the extended IDs aren't being checked in panda, but it's there if needed. We run the same checksum changes in opendbc with success.

Some message examples:
```
- Original -
0x33d: 0000104000
0x33d: 000010401f
0x33d: 000010402e
0x33d: 000010403d

- Modified sum -
0x33da: 0000104009
0x33da: 0000104018
0x33da: 0000104027
0x33da: 0000104036

0x33db: 0000104000000008
0x33db: 0000104000000017
0x33db: 0000104000000026
0x33db: 0000104000000035
```

### DIFF
--- a/board/safety/safety_honda.h
+++ b/board/safety/safety_honda.h
@@ -61,7 +61,6 @@ static uint8_t honda_compute_checksum(CAN_FIFOMailBox_TypeDef *to_push) {
   int len = GET_LEN(to_push);
   uint8_t checksum = 0U;
   unsigned int addr = GET_ADDR(to_push);
-  bool extended = (addr > 0x7FFU);
   while (addr > 0U) {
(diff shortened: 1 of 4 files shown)


adeebshihadeh:
I removed the change to the checksum since it's unused. We can add it once it's needed.

csouers:
> I removed the change to the checksum since it's unused. We can add it once it's needed.

Got it. Thanks for getting this merged in!

PR #616, opened 2021-01-02
hyundai longitudinal safety

gregjhogan:
copied logic from honda bosch longitudinal safety
* same accel limits
* currrently checking aReqRaw for limits do we need to also check aReqValue?
* brought over `current_controls_allowed` logic (pedal + moving) not sure why this wasn't already there
* used safety param value of 1 to enable

### DIFF
--- a/board/safety/safety_hyundai.h
+++ b/board/safety/safety_hyundai.h
@@ -6,15 +6,25 @@ const int HYUNDAI_MAX_RATE_DOWN = 7;
 const int HYUNDAI_DRIVER_TORQUE_ALLOWANCE = 50;
 const int HYUNDAI_DRIVER_TORQUE_FACTOR = 2;
 const int HYUNDAI_STANDSTILL_THRSLD = 30;  // ~1kph
+const int HYUNDAI_MAX_ACCEL = 200;  // max accel ==  2.0m/s2
+const int HYUNDAI_MIN_ACCEL = -350; // max decel == -3.5m/s2
 const CanMsg HYUNDAI_TX_MSGS[] = {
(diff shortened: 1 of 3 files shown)


gregjhogan:
this is probably good to merge now

board/safety/safety_hyundai.h
```diff
@@ -252,6 +288,7 @@ static void hyundai_legacy_init(int16_t param) {
   relay_malfunction_reset();
 
   hyundai_legacy = true;
+  hyundai_longitudinal = false;
```
adeebshihadeh, line 291: why not allow long for the legacy safety mode?

gregjhogan: addresses could be completely different and I have no way to test, what I have now I know is sufficient at least for my car to not have any errors (which I didn't get on the first try)

adeebshihadeh: hyundai legacy is the same as the regular hyndai safety mode, except with a subset of the counter/checksum checks

gregjhogan: ok, I added it

gregjhogan:
I take it back, it isn't ready yet ...
* SCC12 RX check can only happen for stock longitudinal (since radar sends it).  We could add two more safety modes instead of using the safety param (since RX check list isn't currently dynamic), make RX check list dynamic, we might be able to switch to always using TCS13->ACC_REQ, or switch to always using button press detection for enable/disable.
* How do we feel about using button press detection for setting `controls_allowed` when it doesn't have a checksum?

gregjhogan:
This can be finished whenever a decision is made on https://github.com/commaai/panda/pull/625

PR #618, opened 2021-01-04
Panda Safety Improvement: Toyota BRAKE_MODULE2 

TheInventorMan:
This safety improvement is in conjunction with the following improvement to the openpilot repo: https://github.com/commaai/openpilot/pull/19638 

Adds detection of the brake_pressed signal in the backup brake_module2 message for an added layer of redundancy.

In-car testing required as of Jan 4, 2021


adeebshihadeh:
We still need to find data that confirms this. Will leave the openpilot PR up and close this one until we confirm.

PR #622, opened 2021-01-24
hyundai: WHL_SPD11 checksum

gregjhogan:
note that they excluded the counter from the checksum

PR #623, opened 2021-01-25
hyundai: TCS13 checksum

gregjhogan:
It was pure coincidence that this checksum ever worked.  Makes me wonder if there are any cars that openpilot doesn't work because of this bug.

This is stacked on top of https://github.com/commaai/panda/pull/622 (<- please merge first)

Good example of how adding up the bytes is a horrible checksum!


adeebshihadeh:
There's merge conflicts after merging #622. Can you rebase?

gregjhogan:
done

Issue #587, opened 2020-08-08
Build Windows J2534 driver in CI


adeebshihadeh:
Use the github actions windows runner for https://github.com/commaai/panda/tree/master/drivers/windows#building-the-project

PR #624, opened 2021-01-27
get value by reference in ranged based loop

deanlee:


robbederks:
Will merge because this is technically correct, but this code is very rarely used and untested.
It's also irrelevant for openpilot.

deanlee:
OIC

adeebshihadeh:
It would be great if this was tested though. @deanlee interested in getting CI up for building the J2534 driver (#587)? GitHub Actions should have a windows runner.

PR #625, opened 2021-01-31
support using safety mode param to control rx checks

gregjhogan:
remove need to create a new safety mode if you only want to change the rx check list
(planning to use for finishing https://github.com/commaai/panda/pull/616)


gregjhogan:
so is this an acceptable approach (would be used for hyundai longitudinal safety without adding a new safety mode)?

jyoung8607:
Piling on: the Panda safety for VW MQB longitudinal control will be logically equivalent to #616, and will have a similar need to tweak the RX check list.

PR #630, opened 2021-02-28
honda: disable controls with main button

spektor56:
On honda, the cancel button only cancel's ACC and not LKAS.  The main button cancels both ACC and LKAS and is the real "main" control for these features.  Whenever the main button is pressed, the controls_allowed should be false regardless of the state the main switch is in.  This is an additional safety measure and matches stock functionality.


adeebshihadeh:
Superseded by #789

PR #633, opened 2021-03-17
add safetyParam to health

pd0wm:


board/safety.h
```diff
@@ -303,6 +304,7 @@ int set_safety_hooks(uint16_t mode, int16_t param) {
   }
   if ((set_status == 0) && (current_hooks->init != NULL)) {
     current_hooks->init(param);
+    current_safety_param = param;
```
adeebshihadeh, line 307: why set this here instead of where `current_safety_mode` is set?

pd0wm: Since the init function could be NULL and it wouldn't actually be used.

pd0wm: But not sure about desired behavior in that case.

PR #634, opened 2021-03-22
support for Tesla Model S with AutoPilot (AP1/AP2)

BogGyver:


robbederks:
Merged the other safety PR #629. If there's stuff missing there, please update this one based on that.

PR #640, opened 2021-04-04
CAN calibration protocol (CCP) python library

gregjhogan:


robbederks:
Does this library work? If so, I don't see why it shouldn't be merged with the linter errors fixed

gregjhogan:
at least parts of it work but it isn't thoroughly tested

also, a lot of it might be useful for XCP with some re-organization, but that can be done later

PR #643, opened 2021-05-02
Subaru global: Enable Stock LKAS during disengagement.

ClockeNessMnstr:
Subaru_Global specific update other than fwd_hook blacklist test. 

common safety test for blacklisting failed when allowing messages through during disengagement. 
This test may be better performed with "self.safety.set_controls_allowed(1)" to be asserting that the two systems are not active together. 

subaru safety test asserted that a torque request with "0" torque should be allowed within tx_hook when disengaged. 
I updated to allow blocking all torque messages when disengaged. To my knowledge, there should not be a reason to inject any messages when OP is disengaged. 

during disengagements the safety of the vehicle is being affected. It is less safe because OP is blocking LKAS and not replacing it.
Stock LKAS appears to be active but is disabled silently. additionally, it does not need to be. 

I updated the forwarding logic to only apply the blacklist when OP is engaged. 
I updated the tx_hook to only allow messages when OP is engaged. 

I had to whitelist the 0x221 message to preserve disengage_on_gas behavior. otherwise stock ACC remains active on subaru when OP disengages.

### DIFF
--- a/board/safety.h
+++ b/board/safety.h
@@ -208,7 +208,7 @@ void generic_rx_checks(bool stock_ecu_detected) {
   }
   gas_pressed_prev = gas_pressed;
 
-  // exit controls on rising edge of brake press
+  // exit controls on rising edge of brake press or if pressed while moving. 
   if (brake_pressed && (!brake_pressed_prev || vehicle_moving)) {
     controls_allowed = 0;
   }
--- a/board/safety.h
+++ b/board/safety.h
@@ -208,11 +208,20 @@ void generic_rx_checks(bool stock_ecu_detected) {
   }
   gas_pressed_prev = gas_pressed;
 
-  // exit controls on rising edge of brake press or if pressed while moving. 
-  if (brake_pressed && (!brake_pressed_prev || vehicle_moving)) {
+  // exit controls on rising edge while vehicle is not moving
+  if (brake_pressed && !brake_pressed_prev && !vehicle_moving) { 
+    controls_allowed = 0;
+	brake_pressed_prev = brake_pressed; // no debounce when not moving
+  } else {
+      if (!(brake_pressed == brake_pressed_prev)) { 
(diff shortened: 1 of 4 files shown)


adeebshihadeh:
This is intentional. When disengaged, openpilot will send the LKAS messages without any actuation requests, which the panda enforces.

ClockeNessMnstr:
> This is intentional. When disengaged, openpilot will send the LKAS messages without any actuation requests, which the panda enforces.

What's the stance on the loss of a stock safety feature when disengaged?

adeebshihadeh:
When openpilot is enabled, it completely replaces your stock LKAS. openpilot is an all or nothing system, and this is enforced even when using stock ACC.

ClockeNessMnstr:
> When openpilot is enabled, it completely replaces your stock LKAS. openpilot is an all or nothing system, and this is enforced even when using stock ACC.

Thanks, Even on a long vehicle it would be beneficial to have stock lkas as a fallback when OP disengaged. 

My concern is that removing stock LKAS even while OP is disengaged makes the vehicle less safe than stock during disengagements.

I get it if it's just a hard and fast rule that OP means no stock functions. However it can only be an added risk to be disabling safety features when disengaged.

ClockeNessMnstr:
jyoung86 pointed out that the stock LKAS is not a safety feature. contrarily, FCW is preserved even when OP is active as a safety feature. 

However, this is still a significant quality improvement to have stock LKAS when OP is disengaged. 
I understand that some vehicles may not work as easily and that it may even need some checks for my vehicle which appears to work, but it seems like something that should be allowed to be worked on. doing this right is still a positive thing.

ClockeNessMnstr:
PR branch is out of sync. I raised some questions to adeeb on discord. Will need a new PR if we want to proceed. sounds like it would just be upstream.

PR #645, opened 2021-05-10
Debounce Brake signal for single frame errors.

ClockeNessMnstr:
Single messages for user braking are appearing on Subaru vehicles causing unsafe ACC cancelation.
The vehicle cancels ACC and "coast" into/towards a vehicle that should be being braked for.
It seems to happen with moderate braking when a new lead is acquired.
It is also appearing more frequently with gas pedal enabled. 

My testing is only on subaru_global but change preserves the previous logic to ensure compatibility with all vehicles.
When vehicle is in motion this checks for 2 consecutive brake messages before exiting controls.
This prevents ACC from being canceled by itself if OP or stock ACC is actively braking. I do not know if this would be an issue with long vehicles but I imagine it would be similar behavior.

Openpilot PR https://github.com/commaai/openpilot/pull/20866

### DIFF
--- a/board/safety.h
+++ b/board/safety.h
@@ -208,10 +208,15 @@ void generic_rx_checks(bool stock_ecu_detected) {
   }
   gas_pressed_prev = gas_pressed;
 
-  // exit controls on rising edge of brake press
-  if (brake_pressed && (!brake_pressed_prev || vehicle_moving)) {
+  // exit controls on rising edge while vehicle is not moving
+  if (brake_pressed && !brake_pressed_prev && !vehicle_moving) { 
     controls_allowed = 0;
-  }
+  } 
+  // exit controls while in motion if brake is pressed.
+  // debounce rising edge
+  if (brake_pressed && brake_pressed_prev && vehicle_moving) {
+	controls_allowed = 0;
+  } 
   brake_pressed_prev = brake_pressed;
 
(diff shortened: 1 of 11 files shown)


ClockeNessMnstr:
safety_subaru.h is updated to use the debounced signal with 54d0709.
other vehicles were update with ", true, true" to ignore the changes. 
they can be easily updated to match their respective logic used to check brake / gas signal updates. 

MartinL has been testing different signals for the brake status on his subaru branches. 
In testing I was able to see consecutive erroneous messages when ACC was braking heavily so I'm thinking of using an updated message for the subaru issue specifically anyway. (routes noted in the OP PR) However I'm trying to see this through as I believe it would still be a good safety improvement. 

Panda is now working and ready to be out of draft but the matching OP code is not ready and may be more difficult.

ClockeNessMnstr:
See comment in OP PR. or don't. 

tl;dr: debouncing is good. but this would be a bigger refactor in OP than 54d0709 was in panda.

ClockeNessMnstr:
Is the brake signal debounced for Honda?
it appears to be debounced in OP/selfdrive but the brake signal here looks like it is passed through raw which would give a controls mismatch on the single frame brake noise for Honda

ClockeNessMnstr:
This can just be done per vehicle line. it looks like Honda had done something similar at one point but has since abandoned it.

PR #648, opened 2021-05-17
Red H7

briskspirit:


board/SConscript
```diff
@@ -85,13 +99,17 @@ def objcopy(source, target, env, for_signature):
...
+else:
+  linkerscript_fn = File("stm32_flash.ld").srcnode().abspath
 
 flags = [
-  "-Wall",
+  #"-Wall",
   "-Wextra",
   "-Wstrict-prototypes",
   "-Werror",
+  "-Wno-unused-variable", # Remove in prod
```
robbederks, line 112: Added a comment here to make sure this gets reverted before merging!

board/board.h
```diff
@@ -7,39 +7,53 @@
...
-  #include "drivers/rtc.h"
-  #include "drivers/clock_source.h"
-  #include "boards/white.h"
-  #include "boards/grey.h"
-  #include "boards/black.h"
-  #include "boards/uno.h"
-  #include "boards/dos.h"
+  #ifdef STM32H7
+    #include "boards/red.h"
+#include "drivers/rtc.h"
```
robbederks, line 12: Indentation

board/board.h
```diff
@@ -7,39 +7,53 @@
...
+    #include "boards/dos.h"
+  #endif
 #else
   #include "boards/pedal.h"
 #endif
 
 void detect_board_type(void) {
   #ifdef PANDA
+    #ifdef STM32H7
+      if(!detect_with_pull(GPIOF, 7, PULL_UP) && !detect_with_pull(GPIOF, 8, PULL_UP) && !detect_with_pull(GPIOF, 9, PULL_UP) && !detect_with_pull(GPIOF, 10, PULL_UP)){
```
robbederks, line 30: For the H7 line with the new board detection pins, let's already abstract this in a function to read the ID from those pins, and return it as a uint8

board/config.h
```diff
@@ -7,11 +7,17 @@
 //#define DEBUG_SPI
 //#define DEBUG_FAULTS
 
-#ifdef STM32F4
+#ifdef STM32H7
+  #define PANDA
+  #include "stm32h7xx.h"
+  #define FREQ 550U // 550Mhz
```
robbederks, line 13: Let's rename this to `CORE_FREQ` or something, might be easier to understand when you come across it. Also, `Hz` not `hz` ;)

board/drivers/clock.h
```diff
@@ -1,30 +1,135 @@
...
+    register_set_bits(&(RCC->CFGR), RCC_CFGR_SW_PLL);
+    while ((RCC->CFGR & RCC_CFGR_SWS) != RCC_CFGR_SWS_PLL);
+
+    // *** running on PLL ***
+  #endif
 }
 
+#ifdef STM32H7
 void watchdog_init(void) {
+
```
robbederks, line 117: Don't think the regular pandas enable the watchdog. This is only on pedals, so for now I think it'd be fine to not implement this

PR #649, opened 2021-05-18
Honda Bosch: add alt brake msg to rx checks

adeebshihadeh:


gregjhogan:
Does this break cars that do not use the alt brake message?

Apparently that was why I at one point tried to add this, then reverted it:
https://github.com/commaai/panda/pull/268#issuecomment-627738900

And this does really need to get added at some point so we can enable this test:
https://github.com/commaai/panda/blob/f96514621435928a7f843172d78db0115d50c813/tests/safety/test_honda.py#L287-L295

adeebshihadeh:
No, this won't break cars without the brake check. We still need to add something to be able to check a signal conditionally, since the alt brake cars still use 0x17C for gas.

PR #651, opened 2021-05-25
Subaru Global: use better brake pressed signal

martinl:
Use alternative Brake_Status Brake signal instead of Brake_Pedal Brake_Pedal signal. Fixes ACC disengage while braking issue for various newer models (https://github.com/martinl/openpilot/issues/27, https://github.com/commaai/openpilot/pull/20866)

Openpilot PR https://github.com/commaai/openpilot/pull/21012

### DIFF
--- a/board/safety/safety_subaru.h
+++ b/board/safety/safety_subaru.h
@@ -88,7 +88,7 @@ static int subaru_rx_hook(CAN_FIFOMailBox_TypeDef *to_push) {
     }
 
     if (addr == 0x13c) {
-      brake_pressed = ((GET_BYTES_48(to_push) >> 30) & 1);
+      brake_pressed = ((GET_BYTE(to_push, 7) >> 6) & 1);
     }
 
     if (addr == 0x40) {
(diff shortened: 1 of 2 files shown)


martinl:
I did some additional research and found that Brake_Status Brake signal is set to 1 when Brake_Pedal signal goes above 48. 

Screenshot 2021-05-26 at 12 13 24

What do you think, would increasing the brake_pressed threshold to Brake_Pedal > 16 be a better solution than using the Brake_Status Brake signal?

ClockeNessMnstr:
I believe I had brake_pedal = 32 on a false positive as my minimum signal was 16 in our 2019 forester

board/safety/safety_subaru.h
```diff
@@ -87,8 +87,8 @@ static int subaru_rx_hook(CAN_FIFOMailBox_TypeDef *to_push) {
       vehicle_moving = subaru_speed > SUBARU_STANDSTILL_THRSLD;
     }
 
-    if (addr == 0x139) {
-      brake_pressed = (GET_BYTES_48(to_push) & 0xFFF0) > 0;
+    if (addr == 0x13c) {
+      brake_pressed = ((GET_BYTES_48(to_push) >> 30) & 1);
```
adeebshihadeh, line 91:
```suggestion
      brake_pressed = ((GET_BYTE(to_push, 7) >> 6) & 1);
```

PR #653, opened 2021-05-31
Improve type hinting & minor cleanup

cmyui:
Pretty much just added some missing type hints or fixed incorrect ones, along with minor optimizations.

### DIFF
--- a/python/uds.py
+++ b/python/uds.py
@@ -314,7 +314,7 @@ class CanClient:
             rx_data = bytes(rx_data)  # convert bytearray to bytes
 
             if self.debug:
-              print(f"CAN-RX: {hex(rx_addr)} - 0x{bytes.hex(rx_data)}")
+              print(f"CAN-RX: {hex(rx_addr)} - 0x{rx_data.hex()}")
 
             # Cut off sub addr in first byte
             if self.sub_addr is not None:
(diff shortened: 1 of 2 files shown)


python/uds.py
```diff
@@ -502,8 +503,8 @@ def get_rx_addr_for_tx_addr(tx_addr, rx_offset=0x8):
   raise ValueError("invalid tx_addr: {}".format(tx_addr))
 
 
-class UdsClient():
-  def __init__(self, panda, tx_addr: int, rx_addr: int = None, bus: int = 0, timeout: float = 1, debug: bool = False):
+class UdsClient:
+  def __init__(self, panda, tx_addr: int, rx_addr: Optional[int] = None, bus: int = 0, timeout: float = 1.0, debug: bool = False):
```
cmyui, line 507: I ran into one problem during this pr, basically this rx_addr may be None, if so it will call `get_rx_addr_for_tx_addr(tx_addr)`, but if `tx_addr` is in `FUNCTIONAL_ADDRS` then it will return None while CanClient requires an int for rx_addr.. So I suppose it's either gotta throw an exception somewhere or `CanClient.rx_addr` has to be `Optional[int]`.

I don't really have enough info about what this stuff actually does/is to answer this one.

robbederks: When using the functional address to send (like a broadcast), it's not certain which address the ECU(s) will respond on. See `_recv_filter` for how those are handled now, not sure about the best way to clean this up

PR #654, opened 2021-06-05
add flag to health inidicate heartbeat loss

adeebshihadeh:


python/__init__.py
```diff
@@ -353,6 +353,7 @@ def health(self):
       "safety_param": a[15],
       "fault_status": a[16],
       "power_save_enabled": a[17]
```
adeebshihadeh, line 355:
```suggestion
      "power_save_enabled": a[17],
```

PR #656, opened 2021-06-09
Refactor HAL

briskspirit:
Refactoring MCU specific stuff. 

P.S.: Getting strange misra error that I am not getting locally upon checking.


board/bootmode.h
```diff
@@ -20,7 +20,7 @@ void jump_to_bootloader(void) {
   NVIC_SystemReset();
 }
 
-void early(void) {
+void bootmode(void) {
```
robbederks, line 23: I agree early wasn't a great name, but 'bootmode' also doesn't say much imo

board/bootstub.c
```diff
@@ -72,11 +68,12 @@ extern void *_app_start[];
 // BOUNTY: $200 coupon on shop.comma.ai or $100 check.
 
 int main(void) {
+  clock_init();
```
robbederks, line 71: Sure it's fine to change the order of this? Any reason? We had some issues with the pedal code a few weeks ago which were due to having interrupts enabled while this runs or something iirc.
If not needed, I'd like to keep this unchanged for now

board/drivers/interrupts.h
```diff
@@ -53,9 +52,8 @@ void init_interrupts(bool check_rate_limit){
   }
 
   // Init timer 10 for a 1s interval
-  register_set_bits(&(RCC->APB1ENR), RCC_APB1ENR_TIM6EN);  // Enable interrupt timer peripheral
```
robbederks, line 56: Where/why did this go?

board/drivers/rtc.h
```diff
@@ -78,11 +78,11 @@ void rtc_set_time(timestamp_t time){
 
 timestamp_t rtc_get_time(void){
     timestamp_t result;
-    // Init with zero values in case there is no RTC running
-    result.year = 0U;
-    result.month = 0U;
-    result.day = 0U;
-    result.weekday = 0U;
+    // Init with correct starting values in case there is no RTC running(zeros will break python datetime)
```
robbederks, line 81: does it? if so, shouldn't we handle this in the python lib instead?

board/faults.h
```diff
@@ -24,7 +24,7 @@
 #define FAULT_REGISTER_DIVERGENT            (1U << 18)
 #define FAULT_INTERRUPT_RATE_KLINE_INIT     (1U << 19)
 #define FAULT_INTERRUPT_RATE_CLOCK_SOURCE   (1U << 20)
-#define FAULT_INTERRUPT_RATE_TIM9           (1U << 21)
+#define FAULT_INTERRUPT_RATE_HEARTBEAT      (1U << 21)
```
robbederks, line 27: Don't agree with the name "heartbeat", we usually use this to refer to the message openpilot sends to panda periodically such that the panda knows it's still "alive".
I suggest something more like "tick" or something, not 100% sure what it should be called

board/libc.h
```diff
@@ -5,6 +5,15 @@ void delay(uint32_t a) {
...
 
+void delay_us(uint32_t a) {
+  uint32_t end = MICROSECOND_TIMER->CNT + a;
+  while(MICROSECOND_TIMER->CNT < end);
+}
+
+void delay_ms(uint32_t a) {
+  delay_us(a*1000U);
+}
+
```
robbederks, line 16: This looks like a great place for edge cases, and I'm pretty sure this doesn't work.
Example: counter is 50 away from overflowing. a is set to 100, such that 'end' is already overflown. The while loop will exit immediately.
This also relies on the fact that nobody will set the auto reset value of the timer and that it's a 32 bits one.

<s>Also: do we have enough timers that we can assign one to this?</s> wasn't aware we already had this timer

board/main.c
```diff
@@ -685,8 +690,8 @@ void __attribute__ ((noinline)) enable_fpu(void) {
 
 // called at 8Hz
 uint8_t loop_counter = 0U;
-void TIM1_BRK_TIM9_IRQ_Handler(void) {
-  if (TIM9->SR != 0) {
+void heartbeat_handler(void) {
```
robbederks, line 693: see naming comment above

board/main.c
```diff
@@ -778,23 +783,22 @@ void TIM1_BRK_TIM9_IRQ_Handler(void) {
...
     loop_counter %= 8U;
   }
-  TIM9->SR = 0;
+  HEARTBEAT_TIMER->SR = 0;
 }
 
-#define MAX_FADE 8192U
+
 int main(void) {
+  clock_init();
```
robbederks, line 791: again, why the move? this needs to be tested well

board/main.c
```diff
@@ -843,10 +847,10 @@ int main(void) {
...
   // increments 1000000 times per second
   // generate an update to set the prescaler
-  TIM2->PSC = 48-1;
-  TIM2->CR1 = TIM_CR1_CEN;
-  TIM2->EGR = TIM_EGR_UG;
-  // use TIM2->CNT to read
+  MICROSECOND_TIMER->PSC = (APB1_FREQ)-1U;
+  MICROSECOND_TIMER->CR1 = TIM_CR1_CEN;
+  MICROSECOND_TIMER->EGR = TIM_EGR_UG;
+  // use MICROSECOND_TIMER->CNT to read
```
robbederks, line 853: oh, wasn't aware we already had a microsecond timer, my bad

board/pedal/main.c
```diff
@@ -280,21 +281,20 @@ void pedal(void) {
 }
 
 int main(void) {
+  // init devices
+  clock_init();
```
robbederks, line 285: again, hahah
also, I think we need to disable interrupts first

board/stm32f4_flash.ld
```diff
@@ -15,7 +15,7 @@
 **
 **  Environment : Atollic TrueSTUDIO(R)
 **
-**  Distribution: The file is distributed �as is,� without any warranty
+**  Distribution: The file is distributed �as is,� without any warranty
```
robbederks, line 18: ?

tests/safety/test.c
```diff
@@ -36,6 +36,7 @@ TIM_TypeDef *TIM2 = &timer;
 #define HW_TYPE_BLACK_PANDA 3U
 #define HW_TYPE_PEDAL 4U
 #define HW_TYPE_UNO 5U
+#define HW_TYPE_DOS 6U
```
robbederks, line 39: good catch!

robbederks:
Let's also fix the misra issue, maybe it's a version thing?

board/main.c
```diff
@@ -3,6 +3,11 @@
 
 // ********************* Includes *********************
 #include "config.h"
+#ifdef STM32F4
```
adeebshihadeh, line 6: I'd put this ifdef in its own header then include that header instead of duplicating this everywhere

board/power_saving.h
```diff
@@ -1,4 +1,5 @@
 // WARNING: To stay in compliance with the SIL2 rules laid out in STM UM1840, we should never implement any of the available hardware low power modes.
+// Except while offroad (!)
```
adeebshihadeh, line 2: not relevant to the PR

board/stm32f4_config.h
```diff
@@ -0,0 +1,34 @@
+#define PANDA
```
adeebshihadeh, line 1: should we make another sub directory for these like we do with boards? then we can move the startup and linker scripts there too

board/stm32f4_config.h
```diff
@@ -0,0 +1,34 @@
...
+
+#define HEARTBEAT_TIMER_IRQ TIM1_BRK_TIM9_IRQn
+#define HEARTBEAT_TIMER TIM9
+
+#define MICROSECOND_TIMER TIM2
+
+#define PROVISION_CHUNK_ADDRESS 0x1FFF79E0
+#define SERIAL_NUMBER_ADDRESS 0x1FFF79C0
+
+void early_gpio_float(void) {
```
adeebshihadeh, line 28: a lot of this is shared with the f2, should we move some of it to a common header?

board/stm32f4_flash.ld
briskspirit: Wow, very strange

board/main.c
briskspirit: Already tested, but will do more and will get more info. But always had a rule to init clock first and peripherals - second. (and interrupts are off from the start, but understand that precaution)

board/faults.h
briskspirit: I understand, has this in my notes not to forget ;)

board/drivers/rtc.h
briskspirit: Does, tested. Python crashes on datetime. 
My opinion that it should be handled in both places. 
If panda responds to rtc request - it should be the right one. And lib also should be protected (also noted to add this check to lib)

board/drivers/interrupts.h
briskspirit: It is a duplicate. This is already enabled in peripherals_init. Main peripheral clocks should be enabled only in one place imho. (If they are not tied to specific driver that is not used in every panda generation)

board/bootstub.c
briskspirit: Think I will keep it as is for now and will change later after checks and more reading, but think clocks should always go first(power, core clock, buses etc)

board/bootmode.h
briskspirit: Mmm, ideas? Because as I understood the purpose of that function is to check boot flag and select how to boot - normal mode, flasher or DFU. I don't see something special there about hardware that must be initialized before main.c

tests/safety/test.c
briskspirit: Also need to add to python lib, just a note for myself

board/main.c
briskspirit: Noted, maybe even will add this ifdef to config.h, will check, thx

board/power_saving.h
briskspirit: Discussed with George, added as a note for the future, also will add TODO comment there about power sleep modes.

board/stm32f4_config.h
briskspirit: Agree, will do

board/power_saving.h
briskspirit: Will remove

board/bootmode.h
robbederks: what do you mean? it does a lot of stuff:
- disable interrupts
- initialize the register driver
- decide the boot more
- check that it's running on the right chip
- setup gpio and the interrupt table
- check which board we're running on

let's name it 'early_initialization' or something, no?

board/bootstub.c
robbederks: I;m really reluctant to changing this without proper testing (should be a separate PR too imo), so let's try to keep all the functional stuff the same in this PR. This should purely be a refactor without any functional changes

board/drivers/interrupts.h
robbederks: I see, sounds good then

board/drivers/rtc.h
robbederks: this seems more like a bug in python tbh, I'd rather have the panda respond with all 0s if it's not valid

board/main.c
robbederks: are you sure interrupts are off from the start? something tells me that they might be enabled if you reset out of DFU

board/drivers/rtc.h
briskspirit: Python's datetime lib will complain if values are wrong. But ok, won't argue on that, but still would like to reply with correct format )

board/bootmode.h
briskspirit: So why do we initialize peripherals, clocks and etc again in main.c?

briskspirit: Ok, will name it as 'early_initialization'.

board/drivers/rtc.h
robbederks: well, don't we want it to complain? seems more like a feature than a bug :)

adeebshihadeh: agree, think this should be handled in the python lib

adeebshihadeh: comment is out of date now

adeebshihadeh:
This PR seems to only abstract out the set of things needed to support the H7. I think we want to fully abstract away the hardware from the main panda logic. For example, instead of abstracting out [which timer is used here](https://github.com/commaai/panda/blob/refactor_hal/board/main.c#L817), abstract the whole timer init into a function. This shouldn't make the PR too much longer, and it's a refactor we'd want regardless of the H7.

briskspirit:
Moving everything hardware related to stm32fx (and later will add stm32h7). Even board.h was moved as it has hardware specific includes and board detect method(that was changed for H7 future boards)
usb.h was split into usb.h and llusb.h 
Also would like to rename llgpio.h to gpio.h
common.h was renamed to peripherals.h and moved to HAL dir, don't think that it belongs to boards now

Question: what future is of kline and gmlan for next gen future boards?

board/boards/board_functions.h
```diff
@@ -0,0 +1,25 @@
+// ///// Board functions ///// //
+// TODO: Make these config options in the board struct
```
robbederks, line 2: Let's not put this in a separate file, but rather add these as boolean fields in the board struct, makes way more sense imo

board/drivers/llgpio.h
```diff
@@ -63,3 +63,13 @@ int get_gpio_input(GPIO_TypeDef *GPIO, unsigned int pin) {
...
+// Detection with internal pullup
+#define PULL_EFFECTIVE_DELAY 4096
+bool detect_with_pull(GPIO_TypeDef *GPIO, int pin, int mode) {
+  set_gpio_mode(GPIO, pin, MODE_INPUT);
+  set_gpio_pullup(GPIO, pin, mode);
+  for (volatile int i=0; i<PULL_EFFECTIVE_DELAY; i++);
+  bool ret = get_gpio_input(GPIO, pin);
+  set_gpio_pullup(GPIO, pin, PULL_NONE);
+  return ret;
+}
```
robbederks, line 75: Let's make sure we have newlines at the end of every file. There are vscode plugins to fix these kinds of things if you're using that

robbederks:
> Moving everything hardware related to stm32fx (and later will add stm32h7). Even board.h was moved as it has hardware specific includes and board detect method(that was changed for H7 future boards)
> usb.h was split into usb.h and llusb.h
> Also would like to rename llgpio.h to gpio.h
> common.h was renamed to peripherals.h and moved to HAL dir, don't think that it belongs to boards now
> 
> Question: what future is of kline and gmlan for next gen future boards?

Looks good to me. GMLAN bitbanging is something we'll still have to maintain, since we also have support for this with the OBDII harness.
Kline is only present on the white/grey pandas, don't think we'll ever implement that again

board/boards/board_functions.h
robbederks: `board_has_obd` and `board_has_relay` is kinda already covered by the `.has_harness` in the harness config. 
I think we can also add `.has_rtc` , `.has_lin` and `.has_gps` in the `board` struct.
Maybe the `has_gmlan` should be something like `.has_onboard_gmlan`?

board/drivers/llgpio.h
briskspirit: Found it in the settings finally, will fix in all files.

board/boards/board_functions.h
briskspirit: Kept .has_obd (as we discussed on Slack), removed .has_relay as it was not even used, everything else moved to struct.

board/boards/white.h
```diff
@@ -265,6 +265,11 @@ const harness_configuration white_harness_config = {
 const board board_white = {
   .board_type = "White",
   .harness_config = &white_harness_config,
+  .has_gps = false,
+  .has_onboard_gmlan = true,
```
adeebshihadeh, line 269:
```suggestion
  .has_hw_gmlan = true,
```

board/bootstub.c
```diff
@@ -74,10 +68,10 @@ extern void *_app_start[];
 int main(void) {
   // Init interrupt table
   init_interrupts(true);
-
+
```
adeebshihadeh, line 71: be careful with these unintended changes

board/drivers/llgpio.h
adeebshihadeh: disagree with this if we're not going to enforce this with a linter in CI. there's a bunch of files in this PR that only change the newline at the end of the file which clutters up the diff

board/safety/safety_volkswagen.h
```diff
@@ -244,7 +244,7 @@ static int volkswagen_pq_rx_hook(CAN_FIFOMailBox_TypeDef *to_push) {
 
 static bool volkswagen_steering_check(int desired_torque) {
   bool violation = false;
-  uint32_t ts = TIM2->CNT;
+  uint32_t ts = MICROSECOND_TIMER->CNT;
```
adeebshihadeh, line 247: we shouldn't be reading registers like this in safety code

adeebshihadeh: maybe for another PR

board/pedal/main.c
```diff
@@ -8,17 +8,18 @@
...
+#include "stm32fx/llcan.h"
+#include "drivers/gpio.h"
+#include "stm32fx/lladc.h"
 
-#include "board.h"
+#include "boards/board_declarations.h"
+#include "stm32fx/board.h"
 
-#include "drivers/clock.h"
+#include "stm32fx/clock.h"
```
adeebshihadeh, line 19: include config header instead?

board/main.c
```diff
@@ -816,20 +813,18 @@ int main(void) {
   // init microsecond system timer
   // increments 1000000 times per second
   // generate an update to set the prescaler
-  TIM2->PSC = 48-1;
-  TIM2->CR1 = TIM_CR1_CEN;
-  TIM2->EGR = TIM_EGR_UG;
-  // use TIM2->CNT to read
+  microsecond_timer_init();
```
adeebshihadeh, line 816: comments are out of date now

board/stm32fx/board.h
```diff
@@ -1,8 +1,8 @@
 // ///////////////////////////////////////////////////////////// //
 // Hardware abstraction layer for all different supported boards //
 // ///////////////////////////////////////////////////////////// //
-#include "board_declarations.h"
-#include "boards/common.h"
+#include "stm32fx/peripherals.h"
+#include "../boards/board_declarations.h"
```
adeebshihadeh, line 5: fix this in cppcheck. just add the include path

PR #658, opened 2021-06-10
update cppcheck + fix new misra violations

adeebshihadeh:


adeebshihadeh:
@briskspirit this PR is getting pretty big. let's break this up into smaller chunks.

board/drivers/can.h
```diff
@@ -55,6 +61,7 @@ can_buffer(tx1_q, 0x100)
 can_buffer(tx2_q, 0x100)
 can_buffer(tx3_q, 0x100)
 can_buffer(txgmlan_q, 0x100)
+// cppcheck-suppress misra-c2012-9.3
```
robbederks, line 64: Probably fixed by doing `can_ring *can_queues[4] = {&can_tx1_q, &can_tx2_q, &can_tx3_q, &can_txgmlan_q};`

board/drivers/can.h
```diff
@@ -42,10 +44,14 @@ uint32_t ignition_can_cnt = 0U;
...
-int can_live = 0, pending_can_live = 0, can_loopback = 0, can_silent = ALL_CAN_SILENT;
+int can_live = 0;
+int pending_can_live = 0;
+int can_loopback = 0;
+int can_silent = ALL_CAN_SILENT;
 
 // ********************* instantiate queues *********************
-
+// cppcheck-suppress misra-c2012-20.10
+// cppcheck-suppress misra-c2012-20.7
```
robbederks, line 54: kinda agree with this rule, let's add parentheses

board/drivers/can.h
```diff
@@ -42,10 +44,14 @@ uint32_t ignition_can_cnt = 0U;
...
 
-int can_live = 0, pending_can_live = 0, can_loopback = 0, can_silent = ALL_CAN_SILENT;
+int can_live = 0;
+int pending_can_live = 0;
+int can_loopback = 0;
+int can_silent = ALL_CAN_SILENT;
 
 // ********************* instantiate queues *********************
-
+// cppcheck-suppress misra-c2012-20.10
```
robbederks, line 53: this one is stupid imo, do we want to suppress it here, or add it to the global suppressions?

board/drivers/can.h
```diff
@@ -65,7 +72,7 @@ int can_err_cnt = 0;
 int can_overflow_cnt = 0;
 
 // ********************* interrupt safe queue *********************
-
+// cppcheck-suppress misra-c2012-8.2
```
robbederks, line 75: to fix all these, we could declare the prototypes somewhere on top in this file

board/drivers/interrupts.h
```diff
@@ -14,7 +14,7 @@ void unused_interrupt_handler(void) {
 
 #define NUM_INTERRUPTS 102U                // There are 102 external interrupt sources (see stm32f413.h)
 interrupt interrupts[NUM_INTERRUPTS];
-
+// cppcheck-suppress misra-c2012-20.7
```
robbederks, line 17: let's fix instead of suppress all of the 20.7s

board/drivers/can.h
briskspirit: Nope, that's what I checked first

board/drivers/can.h
briskspirit: Prototypes are already declared. It stops complaining if you will remove prototypes, but then you need to shuffle some funcs. Adeeb told to do that as a second PR as it looks massive.

board/drivers/interrupts.h
briskspirit: They are suppressed till next PR as this one is pretty big already.

board/drivers/can.h
briskspirit: I would add it to the globals as it makes no sense imho. "The # and ## preprocessor operators should not be used"

adeebshihadeh: this one is advisory, but I agree with the rationale and think we should enforce it

board/drivers/interrupts.h
adeebshihadeh: the diff from 20.7 should be pretty small, let's do it in this PR

board/drivers/can.h
yaddd7: Adeeb mother tucker

PR #662, opened 2021-06-18
Extra checks for Honda Bosch Longitudinal

pd0wm:


board/safety/safety_honda.h
```diff
@@ -290,6 +299,13 @@ static int honda_tx_hook(CAN_FIFOMailBox_TypeDef *to_send) {
     }
   }
 
+  // Only tester present ("\x02\x3E\x80\x00\x00\x00\x00\x00") allowed on diagnostics address
+  if (addr == 0x18DAB0F1) {
+    tx &= (GET_BYTE((to_send), 0) == 0x02);
+    tx &= (GET_BYTE((to_send), 1) == 0x3E);
+    tx &= (GET_BYTE((to_send), 2) == 0x80);
```
adeebshihadeh, line 306: check the rest of the bytes?
```suggestion
  if ((GET_BYTES_04(to_send) != 0x023E8000) || (GET_BYTES_48(to_send) != 0x0)) {
    tx = 0;
  }
```

pd0wm: We already check the length is 2, so the exact padding shouldn’t matter.

adeebshihadeh: hmm, the addr checks have the length as 8
https://github.com/commaai/panda/blob/12fccafc47109fc9e514f680dfe722a12ebe6823/board/safety/safety_honda.h#L13

pd0wm: The message is iso-tp, so the first byte is the length (2 bytes). The other 5 bytes are just padding, so the contents are ignored by the receiver (some senders pad with `\x00`, some with `\xaa` I think),

pd0wm: But if you're more comfortable constraining the whole message that's fine with me too!

adeebshihadeh: yeah, think it's best to make the safety code as unambiguous as possible

PR #663, opened 2021-06-18
Update cppcheck and misra, suppress for now

briskspirit:
Updated cppcheck and misra to the latest, suppressed inline new violations, will fix with next PRs.

### DIFF
--- a/board/crc.h
+++ b/board/crc.h
@@ -1,6 +1,8 @@
 uint8_t crc_checksum(uint8_t *dat, int len, const uint8_t poly) {
+  // FIXME:
   // cppcheck-suppress misra-c2012-7.2
   uint8_t crc = 0xFF;
+  // FIXME:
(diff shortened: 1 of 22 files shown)


adeebshihadeh (commented):
LGTM, but let's prefix the temp suppressions with `FIXME: `

board/crc.h
```diff
@@ -1,5 +1,9 @@
 uint8_t crc_checksum(uint8_t *dat, int len, const uint8_t poly) {
+  // FIXME:
+  // cppcheck-suppress misra-c2012-7.2
```
adeebshihadeh, line 3: meant like this
```suggestion
  // FIXME: cppcheck-suppress misra-c2012-7.2
```

briskspirit: It won't work! Won't suppress

PR #669, opened 2021-06-19
Fix MISRA 8.2 violations

briskspirit:
uart.h has very small change, so decided to combine both in one PR

### DIFF
--- a/board/drivers/can.h
+++ b/board/drivers/can.h
@@ -27,14 +27,6 @@ extern int can_loopback;
 extern int can_silent;
 extern uint32_t can_speed[4];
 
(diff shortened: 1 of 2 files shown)


board/drivers/can.h
```diff
@@ -72,8 +64,6 @@ int can_err_cnt = 0;
 int can_overflow_cnt = 0;
 
 // ********************* interrupt safe queue *********************
-// FIXME:
-// cppcheck-suppress misra-c2012-8.2
```
adeebshihadeh, line 76: why is it flagging this as a violation of 8.2? false positive?

PR #671, opened 2021-06-19
Remove non-EON panda build

adeebshihadeh:


pd0wm:
Should we make the python library disable the heartbeat as soon as you set a safety mode (with an optional param to not do that)?

adeebshihadeh:
I was thinking something more explicit. I added a USB command that's only built with the debug build which disables the heartbeat checks. Sending a heartbeat will also reenable the checks if disabled.

pd0wm:
I think the python library should always send that command when you set a safety mode, think it will be too confusing/hard to use otherwise.

PR #672, opened 2021-06-20
Verify flash succeeded

mogorman:
I had a few bad panda flashes the other day, not sure what caused it
thought it might be nice to have the panda verify the flash after flashing.


adeebshihadeh:
Let's add an argument to Panda.flash to do this

pd0wm:
Can you also set the exit code of the bash script to a non zero value if verify fails? That might be useful for other scripts calling this.

Issue #599, opened 2020-09-22
If EON hardware failure - alert driver using car chime.


dzid26:
Correct me if I am wrong, but I think currently if EON dies quietly, panda only [switches](https://github.com/commaai/panda/blob/5e07142174e583dbdc8238315fd2f76cd1651b0d/board/main.c#L718)  to `SAFETY_SILENT` after 2s of missing heartbeat without extra safety consideration, such as alarming the driver using car chime.

Below are two basic scenarios that come in my mind (described using pseudo iso26262 nomenclature hazard assessment.) 

**Failure mode**: EON dies quietly (e.g. due to CPU, PSU, faulty battery) causing steering requests to be halted, then: 
| id | Scenario|  Effect | Hazard | Exposure | Controlability | Severity |
| --- | --- | --- | --- |  --- |  --- |  --- |
| 1 | straght highway | Car continues straight | Car slowly crosses lane  | E4 | C1 | S2 |
| 2 | curved country road | Steering centers rapidly* | Car crosses onto oncoming traffic or sidewalk |  E3 | C2 | S3 |

*Depending on a car, the effect and controllability may be less severe. 

The probability of failure itself is hard to guess. _(It did happen to me before that EON died on me silently, but frankly it was mainly my development fault.)_
Luckily, I think these hazards can be largely mitigated regardless. My proposal is to: 
- Play loud chime on car speakers using CAN request when the EON heartbeat is not detected for 1s. 
(Perhaps for 2nd scenario, 1s may be too long - depending on how and how quickly a particular car steering responds.)

A framework for that could be to have `safety_comm_loss_hook` specific for each car. (Besides a chime, additionally canceling of cruise control should be commanded for some cars.)

robbederks:
Most cars will practically immediately fault due to the lack of LKAS messages (which is why we only switch back to silent after 5 seconds, that re-connects the stock system).
We have another feature planned to take care of this issue in the future: we want to always keep a stream of CAN messages queued up on the panda which can be sent in the event of a loss of openpilot, which would contain safe actuator commands for the next few seconds and alerts.
Also, see #594  for spoilers

PR #674, opened 2021-06-29
Enable siren when heartbeat is lost

pd0wm:


dzid26:
Nice, that should solve this https://github.com/commaai/panda/issues/599

PR #675, opened 2021-06-30
Fix MISRA 8.2 violations

briskspirit:
Update cppcheck with the latest upstream and fix legit 8.2 violations

### DIFF
--- a/Dockerfile.panda
+++ b/Dockerfile.panda
@@ -50,7 +50,7 @@ RUN pyenv install 3.8.5 && \
     pip install --no-cache-dir -r /tmp/requirements.txt
 
 RUN cd /tmp && \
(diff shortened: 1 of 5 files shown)


Dockerfile.panda
```diff
@@ -53,7 +53,7 @@ RUN cd /tmp && \
     git clone https://github.com/commaai/cppcheck.git && \
     cd cppcheck && \
     git fetch && \
-    git checkout b440734ba7bda34c9d4e3cd2d9c26dc2303db542 && \
+    git checkout e1cff1d1ef92f6a1c6962e0e4153b7353ccad04c && \
```
adeebshihadeh, line 56: we can point to the cppcheck remote again

adeebshihadeh (approved):

PR #682, opened 2021-07-08
HAL refactoring part 2

briskspirit:
WIP!


board/drivers/timers.h
```diff
@@ -5,14 +5,18 @@ void timer_init(TIM_TypeDef *TIM, int psc) {
   TIM->SR = 0;
 }
 
-void microsecond_timer_init(void) {
-  MICROSECOND_TIMER->PSC = (APB1_FREQ)-1U;
-  MICROSECOND_TIMER->CR1 = TIM_CR1_CEN;
-  MICROSECOND_TIMER->EGR = TIM_EGR_UG;
+void us_timer_init(void) {
```
adeebshihadeh, line 8: should we make a timer struct instead?

board/boards/black.h
```diff
@@ -213,9 +213,6 @@ void black_init(void) {
   if (car_harness_status == HARNESS_STATUS_FLIPPED) {
     can_flip_buses(0, 2);
   }
-
-  // init multiplexer
-  can_set_obd(car_harness_status, false);
```
adeebshihadeh, line 218: why was this removed?

board/stm32fx/stm32fx_config.h
```diff
@@ -49,17 +56,25 @@
...
-#include "drivers/timers.h"
 #include "drivers/gpio.h"
+#include "stm32fx/peripherals.h"
+#include "stm32fx/interrupt_handlers.h"
+#include "drivers/timers.h"
+#include "stm32fx/lladc.h"
+#include "stm32fx/board.h"
+#include "stm32fx/clock.h"
+
+#if (defined(PANDA) && !defined (BOOTSTUB)) || (defined(PEDAL_USB) && !defined (BOOTSTUB))
```
adeebshihadeh, line 67:
```suggestion
#if defined(BOOTSTUB) || (defined(PANDA) && defined (PEDAL_USB))
```

board/stm32fx/stm32fx_config.h
```diff
@@ -20,12 +24,15 @@
...
+// Threshold voltage (mV) for either of the SBUs to be below before deciding harness is connected
+#define HARNESS_CONNECTED_THRESHOLD 2500U
+
 #define NUM_INTERRUPTS 102U                // There are 102 external interrupt sources (see stm32f413.h)
 
 #define TICK_TIMER_IRQ TIM1_BRK_TIM9_IRQn
 #define TICK_TIMER TIM9
 
-#define MICROSECOND_TIMER TIM2
+#define US_TIMER TIM2
```
adeebshihadeh, line 35: think microsecond timer was a better name

board/stm32fx/llflash.h
```diff
@@ -0,0 +1,23 @@
...
+    FLASH->CR = (sector << 3) | FLASH_CR_SER;
+    FLASH->CR |= FLASH_CR_STRT;
+    while (FLASH->SR & FLASH_SR_BSY);
+  }
+}
+
+void flash_write_byte(void *prog_ptr, const void *data) {
+  uint32_t *pp = prog_ptr;
+  const uint32_t *d = data;
+    FLASH->CR = FLASH_CR_PSIZE_1 | FLASH_CR_PG;
```
adeebshihadeh, line 20: indentation

board/stm32fx/llflash.h
```diff
@@ -0,0 +1,23 @@
...
+void flash_erase_sector(uint8_t sector, uint8_t unlocked) {
+  // don't erase the bootloader(sector 0)
+  if (sector != 0 && sector < 12 && unlocked) {
+    FLASH->CR = (sector << 3) | FLASH_CR_SER;
+    FLASH->CR |= FLASH_CR_STRT;
+    while (FLASH->SR & FLASH_SR_BSY);
+  }
+}
+
+void flash_write_byte(void *prog_ptr, const void *data) {
```
adeebshihadeh, line 17: more than a byte

board/flasher.h
```diff
@@ -26,32 +26,24 @@ int usb_cb_control_msg(USB_Setup_TypeDef *setup, uint8_t *resp, bool hardwired)
...
-        FLASH->KEYR = 0x45670123;
-        FLASH->KEYR = 0xCDEF89AB;
-        resp[1] = 0xff;
-      }
+      flash_unlock();
+      resp[1] = 0xff;
       current_board->set_led(LED_GREEN, 1);
       unlocked = 1;
-      prog_ptr = (uint32_t *)0x8004000;
+      prog_ptr = (uint32_t *)APP_START;
```
adeebshihadeh, line 33: APP_START_ADDRESS?

board/flasher.h
```diff
@@ -26,32 +26,24 @@ int usb_cb_control_msg(USB_Setup_TypeDef *setup, uint8_t *resp, bool hardwired)
...
+      flash_erase_sector(sec, unlocked);
+      resp[1] = 0xff;
       break;
     // **** 0xd0: fetch serial number
     case 0xd0:
       #ifdef STM32F4
         // addresses are OTP
         if (setup->b.wValue.w == 1) {
-          memcpy(resp, (void *)0x1fff79c0, 0x10);
+          memcpy(resp, (void *)SERIAL_NUMBER_ADDRESS, 0x10);
```
adeebshihadeh, line 46: DEVICE_SERIAL_NUMBER_ADDRESS

board/flasher.h
```diff
@@ -26,32 +26,24 @@ int usb_cb_control_msg(USB_Setup_TypeDef *setup, uint8_t *resp, bool hardwired)
       break;
     // **** 0xb1: unlock flash
     case 0xb1:
-      if (FLASH->CR & FLASH_CR_LOCK) {
-        FLASH->KEYR = 0x45670123;
-        FLASH->KEYR = 0xCDEF89AB;
-        resp[1] = 0xff;
-      }
+      flash_unlock();
+      resp[1] = 0xff;
```
adeebshihadeh, line 30: check result of flash unlock like was done previously

board/flasher.h
```diff
@@ -1,6 +1,6 @@
 // flasher state variables
 uint32_t *prog_ptr = NULL;
-int unlocked = 0;
+uint8_t unlocked = 0;
```
adeebshihadeh, line 3: bool?

board/drivers/timers.h
briskspirit: ok, will do but in the next separate PR as there is too many places in the safety code

board/stm32fx/stm32fx_config.h
briskspirit: #if !defined (BOOTSTUB) && (defined(PANDA) || defined(PEDAL_USB))

adeebshihadeh: oops, that's what I meant

board/boards/dos.h
```diff
@@ -204,9 +204,6 @@ void dos_init(void) {
     can_flip_buses(0, 2);
   }
 
-  // init multiplexer
-  can_set_obd(car_harness_status, false);
```
adeebshihadeh, line 208: this should go in a separate PR next time

board/flasher.h
```diff
@@ -26,32 +26,25 @@ int usb_cb_control_msg(USB_Setup_TypeDef *setup, uint8_t *resp, bool hardwired)
...
       sec = setup->b.wValue.w;
-      // don't erase the bootloader
-      if (sec != 0 && sec < 12 && unlocked) {
-        FLASH->CR = (sec << 3) | FLASH_CR_SER;
-        FLASH->CR |= FLASH_CR_STRT;
-        while (FLASH->SR & FLASH_SR_BSY);
-        resp[1] = 0xff;
-      }
+      flash_erase_sector(sec, unlocked);
+      resp[1] = 0xff;
```
adeebshihadeh, line 40: resp is always set now

board/stm32fx/stm32fx_config.h
```diff
@@ -2,11 +2,15 @@
   #include "stm32fx/inc/stm32f4xx.h"
   #include "stm32fx/inc/stm32f4xx_hal_gpio_ex.h"
   #define MCU_IDCODE 0x463U
+  #define BCDDEVICE 0x23
```
adeebshihadeh, line 5: generally prefer typed constants instead of defines, i.e. `const uint8_t` in this case

board/drivers/gpio.h
```diff
@@ -59,6 +64,15 @@ void set_gpio_pullup(GPIO_TypeDef *GPIO, unsigned int pin, unsigned int mode) {
   EXIT_CRITICAL();
 }
 
+void set_gpio_speed(GPIO_TypeDef *GPIO, unsigned int pin, unsigned int speed) {
```
adeebshihadeh, line 67: why was this added?

briskspirit: setting GPIO speed for USB

adeebshihadeh:
Should we move the low level drivers into their own folder like stm32fx/drivers? I think it would also be nice to make proper headers for all the low level drivers, but maybe for a separate PR.

board/drivers/can.h
```diff
@@ -235,34 +235,6 @@ void can_set_gmlan(uint8_t bus) {
   }
 }
 
-// TODO: remove
-void can_set_obd(uint8_t harness_orientation, bool obd){
```
robbederks, line 239: Where did this go?

board/stm32fx/llrtc.h
```diff
@@ -0,0 +1,57 @@
+#define RCC_BDCR_OPTIONS (RCC_BDCR_RTCEN | RCC_BDCR_RTCSEL_0 | RCC_BDCR_LSEON)
+#define RCC_BDCR_MASK (RCC_BDCR_RTCEN | RCC_BDCR_RTCSEL | RCC_BDCR_LSEMOD | RCC_BDCR_LSEBYP | RCC_BDCR_LSEON)
+
+void rtc_init(void){
```
robbederks, line 4: How different is this on the H7?

board/drivers/gpio.h
```diff
@@ -10,6 +10,11 @@
 #define OUTPUT_TYPE_PUSH_PULL 0U
 #define OUTPUT_TYPE_OPEN_DRAIN 1U
 
+#define SPEED_LOW 0U
```
robbederks, line 13: maybe change to `GPIO_SPEED_x` for clarity since these will be globally in scope.  Are these speed grades (`LOW`, `MEDIUM`, etc) the same name as in the datasheet?

board/stm32fx/peripherals.h
```diff
@@ -1,3 +1,17 @@
+void gpio_usb_init(void) {
+  // A11,A12: USB
+  set_gpio_alternate(GPIOA, 11, GPIO_AF10_OTG_FS);
+  set_gpio_alternate(GPIOA, 12, GPIO_AF10_OTG_FS);
+  set_gpio_speed(GPIOA, 11, SPEED_VERY_HIGH);
```
robbederks, line 5: why do we set the speed now on the F4? we never did this, correct?

robbederks: if we don't need to set this, let's not

board/stm32fx/stm32fx_config.h
```diff
@@ -20,6 +24,9 @@
 
 #define MAX_LED_FADE 8192U
 
+// Threshold voltage (mV) for either of the SBUs to be below before deciding harness is connected
+#define HARNESS_CONNECTED_THRESHOLD 2500U
```
robbederks, line 28: Why is this processor specific now? Doesn't look like something that should be!

robbederks:
This is a huge PR with multiple distinct changes. Let's split all distinct changes into separate PRs. The diff is very hard to read and review if it's this big. For example, this diff could have been split up in:

- Remove can_set_obd (which I'm not convinced we can do tbh, are we sure this doesn't break anything??)
- Add speed setting on USB GPIO pins (another change I'm not convinced we need or want)
- Move hardware specific driver code into HAL (interrupt definitions, RTC, UART, bcdDevice)
- Refactor USB init
- Refactor flasher
- Refactor microsecond timer in safety

I'm not confident in reviewing and approving all these 1000+ changed lines in one PR

board/drivers/can.h
briskspirit: It was replaced by set_can_mode , isn't it?

board/drivers/gpio.h
briskspirit: Speed grades are the same and naming was taken from HAL lib. If you want to add GPIO_ in front - I would add it to all defines. What do you think?

board/stm32fx/peripherals.h
briskspirit: it was set before: https://github.com/commaai/panda/blob/3815bf8f41931f957c3cdeeb9cf628109368b110/board/spi_flasher.h#L313

board/stm32fx/stm32fx_config.h
briskspirit: I will take a look why I needed this but think it was because of 12 bit vs 16 bit ADC and getting a lot different values from H7 ADC (and conversion formula is used for adc_get_voltage for one divider only.

board/stm32fx/llrtc.h
briskspirit: At minimum some macros naming is different for H7 and also few isn't available

board/stm32fx/lluart.h
```diff
@@ -0,0 +1,230 @@
...
+    NVIC_EnableIRQ(UART5_IRQn);
+  } else {
+    // UART not used. Skip enabling interrupts
+  }
+
+  // Initialise RX DMA if used
+  if(q->dma_rx){
+    dma_rx_init(q);
+  }
+}
```
robbederks, line 230: There is quite a bit of logic in here which I can't believe will be very (if at all) different when implementing on the H7 (stuff like the ring handlers).

Wouldn't it be better to keep this common, but abstract away the specific register differences if there are any? Would like to avoid having the same code twice because they added a bit in a register somewhere...

board/stm32fx/llrtc.h
robbederks: Which ones are you talking about specifically? maybe let's also keep a common version of this with only the (tiny?) differences abstracted out

board/stm32fx/lluart.h
adeebshihadeh: I think we should fully abstract away the hardware from the main panda code then we can abstract the common STMXX stuff on top of that. I think only abstracting away the differences for H7 would leave the code in a weird semi-abstracted state.

PR #686, opened 2021-07-13
Refactor flasher

briskspirit:
Refactor flasher.h (previously spi_flasher.h)

### DIFF
--- a/board/flasher.h
+++ b/board/flasher.h
@@ -117,7 +117,7 @@ void usb_cb_ep2_out(void *usbdata, int len, bool hardwired) {
   UNUSED(hardwired);
   current_board->set_led(LED_RED, 0);
   for (int i = 0; i < len/4; i++) {
-    flash_write_word(prog_ptr, (uint32_t*)(usbdata+(i*4)));
+    flash_write_word(prog_ptr, *(uint32_t*)(usbdata+(i*4)));
 
     //*(uint64_t*)(&spi_tx_buf[0x30+(i*4)]) = *prog_ptr;
     prog_ptr++;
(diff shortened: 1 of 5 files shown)


board/flasher.h
```diff
@@ -26,23 +26,17 @@ int usb_cb_control_msg(USB_Setup_TypeDef *setup, uint8_t *resp, bool hardwired)
       break;
     // **** 0xb1: unlock flash
     case 0xb1:
-      if (FLASH->CR & FLASH_CR_LOCK) {
-        FLASH->KEYR = 0x45670123;
-        FLASH->KEYR = 0xCDEF89AB;
+      if (flash_unlock()) {
```
robbederks, line 29: maybe let's split this in two functions instead? `flash_is_locked()` and `flash_unlock()`?

board/stm32fx/llflash.h
```diff
@@ -0,0 +1,27 @@
...
+  if (sector != 0 && sector < 12 && unlocked) {
+    FLASH->CR = (sector << 3) | FLASH_CR_SER;
+    FLASH->CR |= FLASH_CR_STRT;
+    while (FLASH->SR & FLASH_SR_BSY);
+    return true;
+  }
+  return false;
+}
+
+void flash_write_word(void *prog_ptr, const void *data) {
```
robbederks, line 21: why pass in the data as a pointer? if you're always writing 32 bits, let's just pass the data in as a uint32_t directly

board/stm32fx/stm32fx_config.h
```diff
@@ -69,6 +69,8 @@
 
 #ifndef BOOTSTUB
   #include "stm32fx/llcan.h"
+#else
```
robbederks, line 72: maybe let's swap this around to `#ifdef BOOTSTUB`? seems easier to reason about

PR #689, opened 2021-07-13
Deprecate variable bcdDevice in usb.h

briskspirit:
Can't be done as a const while keeping pretty as it will violate MISRA rule 20.1


robbederks:
As far as I can tell, this different BCD is only used to detect if it's a "legacy" panda in the python libs (leftover from the very old revision of the white panda). I'd be fine with completely deprecating this and setting it do a constant value instead.
DIfferentiating boards should be done by `HW_TYPE` anyways.

robbederks:
You should have also removed the logic from the python libraries and the tests. Now this will be broken!

briskspirit:
> You should have also removed the logic from the python libraries and the tests. Now this will be broken!

Will be fixed with #692

PR #690, opened 2021-07-13
Abstract MICROSECOND_TIMER

briskspirit:
Abstract MICROSECOND_TIMER


tests/safety/test.c
```diff
@@ -28,6 +28,7 @@ struct sample_t torque_driver;
 
 TIM_TypeDef timer;
 TIM_TypeDef *MICROSECOND_TIMER = &timer;
+#define microsecond_timer_get() MICROSECOND_TIMER->CNT
```
robbederks, line 31: Any reason you used a define instead of just also having the function here?

PR #691, opened 2021-07-14
Move dac.h to HAL

briskspirit:
LL, used by pedal only


board/stm32fx/stm32fx_config.h
```diff
@@ -75,6 +75,10 @@
   #include "stm32fx/llusb.h"
 #endif
 
+#ifdef PEDAL
```
robbederks, line 78: I'd be fine with not gating this on PEDAL if not needed

PR #697, opened 2021-07-21
Prevent Steering Temporarily Unavailable Fault

JasonJShuler:
This modification fixes the "Steering Temporarily Unavailable" error thrown on certain GM vehicles with a "picky" PSCM. When the error is thrown, the vehicle must be completely powered down to reset.
The error was first noticed on the Chevy Bolt EV, but has been found to occur on several other models.

This fix should be compatible with all GM vehicles, including those previously supported.

There were two situations that would cause a fault:
1. Engaging (typically after having disengaged, but also occasionally on initial engagement)
2. Seemingly at random during continuous use

The root cause is two-fold.
1. These PSCMs are intolerant of out-of-order rolling counter values
2. These PSCMs do not properly handle LKAS messages sent faster than 20ms apart

The existing Panda code (based on the Chevy Volt PSCM) does not ensure LKAS messages are sent with sequential rolling counters. When engaging and disengaging, or when the Panda code blocks a frame, gaps in the sequence occur. The newer steering columns do not permit this and hard-fault. Additionally Openpilot's LKAS messages are occasionally slightly delayed. We initially thought it was the delay causing the fault, when it is actually the shortened duration of the subsequent message.

Most likely the PSCM is using a single "last value received" buffer for the value received via CAN, then polling this value every 20ms for the torque-application code. When messages arrive faster than 20ms the previous value is overwritten before it can be "read," thus creating the appearance of a missed rolling counter value. (It is important to note that the stock LKAS hardware send messages every 20ms while active, and occasionally triggers this)

This solution:
Every time an LKAS message is going to be transmitted, we check that the rolling counter value is the correct next sequential value, and that the time since the last LKAS message is greater than 20ms. If these conditions are satisfied, we store the timestamp and rolling counter value and allow the transmission. If either condition is not satisfied, we simply block the transmission of the frame. After a maximum of 3 dropped frames the incoming rolling counter will be re-aligned, resulting in a maximum added lag of 60ms. In practice this timeframe is imperceptible.

(Note: the previous method for solving this issue was based on the incorrect assumption that it was the delayed messages causing the fault, rather than the truncated timespan after a delayed message. Turns out the complex timer code was not necessary at all. The PSCM is quite tolerant of lower frequency messages, just nothing faster than 20ms)

### DIFF
--- a/board/safety/safety_gm.h
+++ b/board/safety/safety_gm.h
@@ -18,6 +18,7 @@ const int GM_DRIVER_TORQUE_FACTOR = 4;
 const int GM_MAX_GAS = 3072;
 const int GM_MAX_REGEN = 1404;
 const int GM_MAX_BRAKE = 350;
+const uint32_t GM_LKAS_MIN_INTERVAL = 20000;    // 20ms minimum between LKAS frames
 const CanMsg GM_TX_MSGS[] = {{384, 0, 4}, {1033, 0, 7}, {1034, 0, 7}, {715, 0, 8}, {880, 0, 6},  // pt bus
                              {161, 1, 7}, {774, 1, 8}, {776, 1, 7}, {784, 1, 2},   // obs bus
                              {789, 2, 5},  // ch bus
@@ -33,6 +34,10 @@ AddrCheckStruct gm_rx_checks[] = {
   {.msg = {{417, 0, 7, .expected_timestep = 100000U}, { 0 }, { 0 }}},
 };
 const int GM_RX_CHECK_LEN = sizeof(gm_rx_checks) / sizeof(gm_rx_checks[0]);
+
+//LKAS vars for ensuring in-order, correct timing
+int gm_lkas_last_rc = -1;
+uint32_t gm_lkas_last_ts = 0;
 
 static int gm_rx_hook(CAN_FIFOMailBox_TypeDef *to_push) {
 
@@ -143,6 +148,7 @@ static int gm_tx_hook(CAN_FIFOMailBox_TypeDef *to_send) {
 
   // LKA STEER: safety check
   if (addr == 384) {
+    int rolling_counter = GET_BYTE(to_send, 0) >> 4;
     int desired_torque = ((GET_BYTE(to_send, 0) & 0x7U) << 8) + GET_BYTE(to_send, 1);
     uint32_t ts = microsecond_timer_get();
     bool violation = 0;
@@ -187,6 +193,23 @@ static int gm_tx_hook(CAN_FIFOMailBox_TypeDef *to_send) {
     if (violation) {
       tx = 0;
     }
+
+    //TODO: Maybe should be checked at the moment the frame is sent via CAN - rcv interrupt could maybe prevent sending??
+    if (tx == 1) {
+      uint32_t lkas_elapsed = get_ts_elapsed(ts, gm_lkas_last_ts);
+      int expected_lkas_rc = (gm_lkas_last_rc + 1) % 4;
+      //If less than 20ms have passed since last LKAS message or the rolling counter value isn't correct it is a violation
+      //TODO: The interval may need some fine tuning - testing the tolerance of the PSCM / send lag
+      if (lkas_elapsed < GM_LKAS_MIN_INTERVAL || rolling_counter != expected_lkas_rc) {
+        tx = 0;
+      }
+      else {
+        //otherwise, save values
+        gm_lkas_last_rc = rolling_counter;
+        gm_lkas_last_ts = ts;
+      }
+    }
+
   }
 
   // GAS/REGEN: safety check


board/safety/safety_gm.h
```diff
@@ -34,6 +35,10 @@ AddrCheckStruct gm_rx_checks[] = {
 };
 const int GM_RX_CHECK_LEN = sizeof(gm_rx_checks) / sizeof(gm_rx_checks[0]);
 
+//LKAS vars for ensuring in-order, correct timing
+int gm_lkas_last_rc = -1;
```
JasonJShuler, line 39: The -1 value ensures the first Rolling Counter value will be 0

board/safety/safety_gm.h
```diff
@@ -187,6 +193,22 @@ static int gm_tx_hook(CAN_FIFOMailBox_TypeDef *to_send) {
     if (violation) {
       tx = 0;
     }
+
+    //Check that approved LKAS message has correct rolling counter value and isn't arriving too soon
+    if (tx == 1) {
```
JasonJShuler, line 198: Using tx rather than violation just in case tx would ever be 0 without a violation. Since there are other, earlier operations setting the tx flag that may use different logic from the violation flag, this seems necessary.

jyoung8607:
This connects to commaai/openpilot#21557 and also the very old commaai/openpilot#476 (mainly listing that because of the involved end user's findings while troubleshooting). I spent a rather long time working with the end user of the more recent bug, and I broadly agree with how @JasonJShuler describes the problem.

Unfortunately, openpilot isn't that great at smooth isochronous message transmission, and I couldn't make it much better despite trying some fairly radical approaches at load reduction. The problem is helped, but not cured, by having openpilot send slower (at 33Hz instead of 50Hz, for example) but even then, things just seem to pile-up occasionally.

We can't reasonably fix the EPS rack behavior; it is what it is. The STM32 is our best bet at realtime policing or shaping. Unless and until such time as Panda takes over message generation directly, this PR is about the cleanest way I can think to mitigate the problem.

board/safety/safety_gm.h
```diff
@@ -187,6 +193,22 @@ static int gm_tx_hook(CAN_FIFOMailBox_TypeDef *to_send) {
...
+      uint32_t lkas_elapsed = get_ts_elapsed(ts, gm_lkas_last_ts);
+      int expected_lkas_rc = (gm_lkas_last_rc + 1) % 4;
+      //If less than 20ms have passed since last LKAS message or the rolling counter value isn't correct it is a violation
+      if (lkas_elapsed < GM_LKAS_MIN_INTERVAL || rolling_counter != expected_lkas_rc) {
```
jyoung8607, line 202: misra-c2012-12.1

```suggestion
      if ((lkas_elapsed < GM_LKAS_MIN_INTERVAL) || (rolling_counter != expected_lkas_rc)) {
```

adeebshihadeh:
These changes don't contribute to enforcing the safety. Why can't this be in openpilot?

jyoung8607:
> These changes don't contribute to enforcing the safety. Why can't this be in openpilot?

This wasn't exactly Plan A. openpilot doesn't seem to be able to keep timing closely enough to satisfy the EPS, even if we slow down the send rate or take other drastic load-reduction measures.

Speaking for myself, I understand this is sort of at the wrong abstraction layer, and I'm open to fixing it in openpilot if there exists a way to do so, but I've not found a way. Maybe table this until comma_con and chat about it then?

JasonJShuler:
One could argue that the unexpected, surprise hard-faulting of lateral control is a safety hazard.

The proposed changes are enforcing "protocol" requirements of the EPS/PSCM, similar to how the Panda double-checks torque limitations and accel/decel rates - essentially it is imperative that LKAS frames _never under any circumstances_ be sent less than 20ms after the previous LKAS frame. This is not something the OP code is (presently) capable of guaranteeing - as far as OP itself is concerned it _is_ sending them at 50hz. Any OP-side workaround should still be double-checked on the Panda same as the other double-checks.

Slowing the LKAS send rate in OP would result in OP operating different than the stock system and could risk future failures. Even if we were to slow the LKAS send rate, I would advise keeping this check in place - without it, it is possible for OP to glitch and kill lateral control.

JasonJShuler:
> These changes don't contribute to enforcing the safety. Why can't this be in openpilot?

One could argue that the unexpected, surprise hard-faulting of lateral control is a safety hazard.

The proposed changes are enforcing "protocol" requirements of the EPS/PSCM, similar to how the Panda double-checks torque limitations and accel/decel rates - essentially it is imperative that LKAS frames _never under any circumstances_ be sent less than 20ms after the previous LKAS frame. This is not something the OP code is (presently) capable of guaranteeing - as far as OP itself is concerned it _is_ sending them at 50hz. Any OP-side workaround should still be double-checked on the Panda same as the other double-checks.

Slowing the LKAS send rate in OP would result in OP operating different than the stock system and could risk future failures. Even if we were to slow the LKAS send rate, I would advise keeping this check in place - without it, it is possible for OP to glitch and kill lateral control.

adeebshihadeh:
> One could argue that the unexpected, surprise hard-faulting of lateral control is a safety hazard.

There is a concrete definition of safety in this context. I recommend reviewing the [openpilot safety model](https://github.com/commaai/openpilot/blob/master/SAFETY.md). Faulting is acceptable under this safety model. 
 
> The proposed changes are enforcing "protocol" requirements of the EPS/PSCM, similar to how the Panda double-checks torque limitations and accel/decel rates - essentially it is imperative that LKAS frames _never under any circumstances_ be sent less than 20ms after the previous LKAS frame. This is not something the OP code is (presently) capable of guaranteeing - as far as OP itself is concerned it _is_ sending them at 50hz. Any OP-side workaround should still be double-checked on the Panda same as the other double-checks.

I understand what this is doing, but this isn't the same as the other checks. There are a couple of key differences here. The existing checks enforce the safety through actuation limits. Another key difference here is that it's never expected for the messages to be blocked under normal operation, with the exception of braking messages during an AEB event in the Honda Nidec safety mode.

JasonJShuler:
Ultimately at this point the Panda seems to be the only place this issue can be corrected reliably. It is my opinion that we are working around buggy EPS code - it should not be ignoring any messages - but it does, we can't get GM to fix it, and the timing constraints cannot be reliably resolved on the OP side.

I got the impression in the comments on my previous port attempt that Geohot's primary concern was the Panda not generating its own messages. It was a while back - things may have changed, I may have misunderstood - and apologies if that is the case.

There are other things that can only be done on the panda that I was planning to issue PRs for.
 * In order to use the relay, we must monitor the rolling counter(s) from the stock camera and perform this same process to align with OP's rolling counter.
* In order to passthrough AEB, stock ACC, and prevent "Service Front Camera" warnings or DTCs forwarding must be implemented.
* If we are able to use AEB for longitudinal braking in vehicles without ACC, once again we will need to keep the rolling counters aligned. We will also need to add the messages to the database, and support them in the Pedal code
* Code to handle the use of the Pedal interceptor is required on both the OP and Panda sides (this will be required on Bolts pre-2022)

Not all of these things are 100% and only safety related, but there is no alternative to Panda code modification (at present), and using the relay at all will require this "alignment" code.

If this is unacceptable, I'm open to suggestions.

Thanks again for your time!

JasonJShuler:
I updated the code to match the existing violation patterns and mitigated my concerns over running an expensive modulo operation using a power of 2 shortcut (bitwise and vs division).

Since this flags frames as violations the same way other checks do, hopefully there is reduced concern over novel behavior.

tunabelly:
@JasonJShuler thanks so much for taking the time to make this. This will fix a lot of issues with the GM LKAS saying it's unavailable when a LKAS message comes slightly too soon. The GM system is super sensitive, and this will finally fix it!

Thanks again, you rock!

Verylukyguy:
Yes, thank you for solving this issue!  I've been using this fix for about 5,000 miles and have had no errors or faults.  As far as I can tell, this issue has been around since 12/18 and many people have been struggling with it.

You are the best.

adeebshihadeh:
I'm going to close this for the reasons mentioned above. We'll keep tracking this in the [openpilot issue](https://github.com/commaai/openpilot/issues/21557).

JasonJShuler:
> I'm going to close this for the reasons mentioned above. We'll keep tracking this in the [openpilot issue](https://github.com/commaai/openpilot/issues/21557).

I'm confused - I addressed your concerns in comments and code modification.
Is this some sort of personal issue?

jyoung8607:
This has nothing to do with personal feelings about anybody in this PR, it's purely a technical architecture decision about the right place to make the fix. I talked to @adeebshihadeh about this issue P2P at comma_con. Apologies for not updating the PR since then.

@Verylukyguy has been driving very successfully on a 25Hz message rate with a suitably altered ramp-rate and tune for >2000 miles, which means (as you well know, in fact, as you established) the EPS will tolerate that slower rate, which means we should probably use that in order to reduce our exposure to the oversensitive GM EPS. On top of that, he was still running your Panda modification. At the time I just wanted to make sure his vacation trip was successful.

We need to test with him to see how his car does without the modification. I think it may not do as well, but I also think there's a path forward if it doesn't. I had observed some issues with cumLagMs incrementing on his device, and if that's happening, that's something that needs to be addressed within core openpilot code. That won't necessarily be part of the GM car port.

JasonJShuler:
Thanks for the update - I included an explanation in the issue for why I still consider this necessary. The timing glitches are not the only thing that causes faults - essentially because the Panda blocks frames independent of OP and faster than OP can react, I can't see a way for OP to prevent gaps. This code (or something similar) is also necessary to ensure the transition from Camera to OP doesn't involve gaps or squashed timing.

The alternative I _really_ don't want to pursue (but will as a last resort to get these cars upstreamed) is to put this code on a separate microcontroller included in the wiring harness. This seems a little absurd to have to essentially have a duplicate panda enforcing timing...

PR #698, opened 2021-07-23
VW MQB: Longitudinal safety

jyoung8607:
Work in progress, but runs and drives with full tests. It's broadly similar to the HKG long control scheme. The openpilot side has a ways to go before it's ready for a WIP draft PR.

There's no provision to pass-through a UDS diagnostic disable command, because at this time I've not found one that works on our radar, at least not without security access and switching diagnostic sessions. I haven't given up on finding a way, in fact I'm confident I can find something eventually, but it may be awhile. In the meantime, it works with gateway integrations, and when/if a way is found, safety should otherwise work the same for camera integrations.

While here, I discovered and fixed a bug where MQB stock long was re-entering controls whenever stock ACC was active, not just on the rising edge. Both the bug and the tests have been fixed. I can separate that into its own PR if desired.

### DIFF
--- a/board/safety/safety_volkswagen.h
+++ b/board/safety/safety_volkswagen.h
@@ -57,7 +57,7 @@ const int VOLKSWAGEN_PQ_RX_CHECKS_LEN = sizeof(volkswagen_pq_rx_checks) / sizeof
 
 const uint16_t VOLKSWAGEN_PARAM_LONG = 1;
 
-int volkswagen_longitudinal = false;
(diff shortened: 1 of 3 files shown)


jyoung8607:
Closing this one for now. The approach is sound, but it now needs resolution of merge conflicts with #707 and other updates, plus safety checks on ACC_07 (in addition to ACC_06), plus permission to transmit a couple other messages.

PR #681, opened 2021-07-08
UDS fingerprinting development tool updates

jyoung8607:
**Bug fixes:**

* Take Panda out of powersave mode so the OBD bus CAN transceiver actually works

**New features:**

* Support querying a single address instead of the entire address range
* Switch to best available diagnostic session without security access
* Support queries for all standard UDS data identifiers
* Support queries for nonstandard vehicle manufacturer/supplier reserved identifier ranges
* Support German Engineered™ ISO-TP receive offsets

**Known issues:**

* Timeout isn't long enough for certain *really* long ISO-TP replies, which ends up throwing an assert deep in the UDS library when the reply for a previous query is still arriving when we start the next. Timeout would have to be made absurdly longer, or the UDS library assert needs to become a handle-able exception, or the timeout timer needs to be reset every time we receive a new ISO-TP reply fragment. Probably a combination of the last two.

**Example output, all addresses, standard UDS queries only:**

[std_query_0x6a.txt](https://github.com/commaai/panda/files/6786485/std_query_0x6a.txt)

**Example output, single address, all queries:**

```
root@localhost:/data/openpilot/panda/examples$ ./query_fw_versions.py --nonstandard --addr=0x712 --rxoffset=0x6a

*** Results for address 0x712 ***


0xF15B IDENTIFICATION_OPTION_VEHICLE_MANUFACTURER_SPECIFIC_DATA_IDENTIFIER: b'\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00'
0xF17B IDENTIFICATION_OPTION_VEHICLE_MANUFACTURER_SPECIFIC_DATA_IDENTIFIER: b'!\x06\x01'
0xF17C IDENTIFICATION_OPTION_VEHICLE_MANUFACTURER_SPECIFIC_DATA_IDENTIFIER: b'ZF0-00008.01.1802100308'
0xF182 APPLICATION_DATA_IDENTIFICATION: b'\x0571A0JA15A1'
0xF186 ACTIVE_DIAGNOSTIC_SESSION: b'\x03'
0xF187 VEHICLE_MANUFACTURER_SPARE_PART_NUMBER: b'3Q0909144L '
[...]

PR #699, opened 2021-07-24
Timeouts in fw versions query example scan

martinl:
Fixes following traceback (caused by one specific infotainment ecu), allowing the fw query scan to finish and show results
```
./panda/examples/query_fw_versions.py
opening device 37003a001651383432363133 0xddcc
connected
querying addresses ...
0x7d0:  41%|██████████████████████████████████████████████████████▊                                                                                | 208/512 [00:47<01:10,  4.34it/s]
  File "./panda/examples/query_fw_versions.py", line 61, in <module>
    data = uds_client.read_data_by_identifier(uds_data_id)  # type: ignore
  File "/data/pythonpath/panda/python/uds.py", line 654, in read_data_by_identifier
    resp = self._uds_request(SERVICE_TYPE.READ_DATA_BY_IDENTIFIER, subfunction=None, data=data)
  File "/data/pythonpath/panda/python/uds.py", line 526, in _uds_request
    resp = isotp_msg.recv()
  File "/data/pythonpath/panda/python/uds.py", line 401, in recv
    self._isotp_rx_next(msg)
  File "/data/pythonpath/panda/python/uds.py", line 443, in _isotp_rx_next
    assert self.rx_idx & 0xF == rx_data[0] & 0xF, "isotp - rx: invalid consecutive frame index"
AssertionError: isotp - rx: invalid consecutive frame index
```

### DIFF
--- a/python/uds.py
+++ b/python/uds.py
@@ -361,7 +361,6 @@ class IsoTpMessage():
     self.debug = debug
     self.max_len = max_len
     self.response_pending_timeout = response_pending_timeout
-    self.response_pending = False
 
   def send(self, dat: bytes) -> None:
     # throw away any stale data
@@ -385,32 +384,30 @@ class IsoTpMessage():
     if self.tx_len < self.max_len:
       # single frame (send all bytes)
       if self.debug:
-        print(f"ISO-TP: TX - {hex(self._can_client.tx_addr)} single frame")
+        print(f"ISO-TP: TX - single frame - {hex(self._can_client.tx_addr)}")
       msg = (bytes([self.tx_len]) + self.tx_dat).ljust(self.max_len, b"\x00")
       self.tx_done = True
     else:
       # first frame (send first 6 bytes)
       if self.debug:
-        print(f"ISO-TP: TX - {hex(self._can_client.tx_addr)} first frame")
+        print(f"ISO-TP: TX - first frame - {hex(self._can_client.tx_addr)}")
       msg = (struct.pack("!H", 0x1000 | self.tx_len) + self.tx_dat[:self.max_len - 2]).ljust(self.max_len - 2, b"\x00")
     self._can_client.send([msg])
 
-  def recv(self) -> Optional[bytes]:
+  def recv(self, response_pending: bool = False) -> Optional[bytes]:
     start_time = time.time()
-    self.rx_idx_prev = 0
     try:
       while True:
         for msg in self._can_client.recv():
           self._isotp_rx_next(msg)
+          start_time = time.time()
+          response_pending = False
           if self.tx_done and self.rx_done:
             return self.rx_dat
-          if self.rx_idx > self.rx_idx_prev:
-            start_time = time.time()
-            self.rx_idx_prev = self.rx_idx
         # no timeout indicates non-blocking
         if self.timeout == 0:
           return None
-        if self.response_pending:
+        if response_pending:
           if time.time() - start_time > self.response_pending_timeout:
             raise MessageTimeoutError("timeout waiting for pending response")
         elif time.time() - start_time > self.timeout:
@@ -439,7 +436,7 @@ class IsoTpMessage():
       if self.debug:
         print(f"ISO-TP: RX - first frame - {hex(self._can_client.rx_addr)} idx={self.rx_idx} done={self.rx_done}")
       if self.debug:
-        print(f"ISO-TP: TX - {hex(self._can_client.tx_addr)} flow control continue")
+        print(f"ISO-TP: TX - flow control continue - {hex(self._can_client.tx_addr)}")
       # send flow control message (send all bytes)
       msg = b"\x30\x00\x00".ljust(self.max_len, b"\x00")
       self._can_client.send([msg])
@@ -465,7 +462,7 @@ class IsoTpMessage():
       assert rx_data[0] == 0x30 or rx_data[0] == 0x31, "isotp - rx: flow-control transfer state indicator invalid"
       if rx_data[0] == 0x30:
         if self.debug:
-          print(f"ISO-TP: RX - {hex(self._can_client.rx_addr)} flow control continue")
+          print(f"ISO-TP: RX - flow control continue - {hex(self._can_client.tx_addr)}")
         delay_ts = rx_data[2] & 0x7F
         # scale is 1 milliseconds if first bit == 0, 100 micro seconds if first bit == 1
         delay_div = 1000. if rx_data[2] & 0x80 == 0 else 10000.
@@ -491,7 +488,7 @@ class IsoTpMessage():
       elif rx_data[0] == 0x31:
         # wait (do nothing until next flow control message)
         if self.debug:
-          print(f"ISO-TP: TX - {hex(self._can_client.tx_addr)} flow control wait")
+          print(f"ISO-TP: TX - flow control wait - {hex(self._can_client.tx_addr)}")
 
 FUNCTIONAL_ADDRS = [0x7DF, 0x18DB33F1]
 
@@ -512,13 +509,14 @@ def get_rx_addr_for_tx_addr(tx_addr, rx_offset=0x8):
 
 
 class UdsClient():
-  def __init__(self, panda, tx_addr: int, rx_addr: int = None, bus: int = 0, timeout: float = 1, debug: bool = False):
+  def __init__(self, panda, tx_addr: int, rx_addr: int = None, bus: int = 0, timeout: float = 1, debug: bool = False, response_pending_timeout: float = 5):
     self.bus = bus
     self.tx_addr = tx_addr
     self.rx_addr = rx_addr if rx_addr is not None else get_rx_addr_for_tx_addr(tx_addr)
     self.timeout = timeout
     self.debug = debug
     self._can_client = CanClient(panda.can_send, panda.can_recv, self.tx_addr, self.rx_addr, self.bus, debug=self.debug)
+    self.response_pending_timeout = response_pending_timeout
 
   # generic uds request
   def _uds_request(self, service_type: SERVICE_TYPE, subfunction: int = None, data: bytes = None) -> bytes:
@@ -528,11 +526,13 @@ class UdsClient():
     if data is not None:
       req += data
 
+
     # send request, wait for response
-    isotp_msg = IsoTpMessage(self._can_client, self.timeout, self.debug)
+    isotp_msg = IsoTpMessage(self._can_client, self.timeout, self.debug, response_pending_timeout = self.response_pending_timeout)
     isotp_msg.send(req)
+    response_pending = False
     while True:
-      resp = isotp_msg.recv()
+      resp = isotp_msg.recv(response_pending)
 
       if resp is None:
         continue
@@ -553,7 +553,7 @@ class UdsClient():
           error_desc = resp[3:].hex()
         # wait for another message if response pending
         if error_code == 0x78:
-          isotp_msg.response_pending = True
+          response_pending = True
           if self.debug:
             print("UDS-RX: response pending")
           continue
--- a/python/uds.py
+++ b/python/uds.py
@@ -355,11 +355,13 @@ class CanClient():
         self._recv_buffer()
 
 class IsoTpMessage():
-  def __init__(self, can_client: CanClient, timeout: float = 1, debug: bool = False, max_len: int = 8):
+  def __init__(self, can_client: CanClient, timeout: float = 1, debug: bool = False, max_len: int = 8, response_pending_timeout: float = 5):
     self._can_client = can_client
     self.timeout = timeout
     self.debug = debug
     self.max_len = max_len
+    self.response_pending_timeout = response_pending_timeout
+    self.response_pending = False
 
   def send(self, dat: bytes) -> None:
     # throw away any stale data
@@ -376,39 +378,46 @@ class IsoTpMessage():
     self.rx_done = False
 
     if self.debug:
-      print(f"ISO-TP: REQUEST - 0x{bytes.hex(self.tx_dat)}")
+      print(f"ISO-TP: REQUEST - {hex(self._can_client.tx_addr)} 0x{bytes.hex(self.tx_dat)}")
     self._tx_first_frame()
 
   def _tx_first_frame(self) -> None:
     if self.tx_len < self.max_len:
       # single frame (send all bytes)
       if self.debug:
-        print("ISO-TP: TX - single frame")
+        print(f"ISO-TP: TX - {hex(self._can_client.tx_addr)} single frame")
       msg = (bytes([self.tx_len]) + self.tx_dat).ljust(self.max_len, b"\x00")
       self.tx_done = True
     else:
       # first frame (send first 6 bytes)
       if self.debug:
-        print("ISO-TP: TX - first frame")
+        print(f"ISO-TP: TX - {hex(self._can_client.tx_addr)} first frame")
       msg = (struct.pack("!H", 0x1000 | self.tx_len) + self.tx_dat[:self.max_len - 2]).ljust(self.max_len - 2, b"\x00")
     self._can_client.send([msg])
 
   def recv(self) -> Optional[bytes]:
     start_time = time.time()
+    self.rx_idx_prev = 0
     try:
       while True:
         for msg in self._can_client.recv():
           self._isotp_rx_next(msg)
           if self.tx_done and self.rx_done:
             return self.rx_dat
+          if self.rx_idx > self.rx_idx_prev:
+            start_time = time.time()
+            self.rx_idx_prev = self.rx_idx
         # no timeout indicates non-blocking
         if self.timeout == 0:
           return None
-        if time.time() - start_time > self.timeout:
+        if self.response_pending:
+          if time.time() - start_time > self.response_pending_timeout:
+            raise MessageTimeoutError("timeout waiting for pending response")
+        elif time.time() - start_time > self.timeout:
           raise MessageTimeoutError("timeout waiting for response")
     finally:
       if self.debug and self.rx_dat:
-        print(f"ISO-TP: RESPONSE - 0x{bytes.hex(self.rx_dat)}")
+        print(f"ISO-TP: RESPONSE - {hex(self._can_client.rx_addr)} 0x{bytes.hex(self.rx_dat)}")
 
   def _isotp_rx_next(self, rx_data: bytes) -> None:
     # single rx_frame
@@ -418,7 +427,7 @@ class IsoTpMessage():
       self.rx_idx = 0
       self.rx_done = True
       if self.debug:
-        print(f"ISO-TP: RX - single frame - idx={self.rx_idx} done={self.rx_done}")
+        print(f"ISO-TP: RX - single frame - {hex(self._can_client.rx_addr)} idx={self.rx_idx} done={self.rx_done}")
       return
 
     # first rx_frame
@@ -428,9 +437,9 @@ class IsoTpMessage():
       self.rx_idx = 0
       self.rx_done = False
       if self.debug:
-        print(f"ISO-TP: RX - first frame - idx={self.rx_idx} done={self.rx_done}")
+        print(f"ISO-TP: RX - first frame - {hex(self._can_client.rx_addr)} idx={self.rx_idx} done={self.rx_done}")
       if self.debug:
-        print("ISO-TP: TX - flow control continue")
+        print(f"ISO-TP: TX - {hex(self._can_client.tx_addr)} flow control continue")
       # send flow control message (send all bytes)
       msg = b"\x30\x00\x00".ljust(self.max_len, b"\x00")
       self._can_client.send([msg])
@@ -446,7 +455,7 @@ class IsoTpMessage():
       if self.rx_len == len(self.rx_dat):
         self.rx_done = True
       if self.debug:
-        print(f"ISO-TP: RX - consecutive frame - idx={self.rx_idx} done={self.rx_done}")
+        print(f"ISO-TP: RX - consecutive frame - {hex(self._can_client.rx_addr)} idx={self.rx_idx} done={self.rx_done}")
       return
 
     # flow control
@@ -456,7 +465,7 @@ class IsoTpMessage():
       assert rx_data[0] == 0x30 or rx_data[0] == 0x31, "isotp - rx: flow-control transfer state indicator invalid"
       if rx_data[0] == 0x30:
         if self.debug:
-          print("ISO-TP: RX - flow control continue")
+          print(f"ISO-TP: RX - {hex(self._can_client.rx_addr)} flow control continue")
         delay_ts = rx_data[2] & 0x7F
         # scale is 1 milliseconds if first bit == 0, 100 micro seconds if first bit == 1
         delay_div = 1000. if rx_data[2] & 0x80 == 0 else 10000.
@@ -478,11 +487,11 @@ class IsoTpMessage():
         if end >= self.tx_len:
           self.tx_done = True
         if self.debug:
-          print(f"ISO-TP: TX - consecutive frame - idx={self.tx_idx} done={self.tx_done}")
+          print(f"ISO-TP: TX - consecutive frame - {hex(self._can_client.tx_addr)} idx={self.tx_idx} done={self.tx_done}")
       elif rx_data[0] == 0x31:
         # wait (do nothing until next flow control message)
         if self.debug:
-          print("ISO-TP: TX - flow control wait")
+          print(f"ISO-TP: TX - {hex(self._can_client.tx_addr)} flow control wait")
 
 FUNCTIONAL_ADDRS = [0x7DF, 0x18DB33F1]
 
@@ -544,6 +553,7 @@ class UdsClient():
           error_desc = resp[3:].hex()
         # wait for another message if response pending
         if error_code == 0x78:
+          isotp_msg.response_pending = True
           if self.debug:
             print("UDS-RX: response pending")
           continue
@@ -654,7 +664,7 @@ class UdsClient():
     resp = self._uds_request(SERVICE_TYPE.READ_DATA_BY_IDENTIFIER, subfunction=None, data=data)
     resp_id = struct.unpack('!H', resp[0:2])[0] if len(resp) >= 2 else None
     if resp_id != data_identifier_type:
-      raise ValueError('invalid response data identifier: {}'.format(hex(resp_id)))
+      raise ValueError('invalid response data identifier: {} expected: {}'.format(hex(resp_id), hex(data_identifier_type)))
     return resp[2:]
 
   def read_memory_by_address(self, memory_address: int, memory_size: int, memory_address_bytes: int = 4, memory_size_bytes: int = 1):
--- a/python/uds.py
+++ b/python/uds.py
@@ -355,7 +355,7 @@ class CanClient():
         self._recv_buffer()
 
 class IsoTpMessage():
-  def __init__(self, can_client: CanClient, timeout: float = 1, debug: bool = False, max_len: int = 8, response_pending_timeout: float = 5):
+  def __init__(self, can_client: CanClient, timeout: float = 1, debug: bool = False, max_len: int = 8, response_pending_timeout: float = 6):
     self._can_client = can_client
     self.timeout = timeout
     self.debug = debug
@@ -509,7 +509,7 @@ def get_rx_addr_for_tx_addr(tx_addr, rx_offset=0x8):
 
 
 class UdsClient():
-  def __init__(self, panda, tx_addr: int, rx_addr: int = None, bus: int = 0, timeout: float = 1, debug: bool = False, response_pending_timeout: float = 5):
+  def __init__(self, panda, tx_addr: int, rx_addr: int = None, bus: int = 0, timeout: float = 1, debug: bool = False, response_pending_timeout: float = 6):
     self.bus = bus
     self.tx_addr = tx_addr
     self.rx_addr = rx_addr if rx_addr is not None else get_rx_addr_for_tx_addr(tx_addr)


jyoung8607:
I ran into this in #681. I think the more correct solution is to extend the timeout a bit each time we receive a new frame, because that will enable us to finish reading the response, and also make sure we don't try to process the still-arriving replies as part of the next query. Either way, I do absolutely agree these faults should be exceptions rather than asserts.

jyoung8607:
By the way, I'd be very interested to know what the query tool sees from your car!

martinl:
working script output (Subaru XV 2018):
```
./panda/examples/query_fw_versions.py --nonstandard

*** Results for address 0x735 ***


0xF100 IDENTIFICATION_OPTION_VEHICLE_MANUFACTURER_SPECIFIC_DATA_IDENTIFIER: b'\x00\x00\x01'
0xF182 APPLICATION_DATA_IDENTIFICATION: b'\x04\x06\x02\x00\x01'
0xF186 ACTIVE_DIAGNOSTIC_SESSION: b'\x03'
0xF189 VEHICLE_MANUFACTURER_ECU_SOFTWARE_VERSION_NUMBER: b'5AB0205A46'
0xF18E VEHICLE_MANUFACTURER_KIT_ASSEMBLY_PART_NUMBER: b'ALPS Elec.'
0xF197 SYSTEM_NAME_OR_ENGINE_TYPE: b'Keyless Entry System            '
0xF198 REPAIR_SHOP_CODE_OR_TESTER_SERIAL_NUMBER: b'\xe8\x1a\x11\xed'


*** Results for address 0x746 ***


0xF100 IDENTIFICATION_OPTION_VEHICLE_MANUFACTURER_SPECIFIC_DATA_IDENTIFIER: b'\xdc\x04\x00'
0xF182 APPLICATION_DATA_IDENTIFICATION: b'z\xc0\x0c\x00'
0xF197 SYSTEM_NAME_OR_ENGINE_TYPE: b'Power Steering System           '


*** Results for address 0x747 ***


0xF100 IDENTIFICATION_OPTION_VEHICLE_MANUFACTURER_SPECIFIC_DATA_IDENTIFIER: b'D\x00\x00'
0xF182 APPLICATION_DATA_IDENTIFICATION: b'\x16 \x14\x00'
0xF186 ACTIVE_DIAGNOSTIC_SESSION: b'\x03'
0xF189 VEHICLE_MANUFACTURER_ECU_SOFTWARE_VERSION_NUMBER: b'60341-50'
0xF18E VEHICLE_MANUFACTURER_KIT_ASSEMBLY_PART_NUMBER: b'84051FL021'
0xF197 SYSTEM_NAME_OR_ENGINE_TYPE: b'Automatic Headlight Leveling    '


*** Results for address 0x752 ***


0xF100 IDENTIFICATION_OPTION_VEHICLE_MANUFACTURER_SPECIFIC_DATA_IDENTIFIER: b' \x80\x86'
0xF182 APPLICATION_DATA_IDENTIFICATION: b'A\x85\xcc\x00'
0xF186 ACTIVE_DIAGNOSTIC_SESSION: b'\x03'
0xF197 SYSTEM_NAME_OR_ENGINE_TYPE: b'Integ. Unit mode                '
0xF198 REPAIR_SHOP_CODE_OR_TESTER_SERIAL_NUMBER: b'\xe8\x1a\x11\xed'


*** Results for address 0x753 ***


0xF100 IDENTIFICATION_OPTION_VEHICLE_MANUFACTURER_SPECIFIC_DATA_IDENTIFIER: b'\xca\x04\x00'
0xF182 APPLICATION_DATA_IDENTIFICATION: b'z\xaa\x01\x00'
0xF197 SYSTEM_NAME_OR_ENGINE_TYPE: b'Tire pressure monitor           '


*** Results for address 0x780 ***


0xF100 IDENTIFICATION_OPTION_VEHICLE_MANUFACTURER_SPECIFIC_DATA_IDENTIFIER: b'W\x00\x00'
0xF182 APPLICATION_DATA_IDENTIFICATION: b'\x00\x92\x15\x16\x00'
0xF186 ACTIVE_DIAGNOSTIC_SESSION: b'\x03'
0xF18E VEHICLE_MANUFACTURER_KIT_ASSEMBLY_PART_NUMBER: b'98221FL160'
0xF197 SYSTEM_NAME_OR_ENGINE_TYPE: b'Airbag System\x7f\x7f\x7f\x7f\x7f\x7f\x7f\x7f\x7f\x7f\x7f\x7f\x7f\x7f\x7f\x7f\x7f\x7f\x7f'
0xF198 REPAIR_SHOP_CODE_OR_TESTER_SERIAL_NUMBER: b'\xe8\x1a\x11\xed'


*** Results for address 0x783 ***


0xF100 IDENTIFICATION_OPTION_VEHICLE_MANUFACTURER_SPECIFIC_DATA_IDENTIFIER: b'M\x00\x03'
0xF182 APPLICATION_DATA_IDENTIFICATION: b'143\x00\x07'
0xF186 ACTIVE_DIAGNOSTIC_SESSION: b'\x03'
0xF189 VEHICLE_MANUFACTURER_ECU_SOFTWARE_VERSION_NUMBER: b'143A210000'
0xF18E VEHICLE_MANUFACTURER_KIT_ASSEMBLY_PART_NUMBER: b'85002FL100'
0xF197 SYSTEM_NAME_OR_ENGINE_TYPE: b'METER                           '
0xF198 REPAIR_SHOP_CODE_OR_TESTER_SERIAL_NUMBER: b'\x00\x00\x00\x00'


*** Results for address 0x787 ***


0xF100 IDENTIFICATION_OPTION_VEHICLE_MANUFACTURER_SPECIFIC_DATA_IDENTIFIER: b'\xf0\xe0\x0e'
0xF182 APPLICATION_DATA_IDENTIFICATION: b'\x00\x00d\xb5\x1f@ \x0e'
0xF186 ACTIVE_DIAGNOSTIC_SESSION: b'\x03'
0xF189 VEHICLE_MANUFACTURER_ECU_SOFTWARE_VERSION_NUMBER: b'0000000000000'
0xF18E VEHICLE_MANUFACTURER_KIT_ASSEMBLY_PART_NUMBER: b'0000000000'
0xF197 SYSTEM_NAME_OR_ENGINE_TYPE: b'EyeSight System                 '


*** Results for address 0x7B0 ***


0xF100 IDENTIFICATION_OPTION_VEHICLE_MANUFACTURER_SPECIFIC_DATA_IDENTIFIER: b'\xb2\x06\x04'
0xF182 APPLICATION_DATA_IDENTIFICATION: b'z\x94?\x90\x00'
0xF186 ACTIVE_DIAGNOSTIC_SESSION: b'\x03'
0xF197 SYSTEM_NAME_OR_ENGINE_TYPE: b'VDC/Parking Brake System        '
0xF198 REPAIR_SHOP_CODE_OR_TESTER_SERIAL_NUMBER: b'\x11\x11\x11\x11'


*** Results for address 0x7C4 ***


0xF100 IDENTIFICATION_OPTION_VEHICLE_MANUFACTURER_SPECIFIC_DATA_IDENTIFIER: b'\xac\x02\x00'
0xF182 APPLICATION_DATA_IDENTIFICATION: b'\x12\x00\x10\x00\x01'
0xF186 ACTIVE_DIAGNOSTIC_SESSION: b'\x03'
0xF189 VEHICLE_MANUFACTURER_ECU_SOFTWARE_VERSION_NUMBER: b'G2V ver2.30'
0xF197 SYSTEM_NAME_OR_ENGINE_TYPE: b'Air Condition System            '


*** Results for address 0x7D0 ***


0xF100 IDENTIFICATION_OPTION_VEHICLE_MANUFACTURER_SPECIFIC_DATA_IDENTIFIER: b'\x00\x00\x02'
0xF182 APPLICATION_DATA_IDENTIFICATION: b'\x00\x00\x01'
0xF18E VEHICLE_MANUFACTURER_KIT_ASSEMBLY_PART_NUMBER: b'86201FL450'


*** Results for address 0x7D5 ***


0xF100 IDENTIFICATION_OPTION_VEHICLE_MANUFACTURER_SPECIFIC_DATA_IDENTIFIER: b'0\x00\x00'
0xF186 ACTIVE_DIAGNOSTIC_SESSION: b'\x03'
0xF189 VEHICLE_MANUFACTURER_ECU_SOFTWARE_VERSION_NUMBER: b'0300110030'
0xF18E VEHICLE_MANUFACTURER_KIT_ASSEMBLY_PART_NUMBER: b'85261FL121'
0xF197 SYSTEM_NAME_OR_ENGINE_TYPE: b'MFD                             '
0xF198 REPAIR_SHOP_CODE_OR_TESTER_SERIAL_NUMBER: b'\xe8\x1a\x11\xed'


*** Results for address 0x7E0 ***


0xF100 IDENTIFICATION_OPTION_VEHICLE_MANUFACTURER_SPECIFIC_DATA_IDENTIFIER: b'\xa2\x10\t'
0xF182 APPLICATION_DATA_IDENTIFICATION: b'\xaaafs\x07'
0xF186 ACTIVE_DIAGNOSTIC_SESSION: b'\x03'
0xF187 VEHICLE_MANUFACTURER_SPARE_PART_NUMBER: b'\x00\x00\x00\x00\x00\x00\x00\x00'
0xF190 VIN: b'JF1xxxxxxxxxxxxxx'
0xF197 SYSTEM_NAME_OR_ENGINE_TYPE: b'2.0 DOHC                        '


*** Results for address 0x7E1 ***


0xF100 IDENTIFICATION_OPTION_VEHICLE_MANUFACTURER_SPECIFIC_DATA_IDENTIFIER: b'\xa4\x10@'
0xF182 APPLICATION_DATA_IDENTIFICATION: b'\xe3\xe5F1\x00'
0xF186 ACTIVE_DIAGNOSTIC_SESSION: b'\x03'
0xF197 SYSTEM_NAME_OR_ENGINE_TYPE: b'CVT                             '


*** Results for address 0x7F1 ***


0xF100 IDENTIFICATION_OPTION_VEHICLE_MANUFACTURER_SPECIFIC_DATA_IDENTIFIER: b'W\x00\x00'
0xF182 APPLICATION_DATA_IDENTIFICATION: b'\x00\x92\x15\x16\x00'
0xF186 ACTIVE_DIAGNOSTIC_SESSION: b'\x01'
0xF18E VEHICLE_MANUFACTURER_KIT_ASSEMBLY_PART_NUMBER: b'98221FL160'
0xF197 SYSTEM_NAME_OR_ENGINE_TYPE: b'Airbag System\x7f\x7f\x7f\x7f\x7f\x7f\x7f\x7f\x7f\x7f\x7f\x7f\x7f\x7f\x7f\x7f\x7f\x7f\x7f'
0xF198 REPAIR_SHOP_CODE_OR_TESTER_SERIAL_NUMBER: b'\xe8\x1a\x11\xed'
```

martinl:
As sa side note, I'm looking into options for fixing [Subaru FPv2](https://github.com/commaai/openpilot/pull/1878) ten second delay issue. It is interesting that query_fw_versions.py which is using UdsClient returns all ecu responses without issues after ignition. fw_versions.py which is using IsoTpParallelQuery currently requires about 10 second delay for Subarus before all ecus respond reliably to fw queries

martinl:
I did some more testing and catching AssertionError and ValueError in fw query script is enough to allow the scan to finish without changing uds.py. I'll also try to increase the timeout and see if that resolves the underlying issue.

gregjhogan:
It is most likely a bad idea to ignore these errors.

Here are some things I would look into to try to understand what is happening:
* Is it a skip in the index?  If yes, that sounds like the panda experienced a buffer overflow and dropped some messages and then your fix could potentially result in an incomplete fingerprint list.  Longer timeouts will not be helpful if that is the case.  To fix this uds.py needs to be updated to drain the panda RX buffer more often.  I think the panda `debug_console.py` may output when messages are dropped.
* If the issue is old messages are getting picked up from a previous run of this script, the solution would be to ensure the RX/TX buffers are properly flushed before sending these requests.
* Another possibility is that there is a bug in uds.py causing this.

jyoung8607:
> * If the issue is old messages are getting picked up from a previous run of this script, the solution would be to ensure the RX/TX buffers are properly flushed before sending these requests.

It's not messages from a previous run of the script, it's messages from the same run, same control module, previous query. IMO it's a bug in `uds.py`. An example flow:

* Query for 0xF100 (just an example) goes out
* Very long ISO-TP reply starts coming in, I saw one that was at least 60-80 ISO-TP frames before aborting
* `uds.py` timeout fires because the timeout for the overall response doesn't extend with incoming ISO-TP fragments
* Query for 0xF101 (just an example) goes out
* Next ISO-TP frame that arrives is still part of the 0xF100 response with a frame index way out there
* 0xF101 query asserts() due to the frame index not making sense for the 0xF101 query

I think the UDS library timeout handling needs to extend the timeout while we're still receiving frames. There may need to be two timers, a short timer for initial/follow-up ISO-TP frames, and a long timer for max total time waiting for the full answer.

gregjhogan:
That makes sense, in that case I agree we should change how the timeout works.  I think we only need a timeout for the time between consecutive messages (total amount of time elapsed no longer seems useful).

I think we can just reset `start_time` every time a filtered CAN message is received? (inside this for loop)
https://github.com/commaai/panda/blob/74eebb828c0b11205a6fa18082883415e2cf941f/python/uds.py#L400

I agree it would also be a good idea to change the asserts to specific exceptions.  However, after changing how the timeout works, you really can't scan faster than whatever the max time is between any two consecutive relevant messages come in, so at that point the appropriate action is to increase the timeout (and not ignore these exceptions).

martinl:
I changed current query_fw_versions.py default timeout from 0.2 to 0.3 (uds.py default is 1.0) and it fixed the current issue. I also made the timeout configurable via command line argument to support other slow-responding ecus

gregjhogan:
If the issue is that the ECU responds quickly, but the response is large, we should change how the timeouts work instead (as described above).  Slowing down the scanning like this every time we find a car that returns a bigger response would not be good.

jyoung8607:
Turns out there's at least two new timeout cases to handle, not just the one. I ran `./query_fw_versions.py --rxoffset=0x6a --nonstandard --addr 0x767` (just my Telematics control module) with this patch:

```diff
diff --git a/python/uds.py b/python/uds.py
index 5106e6d..afacd45 100644
--- a/python/uds.py
+++ b/python/uds.py
@@ -395,12 +395,16 @@ class IsoTpMessage():
 
   def recv(self) -> Optional[bytes]:
     start_time = time.time()
+    self.rx_idx_prev = 0
     try:
       while True:
         for msg in self._can_client.recv():
           self._isotp_rx_next(msg)
           if self.tx_done and self.rx_done:
             return self.rx_dat
+          if self.rx_idx > self.rx_idx_prev:
+            start_time = time.time()
+            self.rx_idx_prev = self.rx_idx
         # no timeout indicates non-blocking
         if self.timeout == 0:
           return None
```

Which got me to: `ValueError: invalid response data identifier: 0xf15b`. That being maddeningly almost-useful, I added:

```diff
diff --git a/python/uds.py b/python/uds.py
index 5106e6d..afacd45 100644
--- a/python/uds.py
+++ b/python/uds.py
@@ -654,7 +658,7 @@ class UdsClient():
     resp = self._uds_request(SERVICE_TYPE.READ_DATA_BY_IDENTIFIER, subfunction=None, data=data)
     resp_id = struct.unpack('!H', resp[0:2])[0] if len(resp) >= 2 else None
     if resp_id != data_identifier_type:
-      raise ValueError('invalid response data identifier: {}'.format(hex(resp_id)))
+      raise ValueError('invalid response data identifier: {} expected: {}'.format(hex(resp_id), hex(data_identifier_type)))
     return resp[2:]
 
   def read_memory_by_address(self, memory_address: int, memory_size: int, memory_address_bytes: int = 4, memory_size_bytes: int = 1):
```

Which netted me: `ValueError: invalid response data identifier: 0xf15b expected: 0xf16e`.

Review of a debug-mode run revealed the control module immediately responds to the 0xF15B query with 0x7F2278, negative response to a read data item request, with cause "request correctly received - response pending". The UDS library does recognize and essentially skip over that negative response:

https://github.com/commaai/panda/blob/74eebb828c0b11205a6fa18082883415e2cf941f/python/uds.py#L545-L549

But again we have a case where the timeout needs to extend. In this module's case, it needed a solid five (5) seconds to start returning its response, after the query script had moved on. The library correctly assembled the ISO-TP response, but discarded it because it belonged to a stale request, which is the correct move in this situation.

I'm in full agreement with @gregjhogan about correctly handling the timers, but we'll clearly need special handling for the response pending case. It's definitely going to need a separate and much longer timeout. Fortunately it shouldn't interfere in the common case of skipping truly nonresponsive addresses.

Run with timeout cranked to 5 seconds to workaround the above:

```
0xF15B IDENTIFICATION_OPTION_VEHICLE_MANUFACTURER_SPECIFIC_DATA_IDENTIFIER: b'\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00'
0xF179 IDENTIFICATION_OPTION_VEHICLE_MANUFACTURER_SPECIFIC_DATA_IDENTIFIER: b'0060'
0xF17B IDENTIFICATION_OPTION_VEHICLE_MANUFACTURER_SPECIFIC_DATA_IDENTIFIER: b'\x00\x00\x00'
0xF17C IDENTIFICATION_OPTION_VEHICLE_MANUFACTURER_SPECIFIC_DATA_IDENTIFIER: b'NGD-BO130.11.1701720060'
0xF17D IDENTIFICATION_OPTION_VEHICLE_MANUFACTURER_SPECIFIC_DATA_IDENTIFIER: b'0172'
0xF17F IDENTIFICATION_OPTION_VEHICLE_MANUFACTURER_SPECIFIC_DATA_IDENTIFIER: b'NGD'
0xF182 APPLICATION_DATA_IDENTIFICATION: b'\x0601----------'
0xF186 ACTIVE_DIAGNOSTIC_SESSION: b'\x03'
0xF187 VEHICLE_MANUFACTURER_SPARE_PART_NUMBER: b'5QE035285A '
0xF189 VEHICLE_MANUFACTURER_ECU_SOFTWARE_VERSION_NUMBER: b'0570'
0xF18B ECU_MANUFACTURING_DATE: b'0\x11\x17'
0xF190 VIN: b'WVWVF7AU2JW177386'
0xF191 VEHICLE_MANUFACTURER_ECU_HARDWARE_NUMBER: b'5QE035285A '
0xF197 SYSTEM_NAME_OR_ENGINE_TYPE: b'OCULowMQBUS  '
0xF19A CALIBRATION_REPAIR_SHOP_CODE_OR_CALIBRATION_EQUIPMENT_SERIAL_NUMBER: b'\x80\x00\x00\x04\xab\x92'
0xF19B CALIBRATION_DATE: b'\x18\x031'
0xF19E ODX_FILE: b'EV_OCULowMQB\x00'
0xF1A0 IDENTIFICATION_OPTION_VEHICLE_MANUFACTURER_SPECIFIC: b'V03935268PZ'
0xF1A1 IDENTIFICATION_OPTION_VEHICLE_MANUFACTURER_SPECIFIC: b'0001'
0xF1A2 IDENTIFICATION_OPTION_VEHICLE_MANUFACTURER_SPECIFIC: b'001006'
0xF1A3 IDENTIFICATION_OPTION_VEHICLE_MANUFACTURER_SPECIFIC: b'009'
0xF1A4 IDENTIFICATION_OPTION_VEHICLE_MANUFACTURER_SPECIFIC: b'\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff'
0xF1A5 IDENTIFICATION_OPTION_VEHICLE_MANUFACTURER_SPECIFIC: b'\x00\x06@\x16\x05M'
0xF1A8 IDENTIFICATION_OPTION_VEHICLE_MANUFACTURER_SPECIFIC: b'\x00\x06@\x16\x05M'
0xF1A9 IDENTIFICATION_OPTION_VEHICLE_MANUFACTURER_SPECIFIC: b'\x18\x01\x12'
0xF1AA IDENTIFICATION_OPTION_VEHICLE_MANUFACTURER_SPECIFIC: b'J949 '
0xF1AB IDENTIFICATION_OPTION_VEHICLE_MANUFACTURER_SPECIFIC: b'000110911091124500010041005900230006000300290013000300410059005900090030017100300023002300300029003000140015001500040004004000040004004300430017003100050032002901770179016801800065000400150004005200940094009400940094009400520010001600160029016800320029000300070001000100010001000100010001000100010001000100010001000100040001'
0xF1AE IDENTIFICATION_OPTION_VEHICLE_MANUFACTURER_SPECIFIC: b'\xff'
0xF1AF IDENTIFICATION_OPTION_VEHICLE_MANUFACTURER_SPECIFIC: b'\x01\x00\x00 \x00\x01S\x02\x00\x00\x1e\x05\x077\x03\x00\x00 \x02\x02\x02\x07\x00\x00\x1e\x01\x02\x00\t\x00\x00\x1e\x03\x15\x00\n\x00\x00\x1e\x03p\x03\x1a\x00\x00\x1e\x01\x07\x00'
0xF1D5 IDENTIFICATION_OPTION_VEHICLE_MANUFACTURER_SPECIFIC: b'\x01\x00\x00\x05$\x00\x00'
0xF1DF IDENTIFICATION_OPTION_VEHICLE_MANUFACTURER_SPECIFIC: b'@'
0xF1E0 IDENTIFICATION_OPTION_VEHICLE_MANUFACTURER_SPECIFIC: b'\x00'
```

jyoung8607:
@martinl, as the person who opened this PR/effort, you're welcome to tackle these timeout issues. Or, if you've moved on with other things, I can open a separate PR in a few days. I have some other priorities at this particular moment.

martinl:
I reverted my changes and pushed the changes @jyoung8607 recommended above. I can confirm the original issue with ecu 0x7d0 is fixed but scan also triggered another traceback

```
./panda/examples/query_fw_versions.py
opening device 37003a001651383432363133 0xddcc
connected
querying addresses ...
0x7d5:  42%|████████████████████████████████████████████████████████▏                                                                              | 213/512 [00:49<01:09,  4.32it/s]
  File "./panda/examples/query_fw_versions.py", line 61, in <module>
    data = uds_client.read_data_by_identifier(uds_data_id)  # type: ignore
  File "/data/pythonpath/panda/python/uds.py", line 661, in read_data_by_identifier
    raise ValueError('invalid response data identifier: {} expected: {}'.format(hex(resp_id), hex(data_identifier_type)))
ValueError: invalid response data identifier: 0xf182 expected: 0xf183
```

traceback does not trigger if I add --nonstandard parameter.

I consider it as learning experience to better underestand how UDS and ISO-TP work, so far it has been great :)

martinl:
here is full debug output:
```./panda/examples/query_fw_versions.py --debug --addr 7d5
opening device 37003a001651383432363133 0xddcc
connected
querying addresses ...
0x7d5:   0%|                                                                                                                                                   | 0/1 [00:00<?, ?it/s]ISO-TP: REQUEST - 0x3e00
ISO-TP: TX - single frame
CAN-TX: 0x7d5 - 0x023e000000000000
CAN-RX: 0x7dd - 0x027e000000000000
ISO-TP: RX - single frame - idx=0 done=True
ISO-TP: RESPONSE - 0x7e00
ISO-TP: REQUEST - 0x1001
ISO-TP: TX - single frame
CAN-TX: 0x7d5 - 0x0210010000000000
CAN-RX: 0x7dd - 0x065001003201f400
ISO-TP: RX - single frame - idx=0 done=True
ISO-TP: RESPONSE - 0x5001003201f4
ISO-TP: REQUEST - 0x1003
ISO-TP: TX - single frame
CAN-TX: 0x7d5 - 0x0210030000000000
CAN-RX: 0x7dd - 0x065003003201f400
ISO-TP: RX - single frame - idx=0 done=True
ISO-TP: RESPONSE - 0x5003003201f4
ISO-TP: REQUEST - 0x22f180
ISO-TP: TX - single frame
CAN-TX: 0x7d5 - 0x0322f18000000000
CAN-RX: 0x7dd - 0x037f223100000000
ISO-TP: RX - single frame - idx=0 done=True
ISO-TP: RESPONSE - 0x7f2231
ISO-TP: REQUEST - 0x22f181
ISO-TP: TX - single frame
CAN-TX: 0x7d5 - 0x0322f18100000000
CAN-RX: 0x7dd - 0x037f223100000000
ISO-TP: RX - single frame - idx=0 done=True
ISO-TP: RESPONSE - 0x7f2231
ISO-TP: REQUEST - 0x22f182
ISO-TP: TX - single frame
CAN-TX: 0x7d5 - 0x0322f18200000000
CAN-RX: 0x7dd - 0x037f227800000000
ISO-TP: RX - single frame - idx=0 done=True
ISO-TP: RESPONSE - 0x7f2278
UDS-RX: response pending
ISO-TP: RESPONSE - 0x7f2278
ISO-TP: REQUEST - 0x22f183
ISO-TP: TX - single frame
CAN-TX: 0x7d5 - 0x0322f18300000000
CAN-RX: 0x7dd - 0x100d62f18200fc01
ISO-TP: RX - first frame - idx=0 done=False
ISO-TP: TX - flow control continue
CAN-TX: 0x7d5 - 0x3000000000000000
CAN-RX: 0x7dd - 0x2103000030021100
ISO-TP: RX - consecutive frame - idx=1 done=True
ISO-TP: RESPONSE - 0x62f18200fc0103000030021100
0x7d5:   0%|                                                                                                                                                   | 0/1 [00:00<?, ?it/s]
  File "./panda/examples/query_fw_versions.py", line 61, in <module>
    data = uds_client.read_data_by_identifier(uds_data_id)  # type: ignore
  File "/data/pythonpath/panda/python/uds.py", line 661, in read_data_by_identifier
    raise ValueError('invalid response data identifier: {} expected: {}'.format(hex(resp_id), hex(data_identifier_type)))
ValueError: invalid response data identifier: 0xf182 expected: 0xf183
```

jyoung8607:
Glad you're having fun with it. Fortunately or unfortunately as the case may be, you're having the same issue I did:

```
Q. Please show me data ID 0xF181.
A: I don't have any data with that identifier.

Q: Please show me data ID 0xF182.
A: Sure, but I need a minute for this one, standby!

--query script/uds lib waits, but gives up too fast--

Q: Please show me data ID 0xF183.
A: Here's that 0xF182 data you asked for.

--query script/uds lib are confused by this--
```

0xF182 and 0xF183 are standard queries, so they should be run in both the standard and nonstandard cases, not sure why adding nonstandard into the mix would affect things. I will say that the script only managed to move down one query on the list before getting this response and throwing the exception, which means your current timeout value is right on the knife-edge of success or failure for this particular situation. It should be made *much* longer for the case of a responsive module that actively asked us to standby for a response.

gregjhogan:
It looks like [ISO 14229-2:2013](http://www.yikouzu.com/images/CAN_pdf/ISO14229/ISO%2014229-2-2013%E4%BC%9A%E8%AF%9D%E5%B1%82%E6%9C%8D%E5%8A%A1.PDF) describes the different timeouts we should implement (and most have a recommended value)

Also looks like some you can read from the ECU, but that probably isn't useful for this script.

martinl:
I added ISO-TP response pending timeout with default of 5 seconds and it works if I run the script with --debug but does not work if I run without debug..

```
./examples/query_fw_versions.py --addr 7d5 --debug
opening device 37003a001651383432363133 0xddcc
connected
querying addresses ...
0x7d5:   0%|                                                                                                                                                   | 0/1 [00:00<?, ?it/s]CAN-TX: 0x7d5 - 0x023e000000000000
CAN-RX: 0x7dd - 0x027e000000000000
CAN-TX: 0x7d5 - 0x0210010000000000
CAN-RX: 0x7dd - 0x065001003201f400
CAN-TX: 0x7d5 - 0x0210030000000000
CAN-RX: 0x7dd - 0x065003003201f400
CAN-TX: 0x7d5 - 0x0322f18000000000
CAN-RX: 0x7dd - 0x037f223100000000
CAN-TX: 0x7d5 - 0x0322f18100000000
CAN-RX: 0x7dd - 0x037f223100000000
CAN-TX: 0x7d5 - 0x0322f18200000000
CAN-RX: 0x7dd - 0x037f227800000000
UDS-RX: response pending
CAN-RX: 0x7dd - 0x100d62f18200fc01
CAN-TX: 0x7d5 - 0x3000000000000000
CAN-RX: 0x7dd - 0x2103000030021100
CAN-TX: 0x7d5 - 0x0322f18300000000
CAN-RX: 0x7dd - 0x037f223100000000
CAN-TX: 0x7d5 - 0x0322f18400000000
CAN-RX: 0x7dd - 0x037f223100000000
CAN-TX: 0x7d5 - 0x0322f18500000000
CAN-RX: 0x7dd - 0x037f223100000000
CAN-TX: 0x7d5 - 0x0322f18600000000
CAN-RX: 0x7dd - 0x0462f18603000000
CAN-TX: 0x7d5 - 0x0322f18700000000
CAN-RX: 0x7dd - 0x037f223100000000
CAN-TX: 0x7d5 - 0x0322f18800000000
CAN-RX: 0x7dd - 0x037f223100000000
CAN-TX: 0x7d5 - 0x0322f18900000000
CAN-RX: 0x7dd - 0x037f227800000000
UDS-RX: response pending
CAN-RX: 0x7dd - 0x100d62f189303330
CAN-TX: 0x7d5 - 0x3000000000000000
CAN-RX: 0x7dd - 0x2130313130303330
CAN-TX: 0x7d5 - 0x0322f18a00000000
CAN-RX: 0x7dd - 0x037f223100000000
CAN-TX: 0x7d5 - 0x0322f18b00000000
CAN-RX: 0x7dd - 0x037f223100000000
CAN-TX: 0x7d5 - 0x0322f18c00000000
CAN-RX: 0x7dd - 0x037f223100000000
CAN-TX: 0x7d5 - 0x0322f18d00000000
CAN-RX: 0x7dd - 0x037f223100000000
CAN-TX: 0x7d5 - 0x0322f18e00000000
CAN-RX: 0x7dd - 0x037f227800000000
UDS-RX: response pending
CAN-RX: 0x7dd - 0x100d62f18e383532
CAN-TX: 0x7d5 - 0x3000000000000000
CAN-RX: 0x7dd - 0x213631464c313231
CAN-TX: 0x7d5 - 0x0322f19000000000
CAN-RX: 0x7dd - 0x037f223100000000
CAN-TX: 0x7d5 - 0x0322f19100000000
CAN-RX: 0x7dd - 0x037f223100000000
CAN-TX: 0x7d5 - 0x0322f19200000000
CAN-RX: 0x7dd - 0x037f223100000000
CAN-TX: 0x7d5 - 0x0322f19300000000
CAN-RX: 0x7dd - 0x037f223100000000
CAN-TX: 0x7d5 - 0x0322f19400000000
CAN-RX: 0x7dd - 0x037f223100000000
CAN-TX: 0x7d5 - 0x0322f19500000000
CAN-RX: 0x7dd - 0x037f223100000000
CAN-TX: 0x7d5 - 0x0322f19600000000
CAN-RX: 0x7dd - 0x037f223100000000
CAN-TX: 0x7d5 - 0x0322f19700000000
CAN-RX: 0x7dd - 0x102362f1974d4644
CAN-TX: 0x7d5 - 0x3000000000000000
CAN-RX: 0x7dd - 0x2120202020202020
CAN-RX: 0x7dd - 0x2220202020202020
CAN-RX: 0x7dd - 0x2320202020202020
CAN-RX: 0x7dd - 0x2420202020202020
CAN-RX: 0x7dd - 0x2520000000000000
CAN-TX: 0x7d5 - 0x0322f19800000000
CAN-RX: 0x7dd - 0x0762f198e81a11ed
CAN-TX: 0x7d5 - 0x0322f19900000000
CAN-RX: 0x7dd - 0x037f223100000000
CAN-TX: 0x7d5 - 0x0322f19a00000000
CAN-RX: 0x7dd - 0x037f223100000000
CAN-TX: 0x7d5 - 0x0322f19b00000000
CAN-RX: 0x7dd - 0x037f223100000000
CAN-TX: 0x7d5 - 0x0322f19c00000000
CAN-RX: 0x7dd - 0x037f223100000000
CAN-TX: 0x7d5 - 0x0322f19d00000000
CAN-RX: 0x7dd - 0x037f223100000000
CAN-TX: 0x7d5 - 0x0322f19e00000000
CAN-RX: 0x7dd - 0x037f223100000000
CAN-TX: 0x7d5 - 0x0322f19f00000000
CAN-RX: 0x7dd - 0x037f223100000000
0x7d5: 100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 1/1 [00:01<00:00,  1.54s/it]


*** Results for address 0x7D5 ***


0xF182 APPLICATION_DATA_IDENTIFICATION: b'\x00\xfc\x01\x03\x00\x000\x02\x11\x00'
0xF186 ACTIVE_DIAGNOSTIC_SESSION: b'\x03'
0xF189 VEHICLE_MANUFACTURER_ECU_SOFTWARE_VERSION_NUMBER: b'0300110030'
0xF18E VEHICLE_MANUFACTURER_KIT_ASSEMBLY_PART_NUMBER: b'85261FL121'
0xF197 SYSTEM_NAME_OR_ENGINE_TYPE: b'MFD                             '
0xF198 REPAIR_SHOP_CODE_OR_TESTER_SERIAL_NUMBER: b'\xe8\x1a\x11\xed'
```
without debug:
```
./examples/query_fw_versions.py --addr 7d5
opening device 37003a001651383432363133 0xddcc
connected
querying addresses ...
0x7d5:   0%|                                                                                                                                                   | 0/1 [00:00<?, ?it/s]
  File "./examples/query_fw_versions.py", line 61, in <module>
    data = uds_client.read_data_by_identifier(uds_data_id)  # type: ignore
  File "/data/pythonpath/panda/python/uds.py", line 667, in read_data_by_identifier
    raise ValueError('invalid response data identifier: {} expected: {}'.format(hex(resp_id), hex(data_identifier_type)))
ValueError: invalid response data identifier: 0xf182 expected: 0xf184
```

martinl:
I fixed the issue with function parameter order and tested that the example script finishes the scan successfully in all cases. I also added CAN RX/TX addresses to ISO-TP TX and RX messages debug output for better readability, in case there are requests and responses to/from multiple ecus. Is the default 5 second timeout for response pending acceptable?

python/uds.py
```diff
@@ -428,9 +437,9 @@ def _isotp_rx_next(self, rx_data: bytes) -> None:
       self.rx_idx = 0
       self.rx_done = False
       if self.debug:
-        print(f"ISO-TP: RX - first frame - idx={self.rx_idx} done={self.rx_done}")
+        print(f"ISO-TP: RX - first frame - {hex(self._can_client.rx_addr)} idx={self.rx_idx} done={self.rx_done}")
       if self.debug:
-        print("ISO-TP: TX - flow control continue")
+        print(f"ISO-TP: TX - {hex(self._can_client.tx_addr)} flow control continue")
```
gregjhogan, line 442: let's be consistent and always put the data at the end (after text description)

python/uds.py
```diff
@@ -376,39 +378,46 @@ def send(self, dat: bytes) -> None:
...
       if self.debug:
-        print("ISO-TP: TX - single frame")
+        print(f"ISO-TP: TX - {hex(self._can_client.tx_addr)} single frame")
       msg = (bytes([self.tx_len]) + self.tx_dat).ljust(self.max_len, b"\x00")
       self.tx_done = True
     else:
       # first frame (send first 6 bytes)
       if self.debug:
-        print("ISO-TP: TX - first frame")
+        print(f"ISO-TP: TX - {hex(self._can_client.tx_addr)} first frame")
```
gregjhogan, line 394: let's be consistent and always put the addr/data at the end (after text description)

python/uds.py
```diff
@@ -456,7 +465,7 @@ def _isotp_rx_next(self, rx_data: bytes) -> None:
       assert rx_data[0] == 0x30 or rx_data[0] == 0x31, "isotp - rx: flow-control transfer state indicator invalid"
       if rx_data[0] == 0x30:
         if self.debug:
-          print("ISO-TP: RX - flow control continue")
+          print(f"ISO-TP: RX - {hex(self._can_client.rx_addr)} flow control continue")
```
gregjhogan, line 468: let's be consistent and always put the addr/data at the end (after text description)

python/uds.py
```diff
@@ -478,11 +487,11 @@ def _isotp_rx_next(self, rx_data: bytes) -> None:
         if end >= self.tx_len:
           self.tx_done = True
         if self.debug:
-          print(f"ISO-TP: TX - consecutive frame - idx={self.tx_idx} done={self.tx_done}")
+          print(f"ISO-TP: TX - consecutive frame - {hex(self._can_client.tx_addr)} idx={self.tx_idx} done={self.tx_done}")
       elif rx_data[0] == 0x31:
         # wait (do nothing until next flow control message)
         if self.debug:
-          print("ISO-TP: TX - flow control wait")
+          print(f"ISO-TP: TX - {hex(self._can_client.tx_addr)} flow control wait")
```
gregjhogan, line 494: let's be consistent and always put the addr/data at the end (after text description)

python/uds.py
```diff
@@ -376,39 +378,46 @@ def send(self, dat: bytes) -> None:
...
     start_time = time.time()
+    self.rx_idx_prev = 0
     try:
       while True:
         for msg in self._can_client.recv():
           self._isotp_rx_next(msg)
           if self.tx_done and self.rx_done:
             return self.rx_dat
+          if self.rx_idx > self.rx_idx_prev:
+            start_time = time.time()
```
gregjhogan, line 408: I was thinking we can just reset `start_time` every time a message is received (no need to add anything new tracking previous idx), is there a reason we shouldn't do that (I believe we are guaranteed `self._can_client.recv()` returning something guarantees a relevant message came in)?

python/uds.py
```diff
@@ -376,39 +378,46 @@ def send(self, dat: bytes) -> None:
...
+            self.rx_idx_prev = self.rx_idx
         # no timeout indicates non-blocking
         if self.timeout == 0:
           return None
-        if time.time() - start_time > self.timeout:
+        if self.response_pending:
+          if time.time() - start_time > self.response_pending_timeout:
+            raise MessageTimeoutError("timeout waiting for pending response")
```
gregjhogan, line 415: it looks like `response_pending` is never reset to `False` but I think there might be a better way to build this:
* in `_uds_request` use a local variable to track if a response is pending
* add an optional param to this `recv()` method which specifies the timeout for only the first result from `_isotp_rx_next` (subsequent responses respect the normal timeout)

decreasing the scope of where variables are set/reset should decrease the chance of bugs in the future and make the code easier to understand

python/uds.py
```diff
@@ -355,11 +355,13 @@ def send(self, msgs: List[bytes], delay: float = 0) -> None:
         self._recv_buffer()
 
 class IsoTpMessage():
-  def __init__(self, can_client: CanClient, timeout: float = 1, debug: bool = False, max_len: int = 8):
+  def __init__(self, can_client: CanClient, timeout: float = 1, debug: bool = False, max_len: int = 8, response_pending_timeout: float = 5):
```
gregjhogan, line 358: move `response_pending_timeout` up to `UdsClient` because this is a UDS concept (not ISO-TP)

python/uds.py
jyoung8607: At the time I assumed `_can_client.recv()` would hand back whatever was on the wire and depend on `_isotp_rx_next()` to check for relevance and sanity. I never really checked that assumption. If you're sure we can rely on `_can_client.recv()` to only return messages of relevance to our diagnostic conversation, the `prev` stuff can be elided.

gregjhogan: I reviewed how it works (`CanClient` is pretty simple) and came to the above conclusion.  If I missed something let me know!

python/uds.py
martinl: done, I hope I understood this correctly

python/uds.py
martinl: done

python/uds.py
martinl: done

python/uds.py
martinl: done

python/uds.py
martinl: done, tested and worked for my usecase

python/uds.py
martinl: Added `response_pending = False` after `_isotp_rx_next`

python/uds.py
martinl: done

jyoung8607:
I haven't tested this version, and I won't be able to until Sunday evening at the earliest, but on a quick glance it's consistent with my understanding of what we need to do.

> Is the default 5 second timeout for response pending acceptable?

I made that up based on observed behavior from my car, but it turns out to be reasonably close. ISO 14229-2:2013 section 7.2 specifies a timer `P2*Client` for this case, which is essentially `P2*Server` (worst-case module application response time) plus worst case network latency. `P2*Server` alone is given a recommended (but not required) value of 5000ms (5 seconds).

I propose we use a 6 second timer to accommodate that plus worst case network latency. That will handle anything reasonably within the ISO recommendation. If we ever see a real-world need to increase it, we can do so later. Also, my reading of that same ISO 14229-2:2013 section leads me to believe the application should be sending additional periodic negative response/pending replies if it's going to take really long.

gregjhogan:
migrated to #723

PR #700, opened 2021-07-27
fix kernel 5.11 and above driver compilation

cydia2020:
Driver doesn't compile on kernel 5.11 and above, throws `error: implicit declaration of function ‘get_can_dlc’`

### DIFF
--- a/drivers/linux/panda.c
+++ b/drivers/linux/panda.c
@@ -283,7 +283,7 @@ static void panda_usb_process_can_rx(struct panda_dev_priv *priv_dev,
   //if (msg->dlc & MCBA_DLC_RTR_MASK)
   //  cf->can_id |= CAN_RTR_FLAG;
 
-  cf->can_dlc = get_can_dlc(msg->bus_dat_len & PANDA_DLC_MASK);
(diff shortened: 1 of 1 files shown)


drivers/linux/panda.c
```diff
@@ -283,7 +283,7 @@ static void panda_usb_process_can_rx(struct panda_dev_priv *priv_dev,
   //if (msg->dlc & MCBA_DLC_RTR_MASK)
   //  cf->can_id |= CAN_RTR_FLAG;
 
-  cf->can_dlc = get_can_dlc(msg->bus_dat_len & PANDA_DLC_MASK);
+  cf->can_dlc = (min_t(__u8, (msg->bus_dat_len & PANDA_DLC_MASK), 8));
```
adeebshihadeh, line 286: looks like this got renamed to `can_cc_dlc2len` https://lwn.net/Articles/836683/

cydia2020: Will try later

cydia2020: This does work and it seems to be reliable, but I'm not familiar with module development, so idk how I would go about differentiating between between different versions of different kernels.

Assistance in this regard would be appreciated.

adeebshihadeh:
#765 supports the older macro and newer one

PR #701, opened 2021-08-03
Cleanup unused functions in boards configs

briskspirit:
Cleanup unused functions in boards configs


adeebshihadeh:
Moved unused functions to board_declarations?

briskspirit:
> Moved unused functions to board_declarations?

No, moved to unused_funcs.h in prev PR for red and now cleaned other. board_declarations should stay clean.

PR #707, opened 2021-08-27
VW MQB and PQ: Fix controls entry conditions and tests

jyoung8607:
Panda isn't currently handling `controls_allowed` for Volkswagen quite the same as other ports. We're only supposed to enter controls on rising edge of ACC engagement, but right now we're considering controls allowed whenever ACC is engaged. As far as I can tell, the real-world impact of this is pretty low, but this should be fixed anyway.

First commit will revert the overridden tests that allowed this to go undetected, reverting to the common inherited tests that look for this behavior. Second commit will fix the safety to pass the new tests.

Planning to hold this in draft for a short time, while I make sure this is working as-planned for community users.


jyoung8607:
This doesn't seem to have any unexpected effects on community. Ready for review at your convenience.

There is no corresponding openpilot PR. The panda submodule ref can be bumped at leisure.

PR #715, opened 2021-09-07
enable siren if controls allowed within 30s of lost heartbeat

adeebshihadeh:


board/main.c
```diff
@@ -703,15 +703,24 @@ void tick_handler(void) {
         siren_countdown -= 1U;
       }
 
+      if (controls_allowed) {
+        controls_allowed_countdown = 30U;
+      } else if (controls_allowed_countdown > 0U) {
+        controls_allowed_countdown -= 1U;
+      } else {
```
robbederks, line 710: this empty else seems like a dumb rule, maybe we should consider removing it from the misra check?

adeebshihadeh: if we want to do this, let's do it in another PR

PR #721, opened 2021-09-15
Trying to force acc_type1 for enable SnG skills

AlexandreSato:
ATTENTION this is not working here. I created this script https://github.com/AlexandreSato/openpilot-Brazil/blob/master/acc_type1.py which if run during engine start enables SNG on global TSS2.0 Toyotas without native stop and go. Tested models: Rav4, corolla, corollacross


adeebshihadeh:
This should be implemented in openpilot. Panda's job is to enforce the safety model and never modifies messages.

PR #724, opened 2021-09-20
Panda partial fix for GM Steering Temporarily Unavailable issue

JasonJShuler:
### Summary
Partial fix commaai/openpilot#21557
Followup to PRs #697 #432 
Partial resolution for LKAS Steering Temporarily Unavailable Issue on some newer GM vehicles.

Problems fixed:
- Engage / disengage related steering faults
- Steering faults on gas or brake
- Steering faults resulting from **Panda initiated** dropped frames (i.e. torque out of range)
- Unexpected behavior of all PSCM variants when LKAS frames are "missing". I.E. if we send no frames, how long does the PSCM continue applying torque? Do we want it to continue applying the last received torque or apply no torque? (I'm assuming no torque - please let me know otherwise")

Unresolved issues:
- Faults caused by a message arriving faster than every 20ms (Will reduce incidence by slowing the OP message pump)
- Steering fault triggered by relay fault (probably unavoidable)
- Engagement faults caused by misalignment of the OP and stock camera rolling counters while closing the relay (will attempt to resolve in OP)

### Change Description
_This issue only addresses frames that were previously being dropped by the Panda without OP involvement - as such this is the minimal set of changes that can only be made on the Panda._

This simple change should mitigate the vast majority of non-timing related faults. A combination of this change, and slowing the GM LKAS rate to every 30 or 40ms instead of every 20ms _should_ **dramatically** reduce the likelihood of a random fault to a negligible (but still present) level.

This PR changes the behavior on the Panda when a GM LKAS frame is flagged as a violation. Rather than drop the frame, it is replaced with an equivalent "Inactive, zero torque" frame.
Two lines added to safety_gm.h, one line replaced. Apologies for the verbosity - it is a simple change, but **does not add any** new non-safety functionality. _Please provide me with guidance if you believe this specific piece of the fix can be implemented anywhere else!_

- Added const array with the pre-computed, zeroed out & inactive LKAS messages (Since we are avoiding new non-safety code on the Panda I opted to use hard coded values  - let me know if you would prefer I calculate the checksum)
- Extract the rolling counter value from the LKAS frame (low impact bit shifting)
- When a violation occurs, instead of setting tx=0, we send an "empty" frame

### Problem Description
The Power Steering Control Module on the Chevy Bolt EV as well as several other post-2018 GM vehicles contain two differences from past PSCMs: 1. They strongly enforce the order of the rolling counter for LKAS messages. 2. They improperly implement the CANBUS protocol such that LKAS messages received faster than every 20ms overwrite the previous value and create the appearance of a rolling counter gap.

As mentioned in my past PR, there are two situations that cause the GM steering fault: actual dropped frames, and phantom dropped frames resulting from messages coming through too fast.

Thanks!

### DIFF
--- a/board/safety/safety_gm.h
+++ b/board/safety/safety_gm.h
@@ -22,6 +22,8 @@ const CanMsg GM_TX_MSGS[] = {{384, 0, 4}, {1033, 0, 7}, {1034, 0, 7}, {715, 0, 8
                              {161, 1, 7}, {774, 1, 8}, {776, 1, 7}, {784, 1, 2},   // obs bus
                              {789, 2, 5},  // ch bus
                              {0x104c006c, 3, 3}, {0x10400060, 3, 5}};  // gmlan
+//precalculated inactive zero values to be sent when there is a violation or inactivation of LKAS
+const uint32_t GM_LKAS_BLANK[] = {0x00000000U,0x10000fffU,0x20000ffeU,0x30000ffdU};
 
 // TODO: do checksum and counter checks. Add correct timestep, 0.1s for now.
 AddrCheckStruct gm_addr_checks[] = {
@@ -142,6 +144,7 @@ static int gm_tx_hook(CAN_FIFOMailBox_TypeDef *to_send) {
 
   // LKA STEER: safety check
   if (addr == 384) {
+    int rolling_counter = GET_BYTE(to_send, 0) >> 4;
     int desired_torque = ((GET_BYTE(to_send, 0) & 0x7U) << 8) + GET_BYTE(to_send, 1);
     uint32_t ts = microsecond_timer_get();
     bool violation = 0;
@@ -184,7 +187,7 @@ static int gm_tx_hook(CAN_FIFOMailBox_TypeDef *to_send) {
     }
 
     if (violation) {
-      tx = 0;
+      to_send->RDLR = GM_LKAS_BLANK[rolling_counter];
     }
   }
 


Verylukyguy:
I will be more than happy to test this immediately; many users need a upstreamable resolution for this issues, not to mention all of the people looking to install a Comma device and OpenPilot on their GM vehicles.

jyoung8607:
> Faults caused by a message arriving faster than every 20ms (Will reduce incidence by slowing the OP message pump)

As you say, and as some of the test users have verified, it looks like we can help a great deal by slowing the send rate, and by curing other potential sources of lag in openpilot. For the remainder, I think this particular approach will continue to be a hard sell, because it still causes Panda to originate messages when that is not its role under the current openpilot safety architecture.

However, that doesn't mean there's nothing we can do:

> Engage / disengage related steering faults
> Steering faults on gas or brake

@adeebshihadeh, I think the main problem here is Panda policing an openpilot steer message while the corresponding disengage message was still in-flight. What do you think about adding one frame of controls_allowed tolerance before steering messages are dropped, to account for in-flight disengagement events, [similar to what openpilot does today](https://github.com/commaai/openpilot/blob/c4bac6bd68ce2a91beb15bb4f4472d76525288e0/selfdrive/controls/controlsd.py#L348-L356)? That would fix these cases without taking Panda out of its lane. I think I would consider this a potentially valid change for ALL car ports, not just GM.

> Steering faults resulting from Panda initiated dropped frames (i.e. torque out of range)

There shouldn't be any cases where openpilot sends torque commands beyond Panda's tolerance. If such cases exist, the GM Panda safety and GM openpilot car port are out of sync and should be fixed.

> Unexpected behavior of all PSCM variants when LKAS frames are "missing". I.E. if we send no frames, how long does the PSCM continue applying torque? Do we want it to continue applying the last received torque or apply no torque? (I'm assuming no torque - please let me know otherwise")
> Steering fault triggered by relay fault (probably unavoidable)

There exist cases where the EPS *should* probably raise a fault, and these are two examples.

adeebshihadeh:
Again, we cannot merge this. The panda's job is to enforce the safety model, and **never** needs to generate or modify messages to do so. I'm not sure how many other ways we can say it, but the issue you're trying to solve should be fixed in openpilot, unless there's some fundamental panda bug.

> This PR changes the behavior on the Panda when a GM LKAS frame is flagged as a violation. Rather than drop the frame, it is replaced with an equivalent "Inactive, zero torque" frame.
> Two lines added to safety_gm.h, one line replaced. Apologies for the verbosity - it is a simple change, but **does not add any** new non-safety functionality. _Please provide me with guidance if you believe this specific piece of the fix can be implemented anywhere else!_

As @jyoung8607 said, this is working as designed.

> @adeebshihadeh, I think the main complaint here is Panda policing an openpilot steer message while the corresponding engage or disengage message was still in-flight. What do you think about adding one frame of controls_allowed tolerance before steering messages are dropped, to account for in-flight disengagement events, [similar to what openpilot does today](https://github.com/commaai/openpilot/blob/c4bac6bd68ce2a91beb15bb4f4472d76525288e0/selfdrive/controls/controlsd.py#L348-L356)? That would fix these cases without taking Panda out of its lane. I think I would consider this a potentially valid change for ALL car ports, not just GM.

Yes, we can explore something along these lines.

JasonJShuler:
Hey, thanks so much for the reply! Some responses:

> As you say, and as some of the test users have verified, it looks like we can help a great deal by slowing the send rate, and by curing other potential sources of lag in openpilot. For the remainder, I think this particular approach will continue to be a hard sell, because it still causes Panda to originate messages when that is not its role under the current openpilot safety architecture.

1. This PR isn't addressing the timing issue, only the Panda dropping frames when it should not be.
2. I don't see this PR causing the Panda to originate a message... it is neutralizing it the correct way. The panda only sends a message when received by OP. Instead of blocking the frames outright when there is a violation, we are instead making the value zero/inactive. This value is normally permitted by the Panda at all times, even when controls aren't allowed and/or there is a violation. Also, keep in mind that OP never blocks an LKAS frame - if there is a problem OP sends a blank frame - dropping frames shouldn't be happening really at all if the car doesn't ever drop frames. This is only making the Panda do it the same way...
2. Note that AFAIK the rate slowing has only worked when combined with customized Panda firmware. I am somewhat out of the loop though.


> @adeebshihadeh, I think the main complaint here is Panda policing an openpilot steer message while the corresponding disengage message was still in-flight. What do you think about adding one frame of controls_allowed tolerance before steering messages are dropped, to account for in-flight disengagement events, [similar to what openpilot does today](https://github.com/commaai/openpilot/blob/c4bac6bd68ce2a91beb15bb4f4472d76525288e0/selfdrive/controls/controlsd.py#L348-L356)? That would fix these cases without taking Panda out of its lane. I think I would consider this a potentially valid change for ALL car ports, not just GM.

1. This may take care of most disengage faults, but that delay should be at least 2 frames. Who knows how far OP is behind the Panda? Isn't it variable? Perhaps something based on wall clock time rather than frame count? (This is still a vulnerable workaround)
2. Indeed - at least for disengage - the operations are such: Driver taps brake, Panda flips controls_allowed flag near-instantly and drops any non-zero torque values, OP sends an undetermined number of frames before reacting to the disengage and begins sending zero-valued LKAS messages, The Panda - seeing zeroed values allows the messages through again, 75% of the time, the resumed message has the wrong rolling counter and faults permanently.


> There shouldn't be any cases where openpilot sends torque commands beyond Panda's tolerance. If such cases exist, the GM Panda safety and GM openpilot car port are out of sync and should be fixed.

1. There shouldn't, but there are, and with some frequency. (Or at least used to)  One of the rate limits is based on the elapsed time since the last message **calculated independently by unsynchronized timers**. We already know the 10ms message loop on OP isn't always exactly 10ms. This alone is enough to make the mechanism unreliable. This isn't really a bug - unless we start sending timestamps along with messages from OP, using different clocks on different devices with different architectures at different times to measure a duration is guaranteed not to yield consistent results. Guaranteed - like it can be proven.
2. Suppose we fudge the timings to be less restrictive or to compensate for an allowable timer drift. Fixing the issue in this way is akin to throwing in a "sleep 100" to prevent deadlocks. It may work but not reliably - again we are putting unchecked hard-coded "padding" around a timing issue rather than fixing or preventing it.

> > Unexpected behavior of all PSCM variants when LKAS frames are "missing". I.E. if we send no frames, how long does the PSCM continue applying torque? Do we want it to continue applying the last received torque or apply no torque? (I'm assuming no torque - please let me know otherwise")
> > Steering fault triggered by relay fault (probably unavoidable)
> 
> There exist cases where the EPS _should_ probably raise a fault, and these are two examples.

1. The relay failure is pretty much a guaranteed trip to have your DTCs cleared LOL, no disagreement there.
2. Why are we willing to leave this to chance? Presently the PSCM doesn't appear to fault even after LONG delays. Maybe it will in the future. Maybe in the future this abomination of a PSCM goes away.

JasonJShuler:
Yes, this is a Panda bug. The Panda should not be dropping LKAS frames in GM vehicles.

Period.

adeebshihadeh:
I'm locking this thread. Development discussion is always welcome, but re-hashing the same things isn't productive. Feel free to continue development discussion in the openpilot issue thread.

PR #725, opened 2021-09-24
python: disable power save by default when setting safety mode

gregjhogan:


python/__init__.py
```diff
@@ -453,10 +453,12 @@ def esp_reset(self, bootmode=0):
     self._handle.controlWrite(Panda.REQUEST_OUT, 0xda, int(bootmode), 0, b'')
     time.sleep(0.2)
 
-  def set_safety_mode(self, mode=SAFETY_SILENT, disable_heartbeat=True):
+  def set_safety_mode(self, mode=SAFETY_SILENT, disable_heartbeat=True, disable_powersave=True):
     self._handle.controlWrite(Panda.REQUEST_OUT, 0xdc, mode, 0, b'')
     if disable_heartbeat:
       self.set_heartbeat_disabled()
+    if disable_powersave:
```
adeebshihadeh, line 460: I think this can use the same flag as the heartbeat disable

python/__init__.py
```diff
@@ -453,10 +453,12 @@ def esp_reset(self, bootmode=0):
...
+      self.set_power_save(0)
```
adeebshihadeh, line 461:
```suggestion
  def set_safety_mode(self, mode=SAFETY_SILENT, disable_checks=True):
    self._handle.controlWrite(Panda.REQUEST_OUT, 0xdc, mode, 0, b'')
    if disable_checks:
      self.set_heartbeat_disabled()
      self.set_power_save(0)
```

PR #729, opened 2021-09-25
Fix bulk transfer dropping messages sometimes

gregjhogan:
I noticed TX messages getting dropped when sending large amounts of CAN messages really fast with a timeout specified (and USB NAKs are needed).  I am sure I have the max number of messages per USB packet correct, so somehow it seems like a second packet sneaks in when we want to start NAKing but I am failing to see how.  I tried adding `ENTER_CRITICAL()` and `EXIT_CRITICAL()` around `usb_irqhandler` but it doesn't help, so I am not sure how this happens.  Since I don't understand what is going on, I am worried that maybe more than one extra USB packets could somehow sneak in and this isn't a real fix.

### DIFF
--- a/board/drivers/usb.h
+++ b/board/drivers/usb.h
@@ -23,8 +23,6 @@ typedef union _USB_Setup {
 }
 USB_Setup_TypeDef;
 
-#define MAX_CAN_MSGS_PER_BULK_TRANSFER 16U
-
 bool usb_eopf_detected = false;
 
 void usb_init(void);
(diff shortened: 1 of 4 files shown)


robbederks:
Have been trying to reproduce this, but must not fully understand what is happening. Added the `tests/message_drop_test.py` script, but this seems fine even on master.
How do you reproduce this?

robbederks:
Edited the mentioned script to use the timeout, and put it in a while loop. Still no dropped messages, but noticed something else: if it times out, part of the block of messages will still get sent. So if you try to send too many messages with a too short timeout, eventually it will keep repeating the same part of a block and never complete.

This still doesn't result in dropped messages for me though

robbederks:
Updated the test to only re-send the messages for which it timed out, which solved the problem mentioned in my comment above. Will also push this to master.

Can't reproduce the dropped messages with this on `master` though

gregjhogan:
I reproduce it reliably with a script repeatedly sending 256 to 512 messages at once to a single CAN bus in a tight loop.  I saw in the serial console the can_push errors and verified the tx ring buffer is short up to 4 slots by adding additional prints to the serial console.

gregjhogan:
Reproducing this does require sending large amounts of data really fast on a single bus.  Since we can't find any way to control this behavior and after a lot of testing this reliably fixes the issue, I am merging this.

PR #732, opened 2021-10-02
Build panda for both target mcpus

briskspirit:
SCons builds panda firmware for both H7 and F4 (except EON)


pd0wm:
Confirmed the sha hash of the final image is identical

pd0wm (approved):

PR #739, opened 2021-10-12
CAN_FIFOMailBox to CANPacket struct + USB dynamic packet size

briskspirit:
Remove STM's CAN_FIFOMailBox_TypeDef and replace with custom struct CANPacket_t with variable data length for future CAN FD support.


tests/safety/common.py
```diff
@@ -18,16 +17,12 @@ class UNSAFE_MODE:
...
-  else:
-    ret[0].RIR = (addr << 21) | 1
-  ret[0].RDTR = len(dat) | ((bus & 0xF) << 4)
-  ret[0].RDHR = rdhr
-  ret[0].RDLR = rdlr
+  ret = libpandasafety_py.ffi.new('CANPacket_t *')
+  ret[0].extended = 1 if addr >= 0x800 else 0
+  ret[0].addr = addr
+  ret[0].len = len(dat)
+  ret[0].bus = 3 if bus > 3 else bus
```
briskspirit, line 24: @robbederks and @adeebshihadeh , any ideas why/how I am getting weird bus numbers on input to package_can_msg function? It is 832 on Jenkins machine and 579 on my home workstation every run. Of course I am getting overflow as bus now limited to 2 bits. With the same func everything is fine on safety replay.

briskspirit:
Next I am going to check difference in speed between 13 and 16 bytes packets.

briskspirit:
USB refactoring saved space in data payload as previously we had exactly 16 bytes CAN messages(with 64 bytes USB packet size) and now we have variable size CAN messages(5 - 13 bytes for CAN 2.0 and 5 - 72 bytes for CAN-FD)
Also adjustment was made to USB transfer size and packet count, so interrupt being fired not after each 64 byte frame, but when transfer was finished or transfer size/packet count reached threshold.
USB through output tests showed:
C++  ~8000Kbit/s
Python  ~5900Kbit/s (python wastes time to pack/unpack data, tested with real data and CAN-FD enabled at 6Mbit/s x 3 buses)

Before adjustment to USB transfer size and packet count USB through output was maxed out at ~4300Kbit/s with both C++ and Python tests.

Counter was implemented for each USB packet. That's how we can control beginning of each message and that no messages were lost in the middle of transfer(Protection from sending junk on CAN buses).

EP1 and EP3 (USB IN and OUT) max transfer sizes were adjusted by conducting various tests. Think I've found the best values.

TX ring buffers were increased in size from 256 to 416 as now we need to hold more slots empty for each USB transfer. By holding more slots empty we are not sacrificing performance, no regressions.

briskspirit:
Maybe going to remove bitstruct library from python as C implementation lacks support of byte order and python implementation is slow.

tests/safety/test_volkswagen_mqb.py
```diff
@@ -225,7 +225,10 @@ def test_rx_hook(self):
       if msg == MSG_MOTOR_20:
         to_push = self._gas_msg(0)
       self.assertTrue(self._rx(to_push))
-      to_push[0].RDHR ^= 0xFF
```
robbederks, line 228: This isn't equivalent, should be only on the LSB, no?
```
to_push[0].data[4] ^= 0xFF
```

tests/safety/test_volkswagen_pq.py
```diff
@@ -60,15 +60,16 @@ def _set_prev_torque(self, t):
   def _speed_msg(self, speed):
```
robbederks, line 60: Not sure why PQ doesn't use the DBC to pack the values like other brands?

tests/safety_replay/helpers.py
```diff
@@ -58,17 +57,21 @@ def set_desired_torque_last(safety, mode, torque):
   elif mode == Panda.SAFETY_SUBARU:
     safety.set_subaru_desired_torque_last(torque)
 
-def package_can_msg(msg):
-  rdlr, rdhr = struct.unpack('II', msg.dat.ljust(8, b'\x00'))
-
-  ret = libpandasafety_py.ffi.new('CAN_FIFOMailBox_TypeDef *')
-  if msg.address >= 0x800:
-    ret[0].RIR = (msg.address << 3) | 5
+def len_to_dlc(length):
```
robbederks, line 60: why not use the `DLC_TO_LEN` LUT you made in `__init__.py`?

python/__init__.py
```diff
@@ -511,35 +536,58 @@ def set_uart_callback(self, uart, install):
   # Timeout is in ms. If set to 0, the timeout is infinite.
   CAN_SEND_TIMEOUT_MS = 10
 
-  def can_send_many(self, arr, timeout=CAN_SEND_TIMEOUT_MS):
-    snds = []
-    transmit = 1
-    extended = 4
-    for addr, _, dat, bus in arr:
+  def len_to_dlc(self, length):
```
robbederks, line 539: Let's just do a lookup in the `DLC_TO_LEN`

tests/safety_replay/helpers.py
robbederks: and maybe the packaging of the msg should also just use a helper function in `__init__.py` to avoid redundance.
Can also add a unit test to make sure `unpack(pack(msg)) == msg`

tests/safety/test.c
```diff
@@ -66,14 +63,20 @@ void fault_occurred(uint32_t fault) {
...
-// from llcan.h
-#define GET_BUS(msg) (((msg)->RDTR >> 4) & 0xFF)
-#define GET_LEN(msg) ((msg)->RDTR & 0xf)
-#define GET_ADDR(msg) ((((msg)->RIR & 4) != 0) ? ((msg)->RIR >> 3) : ((msg)->RIR >> 21))
-#define GET_BYTE(msg, b) (((int)(b) > 3) ? (((msg)->RDHR >> (8U * ((unsigned int)(b) % 4U))) & 0XFFU) : (((msg)->RDLR >> (8U * (unsigned int)(b))) & 0xFFU))
-#define GET_BYTES_04(msg) ((msg)->RDLR)
-#define GET_BYTES_48(msg) ((msg)->RDHR)
-#define GET_FLAG(value, mask) (((__typeof__(mask))param & mask) == mask)
+// from config.h
+unsigned char dlc_to_len[] = {0U, 1U, 2U, 3U, 4U, 5U, 6U, 7U, 8U, 12U, 16U, 20U, 24U, 32U, 48U, 64U};
```
robbederks, line 67: instead of redefining the struct and these helpers, can you just include the header file straight from the board code?

tests/safety/common.py
```diff
@@ -16,18 +15,22 @@ class UNSAFE_MODE:
   DISABLE_STOCK_AEB = 2
   RAISE_LONGITUDINAL_LIMITS_TO_ISO_MAX = 8
 
+def len_to_dlc(length):
```
robbederks, line 18: same comments as in `helpers.py`. let's get rid of these (re)definitions and define the helpers for this in `__init__.py`

tests/misra/suppressions.txt
```diff
@@ -9,4 +9,8 @@ misra-c2012-19.2
 # Advisory: The # and ## preprocessor operators should not be used
 misra-c2012-20.10
 # Required: it's ok re-defining potentially reserved Macro names. Not likely to cause confusion
-misra-c2012-21.1
\ No newline at end of file
+misra-c2012-21.1
+# MUST BE FIXED, mostly in the safety code
+misra-c2012-10.4
```
robbederks, line 14: can we fix these instead of ignoring them?

tests/gmbitbang/test_packer.c
```diff
@@ -1,21 +1,26 @@
...
 #include <stdint.h>
 
-typedef struct {
-  uint32_t RIR;  /*!< CAN receive FIFO mailbox identifier register */
-  uint32_t RDTR; /*!< CAN receive FIFO mailbox data length control and time stamp register */
-  uint32_t RDLR; /*!< CAN receive FIFO mailbox data low register */
-  uint32_t RDHR; /*!< CAN receive FIFO mailbox data high register */
-} CAN_FIFOMailBox_TypeDef;
+#define CANPACKET_DATA_SIZE_MAX 8
+typedef struct __attribute__((packed)) {
```
robbederks, line 5: should be able to just include a header file with these definitions from the board code

setup.py
```diff
@@ -50,7 +50,8 @@ def find_version(*file_paths):
     'hexdump >= 3.3',
     'pycryptodome >= 3.9.8',
     'tqdm >= 4.14.0',
-    'requests'
+    'requests',
+    'bitstruct'
```
robbederks, line 54: let's get rid of `bitstruct`

python/__init__.py
```diff
@@ -511,35 +536,58 @@ def set_uart_callback(self, uart, install):
...
+    if length <=8:
+      return length
+    if length <=24:
+      return 8 + ((length - 8) // 4) + (1 if length % 4 else 0)
+    else:
+      return 11 + (length // 16) + (1 if length % 16 else 0)
+
+  def pack_can_buffer(self, arr):
```
robbederks, line 547: as mentioned in other comments, let's make a pack and unpack function, add a unit test for these and have this function just do a `list(map(pack, msgs))` or something like that

board/main.c
```diff
@@ -218,26 +273,58 @@ void usb_cb_ep2_out(void *usbdata, int len, bool hardwired) {
...
+    ep3_buffer.tail_size = 0;
+  }
+  // Assembling can message with data from buffer
+  if (usbdata8[0] == ep3_buffer.counter) {
+    uint8_t pos = 1;
+    ep3_buffer.counter++;
+    if (ep3_buffer.ptr != 0) {
+      if (ep3_buffer.tail_size <= 63U) {
+        CANPacket_t to_push;
+        (void)memcpy(&ep3_buffer.data[ep3_buffer.ptr], &usbdata8[pos], ep3_buffer.tail_size);
```
pd0wm, line 295: Since we use a lot of `memcpy` now, it might be worth it to merge (or write a new implementation) https://github.com/commaai/panda/pull/197/files.

board/drivers/bxcan.h
```diff
@@ -161,23 +174,35 @@ void can_rx(uint8_t can_number) {
...
-    to_push.RDTR = (to_push.RDTR & 0xFFFF000F) | (bus_number << 4);
+    CANPacket_t to_push;
+
+    to_push.returned = 0U;
+    to_push.rejected = 0U;
+    to_push.extended = (CAN->sFIFOMailBox[0].RIR >> 2) & 0x1U;
+    to_push.addr = (to_push.extended != 0) ? (CAN->sFIFOMailBox[0].RIR >> 3) : (CAN->sFIFOMailBox[0].RIR >> 21);
+    to_push.data_len_code = CAN->sFIFOMailBox[0].RDTR & 0xFU;
+    to_push.bus = bus_number;
+    to_push.data[0] = CAN->sFIFOMailBox[0].RDLR & 0xFFU;
```
robbederks, line 185: see above

python/__init__.py
```diff
@@ -15,26 +15,51 @@
...
 
-__version__ = '0.0.9'
+__version__ = '0.0.10'
 
 BASEDIR = os.path.join(os.path.dirname(os.path.realpath(__file__)), "../")
 
 DEBUG = os.getenv("PANDADEBUG") is not None
 
+CANPACKET_HEAD_SIZE = 0x5
+DLC_TO_LEN = [0, 1, 2, 3, 4, 5, 6, 7, 8, 12, 16, 20, 24, 32, 48, 64]
```
pd0wm, line 25:
```suggestion
DLC_TO_LEN = [0, 1, 2, 3, 4, 5, 6, 7, 8, 12, 16, 20, 24, 32, 48, 64]
LEN_TO_DLC = {length: dlc for (dlc, length) in enumerate(DLC_TO_LEN)}
```

board/drivers/bxcan.h
```diff
@@ -161,23 +174,35 @@ void can_rx(uint8_t can_number) {
...
-      to_send.RDHR = to_push.RDHR;
+      CANPacket_t to_send;
+
+      to_send.returned = 0U;
+      to_push.rejected = 0U;
+      to_send.extended = to_push.extended; // TXRQ
+      to_send.addr = to_push.addr;
+      to_send.bus = to_push.bus;
+      to_send.data_len_code = to_push.data_len_code;
+      (void)memcpy(to_send.data, to_push.data, dlc_to_len[to_push.data_len_code]);
```
robbederks, line 205: do we sanity check the `data_len_code` somewhere? what if we send a packet where the DLC is larger than the actual data buffer we sent?

tests/safety/test_volkswagen_pq.py
jyoung8607: PQ safety was upstreamed before the tests were converted to use the packers, and at the time they were converted, important parts of the matching DBC weren't yet upstream. It's been on my Round Tuit list to deal with, but fairly far down. If it's a blocker for your work, I can fix it in the next day or two.

python/__init__.py
pd0wm: Not sure what you want to do with unsupported lengths. Now they will just fail on lookup.

board/drivers/fdcan.h
```diff
@@ -42,36 +47,41 @@ void process_can(uint8_t can_number) {
...
+        canfd_fifo *fifo;
+        fifo = (canfd_fifo *)(TxFIFOSA + (tx_index * FDCAN_TX_FIFO_EL_SIZE));
+
+        fifo->header[0] = (to_send.extended << 30) | ((to_send.extended != 0) ? (to_send.addr) : (to_send.addr << 18));
+        fifo->header[1] = (to_send.data_len_code << 16); // | (1U << 21) | (1U << 20) to enable CAN FD , enable BRS
+
+        uint8_t data_len_w = (dlc_to_len[to_send.data_len_code] / 4U);
+        data_len_w += ((dlc_to_len[to_send.data_len_code] % 4) > 0) ? 1U : 0U;
+        for (unsigned int i = 0; i < data_len_w; i++) {
+          fifo->data_word[i] = to_send.data[(i*4)+0] | (to_send.data[(i*4)+1] << 8) | (to_send.data[(i*4)+2] << 16) | (to_send.data[(i*4)+3] << 24);
```
robbederks, line 69: same as in other places, would be faster with some casting magic I think

board/drivers/bxcan.h
```diff
@@ -101,18 +101,29 @@ void process_can(uint8_t can_number) {
...
-          to_push.RDLR = CAN->sTxMailBox[0].TDLR;
-          to_push.RDHR = CAN->sTxMailBox[0].TDHR;
+          CANPacket_t to_push;
+          to_push.returned = 1U;
+          to_push.rejected = 0U;
+          to_push.extended = (CAN->sTxMailBox[0].TIR >> 2) & 0x1U;
+          to_push.addr = (to_push.extended != 0) ? (CAN->sTxMailBox[0].TIR >> 3) : (CAN->sTxMailBox[0].TIR >> 21);
+          to_push.data_len_code = CAN->sTxMailBox[0].TDTR & 0xFU;
+          to_push.bus = bus_number;
+          to_push.data[0] = CAN->sTxMailBox[0].TDLR & 0xFFU;
```
robbederks, line 118: let's maybe make a macro to do these conversions to/from a uint8_t array

board/main.c
```diff
@@ -192,14 +192,69 @@ int get_rtc_pkt(void *dat) {
...
+    }
+    else {
+      pos = 0;
+      ep1_buffer.counter = 0;
+    }
+    if (pos != len) {
+      ep1_buffer.counter = 0;
+      total_rx_size = 0;
+    }
+    total_rx_size += pos;
```
robbederks, line 255: let's add some comments here, kinda hard to follow the logic like this

robbederks: can't easily tell what this does or what it's supposed to do. Just stitch together a single message larger than the 64 byte frame?

robbederks: also: do we want to compile this out and add some unit tests to this to make sure it's good in all edge cases?

board/main.c
```diff
@@ -218,26 +273,58 @@ void usb_cb_ep2_out(void *usbdata, int len, bool hardwired) {
...
+        pos += ep3_buffer.ptr;
+      }
+    }
   }
 }
 
 void usb_cb_ep3_out_complete(void) {
-  // TODO: how does a second USB packet sneek in? (why multiply by 2)
-  if (can_tx_check_min_slots_free(MAX_CAN_MSGS_PER_BULK_TRANSFER * 2U)) {
+  if (can_tx_check_min_slots_free(MAX_CAN_MSGS_PER_BULK_TRANSFER)) {
```
robbederks, line 327: this is fixed too?

board/main_declarations.h
```diff
@@ -8,7 +8,16 @@ typedef struct harness_configuration harness_configuration;
...
-
+typedef struct {
+  unsigned char reserved : 1;
+  unsigned char bus : 3;
+  unsigned char data_len_code : 4;
+  unsigned char rejected : 1;
+  unsigned char returned : 1;
+  unsigned char extended : 1;
+  unsigned int addr : 29;
+  unsigned char data[CANPACKET_DATA_SIZE_MAX];
```
robbederks, line 19: any reason to use `unsigned char` instead of `bool` for the 1-bit fields?

board/main.c
pd0wm: Big fan of putting all the ring buffer code in a tested library!

tests/safety/test_volkswagen_pq.py
jyoung8607: Agreed, will do.

robbederks: Opened an issue: #762

board/main_declarations.h
briskspirit: MISRA violation

board/main.c
briskspirit: Still testing, will ask @greg also to check, but should be fixed

setup.py
briskspirit: ops, missed that one...

tests/safety/test_volkswagen_mqb.py
briskspirit: yes, it's a bug, fixed

tests/safety_replay/helpers.py
briskspirit: It has different logic, but discussed with you already, will change this approach

tests/misra/suppressions.txt
briskspirit: Will fix in follow up PR as those violations were always there but were hidden from MISRA by volatile type. If I am going to fix this in this PR - will add ~150-170 lines of  code. And can't fix before as it is hidden.
Also all those violations are in safety code.

pd0wm:
A few more comments I thought of over the weekend:
 - Can you document the USB packet somewhere? Would be super nice if you can draw a small diagram how the data is laid out in the bytes, like they do in datasheets for register values.
 - I assume this PR breaks the socketcan and Windows J2534 driver. I think we should at least have PRs ready to update those to the new packet format, otherwise I'm worried they will be broken for a while.
 - Can we detect (in a backwards compatible) way if we're dealing with a panda with the hold or the new format on initial connect? Would be nice if the python library tells you to flash if your panda is running a firmware that uses the old format.

briskspirit:
Added CANPacket and USB packet structure diagrams.
Will open PRs for socketcan and Windows driver.
Starting to work on versioning separate PR now.

adeebshihadeh:
Can you document the USB packet in text instead of an image? There should be nice tools to make it look nice in markdown.

briskspirit:
Will look into it later, fast search on how to convert table to markup didn't gave me acceptable results.

board/main.c
gregjhogan: It seems like since we set the USB RX FIFO size `GRXFSIZ` to 256 bytes (64 words) it can't possibly buffer more than that many bytes.  However, `MAX_CAN_MSGS_PER_BULK_TRANSFER` currently is 64 but you can only pack 51.2 messages (min message size is 5 bytes) into 256 bytes so the current value is probably higher than necessary.

board/drivers/usb.h
```diff
@@ -23,7 +23,8 @@ typedef union _USB_Setup {
 }
 USB_Setup_TypeDef;
 
-#define MAX_CAN_MSGS_PER_BULK_TRANSFER 4U
+#define MAX_CAN_MSGS_PER_BULK_TRANSFER 64U
```
gregjhogan, line 26: We should update this variable name, it is now the max number of CAN messages that fix in the RX FIFO buffer.  Perhaps we should define a constant for the RX FIFO size and then divide that by the min CAN message size here.

.github/workflows/test.yaml
```diff
@@ -41,6 +41,18 @@ jobs:
...
+    name: panda usb protocol
+    runs-on: ubuntu-20.04
+    timeout-minutes: 45
+    steps:
+      - uses: actions/checkout@v2
+      - name: Build Docker image
+        run: eval "$BUILD"
+      - name: Test pack/unpack for Panda usb protocol
+        run: $RUN "cd /tmp/openpilot/panda/tests/usbprotocol &&
+                   python -m unittest discover ."
```
adeebshihadeh, line 54:
```suggestion
  unit_tests:
    name: unit tests
    runs-on: ubuntu-20.04
    timeout-minutes: 45
    steps:
      - uses: actions/checkout@v2
      - name: Build Docker image
        run: eval "$BUILD"
      - name: Test pack/unpack for USB protocol
        run: $RUN "cd /tmp/openpilot/panda/tests/usbprotocol &&
                   python -m unittest discover ."
```

tests/usbprotocol/test_pandalib.py
```diff
@@ -0,0 +1,24 @@
...
+      data = bytes([random.getrandbits(8) for _ in range(random.randrange(1, 9))])
+      to_pack.append((address, 0, data, 0))
+
+    packed = pack_can_buffer(to_pack)
+    unpacked = []
+    for dat in packed:
+      unpacked.extend(unpack_can_buffer(dat))
+
+    assert unpacked == to_pack
+    print("Successfully packed and unpacked 10000 messages")
```
adeebshihadeh, line 21: not great to spam the output in unit tests. think about the case where we have dozens of tests running, like in openpilot. you only want to see the relevant stuff, i.e. the failing tests' error messages

tests/safety/test.c
```diff
@@ -2,21 +2,8 @@
...
-} CAN_TxMailBox_TypeDef;
-
-typedef struct
-{
-  uint32_t RIR;  /*!< CAN receive FIFO mailbox identifier register */
-  uint32_t RDTR; /*!< CAN receive FIFO mailbox data length control and time stamp register */
-  uint32_t RDLR; /*!< CAN receive FIFO mailbox data low register */
-  uint32_t RDHR; /*!< CAN receive FIFO mailbox data high register */
-} CAN_FIFOMailBox_TypeDef;
+#define CANPACKET_DATA_SIZE_MAX 8
```
adeebshihadeh, line 5: does it make more sense to define these in the bxcan and fdcan drivers?

PR #740, opened 2021-10-15
Fix bug in safety tests

briskspirit:


robbederks:
@adeebshihadeh why do we even have this check? It obviously fails for Honda Bosch and Nidec, and it might also be very dependent on the actual contents of the messages.
It's introduced in PR #521

tests/safety/common.py
```diff
@@ -414,7 +414,7 @@ def test_sample_speed(self):
     self.safety.safety_rx_hook(self._speed_msg(self.STANDSTILL_THRESHOLD + 1))
     self.assertTrue(self.safety.get_vehicle_moving())
 
-  def test_tx_hook_on_wrong_safety_mode(self):
+def test_tx_hook_on_wrong_safety_mode(self):
```
robbederks, line 417: mistake?

tests/safety/common.py
```diff
@@ -427,14 +427,17 @@ def test_tx_hook_on_wrong_safety_mode(self):
         if attr.startswith("Test") and attr != current_test:
           tx = getattr(getattr(test, attr), "TX_MSGS")
           if tx is not None:
+            # TODO: Temporary, should be fixed in panda firmware, safety_honda.h
+            if attr in ['TestHondaBoschLongGiraffeSafety', 'TestHondaNidecSafety']:
+              tx = [x for x in tx if x[0] != 0x1FA and x[0] != 0x30C ]
```
robbederks, line 432:
```suggestion
              tx = list(filter(lambda m: m[0] not in [0x1FA, 0x30C], tx))
```

PR #742, opened 2021-10-18
expose TX messages blocked by safety hook

gregjhogan:
I have had errors briefly flash on my dash from time to time and I am curious if it is the panda blocking messages because it notices the gas pedal press slightly before openpilot disengages.  This should make it a lot easier to investigate (separating TX ring buffer overflow from safety hook violations).

It may also make it possible to prevent gaps in counters in similar situations on vehicles where this causes more permanent faults.

### DIFF
--- a/tests/automated/3_usb_to_can.py
+++ b/tests/automated/3_usb_to_can.py
@@ -59,9 +59,9 @@ def test_safety_nooutput(p):
   # confirm receive nothing
   time.sleep(0.05)
   r = p.can_recv()
-  # bus 255 is messages blocked by TX safety hook
-  assert len([x for x in r if x[3] != 255]) == 0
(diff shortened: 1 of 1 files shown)


pd0wm:
Think it's fine to add a new field in the capnp struct for this use case. For communication between panda and openpilot we don't have to care about backwards compatibility.

briskspirit:
@greg then we need to discuss CANPacket again as your PR won't be preserved in the current state after my PR. Do you really want to add this feature as MSB to bus number or maybe move to separate bit in packed struct? Whole idea of Mailbox refactoring was to move to struct and omit bitwise operations where we don't need them.

gregjhogan:
Sounds good, so it should be fine to merge this PR as it is now and enhance your PR as you suggested to preserve the functionality?

PR #760, opened 2021-10-31
Universal Measurement and Calibration Protocol (XCP) python library

gregjhogan:


python/xcp.py
```diff
@@ -0,0 +1,178 @@
...
+      "stim_support": resp[0] & 0x08 != 0,
+      "pgm_support": resp[0] & 0x10 != 0,
+      "byte_order": self._byte_order,
+      "address_granularity": 2**((resp[1] & 0x06) >> 1),
+      "slave_block_mode": resp[1] & 0x40 != 0,
+      "optional": resp[1] & 0x80 != 0,
+      "max_cto": resp[2],
+      "max_dto": struct.unpack(f"{self._byte_order}H", resp[3:5])[0],
+      "protocol_version": resp[5],
+      "transprot_version": resp[6],
```
pd0wm, line 164:
```suggestion
      "transport_version": resp[6],
```

pd0wm:
https://github.com/vectorgrp/XCPlite/blob/087023732ba572d913309c887ab63224e3335c81/src/xcp.h#L448 is a good reference for the packet layout

adeebshihadeh:
Let's just merge this if it works?

PR #763, opened 2021-11-02
memcpy with 4 bytes support

briskspirit:
Tested on my refactoring branch, works and adds ~5% to through output so I would like to merge it asap if nobody minds.

### DIFF
--- a/board/libc.h
+++ b/board/libc.h
@@ -14,13 +14,22 @@ void *memset(void *str, int c, unsigned int n) {
   return str;
 }
 
+
 void *memcpy(void *dest, const void *src, unsigned int n) {
-  uint8_t *d = dest;
-  const uint8_t *s = src;
-  for (unsigned int i = 0; i < n; i++) {
-    *d = *s;
-    d++;
-    s++;
+  const uint32_t *s32 = (const uint32_t *)src;
+  uint32_t *d32 = (uint32_t *)dest;
+  for (unsigned int i = (n / 4U); i > 0; i--) {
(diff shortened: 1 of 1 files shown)


briskspirit:
master branch also improved

pd0wm:
How does https://sourceware.org/git/?p=newlib-cygwin.git;a=blob;f=newlib/libc/string/memcpy.c;h=52f716b9275f5d24cedb7d66c41541945d13bfb6;hb=HEAD#l49 perform?

This implementation is probably pretty bad if src or dst is not aligned.

briskspirit:
@pd0wm checked link #1 (adjusted threshold to 4 bytes) - with alignment check it is the same as our old implementation, without check - similar to my previous. Looks like compiler or MCU handles unaligned addresses better, than if we just copy byte by byte.

pd0wm:
Why are we copying unaligned buffers? That should be an edge case.

briskspirit:
struct has 5 bytes for header and rest for data. data won't be aligned (and struct explicitly aligned to 1 byte)
also when assembling usb packets - they get misaligned too I suppose. So seems in our case it is not really an edge case?
P.S.: that's all related only to my new branch.

briskspirit:
Ok, mystery solved after I moved struct alignment from 1 byte to 4 bytes. (Also seems like there is no reason to align it to 1 byte in my case + saved ~600bytes of binary size). Now memcpy with alignment check performs as fast as without one. Commited final version to this PR.

PR #764, opened 2021-11-03
VW PQ: Refactor tests to use CAN packer

### LINKED ISSUE #762: Refactor VW PQ safety tests to use DBC packer
Right now it uses similar bitwise logic to generate the CAN packets in the unit tests. This is problematic since it's easy to introduce the same bug on both sides and have passing tests.

jyoung8607:
Resolves #762.

- [x] Need comma to review and merge commaai/opendbc#463
- [x] Need comma to bump the openpilot submodule ref so I can bump the Panda Dockerfile ref in this PR
- [x] Verify CI (passing for me locally, but need the above for GitHub CI)

### DIFF
--- a/Dockerfile.panda
+++ b/Dockerfile.panda
@@ -60,7 +60,7 @@ RUN cd /tmp && \
     git clone https://github.com/commaai/openpilot.git tmppilot || true && \
     cd /tmp/tmppilot && \
     git fetch && \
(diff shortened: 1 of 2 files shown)


adeebshihadeh:
Merged and bumped

jyoung8607:
Ready for review.

At some point I'll file a follow-up PR to handle the checksums and counters in duties in CPP.

PR #770, opened 2021-11-09
Check panda library and firmware for compatibility

briskspirit:
Adds versioning to health packet structure and CAN/USB packet structure.

### DIFF
--- a/board/config.h
+++ b/board/config.h
@@ -34,6 +34,8 @@
 
 #define MAX_RESP_LEN 0x40U
 
+#define CAN_PACKET_VERSION 1
+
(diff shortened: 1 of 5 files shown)


briskspirit:
Raising exception is not very good idea as then the only way to update firmware is to use DFU and dfu-util.
As an option we can block getting health packet and can_send_many/can_recv if lib and firmware are not compatible.
Not sure if we should, warning should be enough to understand what is happening and what to do.

pd0wm:
I would use `@functools.wraps` to clean up the decorators, and have separate decorators for `@ensure_health_packet_version` and `@ensure_can_packet_version`

pd0wm (approved):

PR #772, opened 2021-11-10
Enable CAN FD support

briskspirit:
Enable CAN FD support.
CAN FD enabled if can_data_speed is set. Also BRS enabled if can_data_speed is higher than can_speed.
can_data_speed must be set after can_speed (and that's seems logical)

P.S.: Based on #739 , testing.

### DIFF
--- a/board/drivers/can_common.h
+++ b/board/drivers/can_common.h
@@ -165,9 +165,9 @@ void can_clear(can_ring *q) {
 // Helpers
 // Panda:       Bus 0=CAN1   Bus 1=CAN2   Bus 2=CAN3
 bus_config_t bus_config[] = {
-  { .bus_lookup =0U, .can_num_lookup = 0U, .can_speed = 5000U, .can_data_speed = 5000U, .canfd_enabled = false, .brs_enabled = false },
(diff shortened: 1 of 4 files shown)


adeebshihadeh:
FYI, you can set the base branch to #739 for a cleaner history. GitHub should automatically change the base to master once the first PR is merged.

board/drivers/can_common.h
```diff
@@ -160,7 +160,11 @@ void can_clear(can_ring *q) {
 uint8_t bus_lookup[] = {0,1,2};
 uint8_t can_num_lookup[] = {0,1,2,-1};
 uint32_t can_speed[] = {5000, 5000, 5000, 333};
-uint32_t can_data_speed[] = {5000, 5000, 5000}; //For CAN FD with BRS only
+// For CAN FD only
+uint32_t can_data_speed[] = {5000, 5000, 5000};
+bool canfd_enabled[] = {false, false, false};
+bool brs_enabled[] = {false, false, false};
```
robbederks, line 166: maybe we can refactor this a bit to a struct with the bus config and just have an array of that?

PR #775, opened 2021-11-12
Improve tests

briskspirit:
Recover test must go before flash test for better stability(to flash main and bootstub firmwares first)
Also 750kbps speed is not used by CAN/CAN FD and prevents to reach both 2 and 5 Mbps with CAN FD because of time quanta.


adeebshihadeh:
Can you use a more descriptive title and commit message next time? Something like "tests/automated: test recover before flash and remove 750kbps tests"

PR #779, opened 2021-11-15
toyota: exit controls with "main on" button

ntegan1:
`controls_allowed` should always be 0 after a "main on" button press.

"I mean there's 0 debate about main being a full disengage"

related to https://github.com/commaai/panda/pull/630

### DIFF
--- a/board/safety/safety_toyota.h
+++ b/board/safety/safety_toyota.h
@@ -35,6 +35,7 @@ AddrCheckStruct toyota_addr_checks[] = {
   {.msg = {{ 0xaa, 0, 8, .check_checksum = false, .expected_timestep = 12000U}, { 0 }, { 0 }}},
   {.msg = {{0x260, 0, 8, .check_checksum = true, .expected_timestep = 20000U}, { 0 }, { 0 }}},
   {.msg = {{0x1D2, 0, 8, .check_checksum = true, .expected_timestep = 30000U}, { 0 }, { 0 }}},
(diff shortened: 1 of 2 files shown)


spektor56:
if you are going to use main_on as the variable name it should be a boolean value that tracks the actual state (true if > 0 else false).  Since you have it as an int value here which doesn't track the actual state it should be named something like, main_state and main_state_prev

ntegan1:
> if you are going to use main_on as the variable name it should be a boolean value that tracks the actual state (true if > 0 else false). Since you have it as an int value here which doesn't track the actual state it should be named something like, main_state and main_state_prev. The only possible values are 0 and 1 for this field though so having it as a boolean would make it more clear to the programmer it only has 2 possible states.

I incorrectly mimicked the `cruise_engaged` example above. Didn't see they were implicitly casting int to bool.

ntegan1:
Misra should be fixed

adeebshihadeh:
This should already be the case, since openpilot's state is tied to the PCM's state. Does controls allowed stay true when disengaging with the main button on your car?

spektor56:
> This should already be the case, since openpilot's state is tied to the PCM's state. Does controls allowed stay true when disengaging with the main button on your car?

This is just a more correct way of doing it, the controls_allowed state doesn't depend on the ACC state this way.  For independent LKAS / ACC for example, you cant tie controls_allowed to the ACC state like panda is currently doing.  ACC off with main on is a possible state that should allow LKAS to be active.

If anything this code just adds some redundancy to your panda code to make sure it disables when the main flag changes.

PR #780, opened 2021-11-15
minor typo in bxcan.h and fdcan.h

briskspirit:


adeebshihadeh:
Can we write tests for this?

briskspirit:
This typo was not affecting anything as this flag is not used while message is being sent, so can't test it.

PR #782, opened 2021-11-15
Add passthrough safety mode

briskspirit:
This safety mode opens relay and forwards messages from bus 0 to 2 and from bus 2 to 0. That way we can know for sure which messages were sent by camera. 
It is only allowed when compiled with DEBUG set to true.

### DIFF
--- a/board/safety/safety_defaults.h
+++ b/board/safety/safety_defaults.h
@@ -45,8 +45,13 @@ const safety_hooks nooutput_hooks = {
 
 // *** all output safety mode ***
 
+// Enables passthrough mode where relay is open and bus 0 gets forwarded to bus 2 and vice versa
+const uint16_t ALLOUTPUT_PARAM_PASSTHROUGH = 1;
+bool alloutput_passthrough = false;
+
 static const addr_checks* alloutput_init(int16_t param) {
   UNUSED(param);
+  alloutput_passthrough = GET_FLAG(param, ALLOUTPUT_PARAM_PASSTHROUGH);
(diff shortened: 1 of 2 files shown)


board/safety/safety_defaults.h
```diff
@@ -71,3 +71,27 @@ const safety_hooks alloutput_hooks = {
...
+
+  return bus_fwd;
+}
+
+const safety_hooks passthrough_hooks = {
+  .init = alloutput_init,
+  .rx = default_rx_hook,
+  .tx = alloutput_tx_hook,
+  .tx_lin = alloutput_tx_lin_hook,
+  .fwd = passthrough_fwd_hook,
```
adeebshihadeh, line 96: if this uses the all output hooks, maybe just make this a flag in all output?

briskspirit: I would like to keep it as a separate safety mode to not confuse anybody.

board/safety/safety_defaults.h
```diff
@@ -64,10 +69,26 @@ static int alloutput_tx_lin_hook(int lin_num, uint8_t *data, int len) {
   return true;
 }
 
+static int alloutput_fwd_hook(int bus_num, CANPacket_t *to_fwd) {
+  UNUSED(to_fwd);
+  int bus_fwd = -1;
+
+  if (alloutput_passthrough == true) {
```
adeebshihadeh, line 76:
```suggestion
  if (alloutput_passthrough) {
```

PR #794, opened 2021-11-24
Subaru SNG

martinl:
This PR is prerequisite for adding Subaru SNG support for both global and preglobal models.

Subaru stock ACC stops behind stopped cars but does not resume automatically. We send Throttle:Throttle_Pedal signal from openpilot to ACC when car in front starts moving to resume from ACC Hold state (mimicking stock ACC resume using gas press behaviour). 

We also tried to resume using ACC resume button signals but that did not work since ACC buttons are directly wired to Eyesight and not accepted as input from CAN.


simonsickle (approved):

martinl:
I opened a new PR https://github.com/commaai/panda/pull/1379 closing this one

PR #808, opened 2021-12-16
Move shared definitions into separate file

briskspirit:
Those that might be needed by openpilot


board/global_definitions.h
```diff
@@ -0,0 +1,26 @@
+// USB definitions
+#define USB_VID 0xBBAAU
+
```
adeebshihadeh, line 3: i'd call this file like `panda.h` or `config.h`

PR #809, opened 2021-12-20
Mazda: Fix CX-5 ignition

Jafaral:
Signed-off-by: Jafar Al-Gharaibeh <to.jafar@gmail.com>

### DIFF
--- a/board/drivers/can_common.h
+++ b/board/drivers/can_common.h
@@ -216,7 +216,7 @@ void ignition_can_hook(CANPacket_t *to_push) {
 
     // Mazda exception
     if ((addr == 0x9E) && (len == 8)) {
-      ignition_can = (GET_BYTE(to_push, 0) >> 4) == 0xDU;
+      ignition_can = (GET_BYTE(to_push, 0) >> 5) == 0x6U;
     }
 
   }


Jafaral:
Now that we allow 0 and 1  for the LSB, maybe we should just shift  right by an additional bit and test for value `0x6U`. thoughts @adeebshihadeh  ?

adeebshihadeh:
Sounds good to me. For CAN ignition cars, we want to make sure the ignition is unique to the brand (to prevent collisions) and generic across the brand. Right now, there's a low cost to changing the signal since there are relatively few users, but it gets harder as the brand grows. I'd like to get this right while the cost is low.

Do we know what this signal is? Can we identify other potential signals that might work?

Jafaral:
I don't know what the message is, but All  supported Mazda cars has it. The MSB is set when ignition is on in CX-5 and CX-9 at least. I did see data from Mazda 3 and it seems to have the same behavior. but the CAN logs are a bit late to see if the bit was zero before ignition, at least from a couple of drives from cars that use the old ignition wiring.  In this PR we test 3 bits. I noticed the second byte is always zero, we can make that as an additional test to reduce the likelihood of collisions.

Jafaral:
@adeebshihadeh, this PR doesn't change the exiting logic, well,  beyond dropping a single bit from the test. This fixes ignition for CX-5 which happens to be the car with most users. Let us get this in for now, and I will collect more data over time on how to proceed if we decide to change.

PR #810, opened 2021-12-20
GM Safety: Switch to AcceleratorPedal2 message for pressed detection

JasonJShuler:
### Update pedal-pressed detection message

(Prep for Silverado / Stock VOACC)

* Switch to the AcceleratorPedal2 (452) message for pedal pressed detection, from 417.
  * The messages are redundant (452 contains an extra field).
  * Both were required in OP but 417 is not present on the 2020 Silverado.
  * Since both were required, it should be safe to assume AcceleratorPedal2 is present on all currently functional vehicles
  * While some sort of conditional may be possible, I see no reason to require both if they both have the same value

### DIFF
--- a/board/safety/safety_gm.h
+++ b/board/safety/safety_gm.h
@@ -18,7 +18,7 @@ const int GM_DRIVER_TORQUE_FACTOR = 4;
 const int GM_MAX_GAS = 3072;
 const int GM_MAX_REGEN = 1404;
 const int GM_MAX_BRAKE = 350;
-const CanMsg GM_TX_MSGS[] = {{384, 0, 4}, {1033, 0, 7}, {1034, 0, 7}, {715, 0, 8}, {880, 0, 6}, {789, 0, 5}, {800, 0, 6},   // pt bus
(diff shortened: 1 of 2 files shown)


adeebshihadeh:
This PR should be split up into two parts: new pedal pressed signal and the new actuation commands. The PR for the new actuation commands will need to include the appropriate checks in the TX hook, as well as unit tests to enforce those checks.

JasonJShuler:
> This PR should be split up into two parts: new pedal pressed signal and the new actuation commands. The PR for the new actuation commands will need to include the appropriate checks in the TX hook, as well as unit tests to enforce those checks.

Ok, this makes sense.

JasonJShuler:
> This PR should be split up into two parts: new pedal pressed signal and the new actuation commands. The PR for the new actuation commands will need to include the appropriate checks in the TX hook, as well as unit tests to enforce those checks.

This is complete

JasonJShuler:
Need to do more cleanup. Will resubmit when tests are passing

PR #812, opened 2021-12-21
Ford safety for LCA vehicles

incognitojam:
Implementing Ford safety for new(-er) cars with LCA option.

The structure of the code is based upon the VW safety code, and the Tesla/Nissan code (other angle based cars). I used the steering angle rate limit values from the Tesla. These values might need changing, but I would appreciate guidance on where to start with that.

## Related PRs
- https://github.com/commaai/openpilot/pull/23331

### DIFF
--- a/board/safety/safety_ford.h
+++ b/board/safety/safety_ford.h
@@ -1,12 +1,45 @@
-// board enforces
-//   in-state
-//      accel set/resume
-//   out-state
-//      cancel button
-//      accel rising edge
-//      brake rising edge
-//      brake > 0mph
+const struct lookup_t FORD_LOOKUP_ANGLE_RATE_UP = {
+  {2., 7., 17.},
+  {5., .8, .25}};
 
+const struct lookup_t FORD_LOOKUP_ANGLE_RATE_DOWN = {
+  {2., 7., 17.},
+  {5., 3.5, .8}};
+
+// Signal: LatCtlPath_An_Actl
+// Factor: 0.0005
+const int FORD_DEG_TO_CAN = 2000;
+
+// Safety-relevant CAN messages for Ford vehicles.
+#define MSG_STEERING_PINION_DATA      0x07E  // RX from PSCM, for steering pinion angle
+#define MSG_ENG_BRAKE_DATA            0x165  // RX from PCM, for driver brake pedal and cruise state
+#define MSG_ENG_VEHICLE_SP_THROTTLE2  0x202  // RX from PCM, for vehicle speed
+#define MSG_ENG_VEHICLE_SP_THROTTLE   0x204  // RX from PCM, for driver throttle input
+#define MSG_STEERING_DATA_FD1         0x083  // TX by OP, ACC control buttons for cancel
+#define MSG_LANE_ASSIST_DATA1         0x3CA  // TX by OP, Lane Keeping Assist
+#define MSG_LATERAL_MOTION_CONTROL    0x3D3  // TX by OP, Lane Centering Assist
+#define MSG_IPMA_DATA                 0x3D8  // TX by OP, IPMA HUD user interface
+
+// CAN bus numbers.
+#define FORD_MAIN 0U
+#define FORD_CAM  2U
+
+const CanMsg FORD_TX_MSGS[] = {
+  {MSG_STEERING_DATA_FD1, 0, 8},
+  {MSG_LANE_ASSIST_DATA1, 0, 8},
+  {MSG_LATERAL_MOTION_CONTROL, 0, 8},
+  {MSG_IPMA_DATA, 0, 8},
+};
+#define FORD_TX_LEN (sizeof(FORD_TX_MSGS) / sizeof(FORD_TX_MSGS[0]))
+
+AddrCheckStruct ford_addr_checks[] = {
+  {.msg = {{MSG_STEERING_PINION_DATA, 0, 8, .expected_timestep = 10000U}, { 0 }, { 0 }}},
+  {.msg = {{MSG_ENG_BRAKE_DATA, 0, 8, .expected_timestep = 100000U}, { 0 }, { 0 }}},
+  {.msg = {{MSG_ENG_VEHICLE_SP_THROTTLE2, 0, 8, .expected_timestep = 20000U}, { 0 }, { 0 }}},
+  {.msg = {{MSG_ENG_VEHICLE_SP_THROTTLE, 0, 8, .expected_timestep = 10000U}, { 0 }, { 0 }}},
+};
+#define FORD_ADDR_CHECK_LEN (sizeof(ford_addr_checks) / sizeof(ford_addr_checks[0]))
+addr_checks ford_rx_checks = {ford_addr_checks, FORD_ADDR_CHECK_LEN};
 
 static int ford_rx_hook(CANPacket_t *to_push) {
 
--- a/board/safety/safety_ford.h
+++ b/board/safety/safety_ford.h
@@ -8,7 +8,7 @@ const struct lookup_t FORD_LOOKUP_ANGLE_RATE_DOWN = {
 
 #define DEG_TO_RAD (3.14159265358979 / 180.0)
 #define RAD_TO_DEG (180.0 / 3.14159265358979)
-#define KPH_TO_MS (0.01 / 3.6)
+#define KPH_TO_MS (1 / 3.6)
 
 // Signal: LatCtlPath_An_Actl
 // Factor: 0.0005
@@ -75,10 +75,9 @@ static int ford_rx_hook(CANPacket_t *to_push) {
     // Update in motion state from speed value
     if (addr == MSG_ENG_VEHICLE_SP_THROTTLE2) {
       // Speed in km/h, convert to m/s
-      // Speed: (0.01 * val) in km/h
+      // Speed: (0.01 * val) * KPH_TO_MS
       // Signal: Veh_V_ActlEng
-      int raw_can_speed = ((GET_BYTE(to_push, 6) << 8) | GET_BYTE(to_push, 7)) * 0.01;
-      vehicle_speed = raw_can_speed * KPH_TO_MS;
+      vehicle_speed = (((GET_BYTE(to_push, 6) << 8) | GET_BYTE(to_push, 7)) * 0.01) * KPH_TO_MS;
       vehicle_moving = vehicle_speed > 0.3;
     }
 
--- a/board/safety/safety_ford.h
+++ b/board/safety/safety_ford.h
@@ -21,9 +21,10 @@ const struct lookup_t FORD_LOOKUP_ANGLE_RATE_DOWN = {
 #define MSG_ENG_VEHICLE_SP_THROTTLE2  0x202  // RX from PCM, for vehicle speed
 #define MSG_ENG_VEHICLE_SP_THROTTLE   0x204  // RX from PCM, for driver throttle input
 #define MSG_STEERING_DATA_FD1         0x083  // TX by OP, ACC control buttons for cancel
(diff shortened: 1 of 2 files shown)


jyoung8607:
Glad to see someone working on Ford!

> Code passes tests but I have yet to do thorough injection testing myself.

If you wrote Ford-specific tests, they don't seem to be in this PR. In fact, even the old Ford safety doesn't seem to have tests, which is why you didn't see any change in test outcomes from your changes. To get yourself started, look at `panda/tests/safety/test_[brand].py` for the brands that inspired your safety.

I see your DBC changes haven't been accepted just yet. In order to have your new message and signal structures available for the tests to run, you'll need to do your local test development with your Panda code checked out inside a fully-built openpilot environment that also has your opendbc changes.

board/safety/safety_ford.h
```diff
@@ -1,112 +1,219 @@
...
-  if (addr == 0x217) {
-    // wheel speeds are 14 bits every 16
-    vehicle_moving = false;
-    for (int i = 0; i < 8; i += 2) {
-      vehicle_moving |= GET_BYTE(to_push, i) | (GET_BYTE(to_push, (int)(i + 1)) & 0xFCU);
+    // Update steering angle
+    if (addr == MSG_STEERING_PINION_DATA) {
+      // Steering angle: (0.1 * val) - 3200 in deg
+      // Signal: StePinRelInit_An_Sns
+      int angle_meas_new = ((GET_BYTE(to_push, 0) << 8) | GET_BYTE(to_push, 1)) - 3200U;
```
jyoung8607, line 64: Need your scaling in addition to your offset, do you not? Haven't checked to see how this interplays with your steering command being in radians; I haven't used angle control before.

```suggestion
      // Steering angle: (0.1 * val) - 3200 in deg
      // Signal: StePinRelInit_An_Sns
      int angle_meas_new = ((GET_BYTE(to_push, 0) << 8) | GET_BYTE(to_push, 1)) * 0.1 - 3200U;
```

board/safety/safety_ford.h
```diff
@@ -1,112 +1,219 @@
...
-    }
-    if (set_or_resume) {
-      controls_allowed = 1;
+    // Update in motion state from speed value
+    if (addr == MSG_ENG_VEHICLE_SP_THROTTLE2) {
+      // Speed in kph, convert to m/s
+      // Speed: (0.01 * val) in kph
+      // Signal: Veh_V_ActlEng
+      vehicle_speed = ((GET_BYTE(to_push, 6) << 8) | GET_BYTE(to_push, 7)) * 0.01 / 3.6;
+      vehicle_moving = vehicle_speed > 0.3;
```
jyoung8607, line 74: If you're going to change the safety to obtain ego vehicle speed and standstill state from this signal, make sure your corresponding openpilot port does the same to avoid any possibility of subtle weird mismatches later. The current Ford port (and several others) does its own calculation of vEgoRaw from the mean value of all four ABS wheel speed sensors.

board/safety/safety_ford.h
```diff
@@ -1,112 +1,219 @@
...
-  if (addr == 0x165) {
-    brake_pressed = GET_BYTE(to_push, 0) & 0x20U;
-    if (brake_pressed && (!brake_pressed_prev || vehicle_moving)) {
-      controls_allowed = 0;
+    // Update gas pedal
+    if (addr == MSG_ENG_VEHICLE_SP_THROTTLE) {
+      // Pedal position: (0.1 * val) in percent
+      // Signal: ApedPos_Pc_ActlArb
+      unsigned int raw_gas_can = ((GET_BYTE(to_push, 0) & 0x03U) << 8) | GET_BYTE(to_push, 1);
+      gas_pressed = raw_gas_can > 0U;
```
jyoung8607, line 82: Would this not work?

```suggestion
      gas_pressed = ((GET_BYTE(to_push, 0) & 0x03U) << 8) | GET_BYTE(to_push, 1) > 0U;
```

board/safety/safety_ford.h
```diff
@@ -1,112 +1,219 @@
...
-    if (!unsafe_allow_gas && gas_pressed && !gas_pressed_prev) {
-      controls_allowed = 0;
+    // Update brake pedal and cruise state
+    if (addr == MSG_ENG_BRAKE_DATA) {
+      // Signal: BpedDrvAppl_D_Actl
+      brake_pressed = (GET_BYTE(to_push, 0) & 0x30U);
+
+      // Signal: CcStat_D_Actl
+      unsigned int cruise_state = (GET_BYTE(to_push, 1) & 0x07U);
+      bool cruise_engaged = (cruise_state != 0U) && (cruise_state != 3U);
```
jyoung8607, line 92: `cruise_state` will have eight potential values, some of which are undefined and some of which may be error or other abnormal states according to your DBC. Don't check for a couple known disengaged states and assume the rest are engaged, check for known and documented *engaged* states and assume the rest are disengaged.

board/safety/safety_ford.h
```diff
@@ -1,112 +1,219 @@
...
+      bus_fwd = FORD_CAM;
+      break;
+    }
+    case FORD_CAM: {
+      // Block stock LKAS messages (if OP is engaged)
+      bool is_lkas_msg = (addr == MSG_LANE_ASSIST_DATA1)
+                      || (addr == MSG_LATERAL_MOTION_CONTROL)
+                      || (addr == MSG_IPMA_DATA);
+      bool block_fwd = is_lkas_msg;
+      if (!block_fwd) {
```
jyoung8607, line 198:
```suggestion
      if (!is_lkas_msg) {
```

jyoung8607 (changes requested):

board/safety/safety_ford.h
incognitojam: I think you'd have to wrap it inside extra parentheses for CI checks to not complain, but yep

board/safety/safety_ford.h
incognitojam: I have already updated carstate to use this signal in my testing fork, so should be okay like this.

I haven't actually checked how similar the value is to the average wheel speed though. It might be worth checking that they do give similar speeds. I wasn't sure how the wheel speed, which is in radians/s according to the DBC, is converted to kph in carstate..
```
WhlFl_W_Meas @ 115.78s: 8.48 rad/s
Veh_V_ActlEng @ 115.79s: 11.90 kph
```

board/safety/safety_ford.h
incognitojam: Ah, I've just realised I meant to change this. In the Tesla safety code, they convert the value to the same units as what is used in the steering command. For the Ford, we will have to convert from degrees to radians I think.

```
Steering angle: (0.1 * val) - 3200 in deg
Steering command: (0.0005 * val) - 0.5 in radians
```

board/safety/safety_ford.h
incognitojam: Fixed

board/safety/safety_ford.h
incognitojam: Fixed, ty

board/safety/safety_ford.h
```diff
@@ -1,112 +1,217 @@
...
+  int bus_fwd = -1;
+
+  switch (bus_num) {
+    case FORD_MAIN: {
+      // Forward all traffic from bus 0 onward
+      bus_fwd = FORD_CAM;
+      break;
+    }
+    case FORD_CAM: {
+      // Block stock LKAS messages (if OP is engaged)
```
incognitojam, line 192: I copied this comment "if OP is engaged" from other safety code. Is this handled somewhere else in the panda?

board/safety/safety_ford.h
```diff
@@ -1,112 +1,227 @@
...
-    }
+
+  // Safety check for Lane_Assist_Data1 action
+  if (addr == MSG_LANE_ASSIST_DATA1) {
+    // Do not allow steering using Lane_Assist_Data1 (Lane-Departure Aid).
+    // This message must be sent for Lane Centering to work, and can include
+    // values such as the steering angle or lane curvature for debugging,
+    // but the action must be set to zero.
+    unsigned int action = (GET_BYTE(to_send, 0) & 0xE0U) >> 5;
+    violation |= (action != 0U);
```
incognitojam, line 129: I've added this check to make sure the action is set to zero on one of the steering commands. It's required to send this along with the lane centering command, otherwise it doesn't work, because at the moment we block the stock message.

We would want to block it, I think, so that the car's Lane Keeping Aid does not interfere with Openpilot. Maybe we should only block it when OP is engaged and pass-thru the action otherwise? Does the fwd_hook already do that for us and the tx_hook only blocks messages OP sends?

jyoung8607: Short version: Panda doesn't try and shouldn't try to make factory LKAS work when openpilot is disengaged.

Long version: You would have to solve a surprising number of technical issues to make switching back and forth work reliably and safely. If you're thinking "why couldn't we just X", I assure you not only be there dragons, but dragons hiding behind the dragons. After that, there's sort of an overall end-user experience design decision to be made, the look and feel of the openpilot experience, and comma have made the decision that either openpilot is driving or the driver is driving. I happen to agree. Some disagree, quite vehemently. It's one of those things that will descend into holy flame wars exceeded only by holy flame wars about sound design. It's okay to tinker with passthrough in a fork if you really want, but the version upstreamed to comma should not attempt passthrough of factory LKAS steering commands.

board/safety/safety_ford.h
jyoung8607: Not getting a hit on a search for that, I don't think it would be a correct comment for stock openpilot. See below.

board/safety/safety_ford.h
incognitojam: I'm happy to agree with upstream for now 😅

So, blocking all stock LKAS commands is the correct approach?

board/safety/safety_ford.h
incognitojam: `Veh_V_ActlEng` is used everywhere now

board/safety/safety_ford.h
incognitojam: Switched to `StePinComp_An_Est` signal, as it accounts for the pinion angle being initialised to zero at the start of the drive (no matter how the wheel was oriented). I've adjusted the scale and offset for this too.

incognitojam:
I've created the unit tests file, working on adding more tests! Copied some from Tesla for now. Something to consider is that the brake signal and PCM signal are both in the same CAN message. For now, I've done this:

```py
def _brake_msg(self, brake):
  values = {
    "BpedDrvAppl_D_Actl": 2 if brake else 1,
    "CcStat_D_Actl": 5 if self.safety.get_controls_allowed() else 0,
  }
  return self.packer.make_can_msg_panda("EngBrakeData", 0, values)
```

tests/safety/test_ford.py
```diff
@@ -0,0 +1,193 @@
...
+    self.packer = CANPackerPanda("ford_lincoln_base_pt")
+    self.safety = libpandasafety_py.libpandasafety
+    self.safety.set_safety_hooks(Panda.SAFETY_FORD, 0)
+    self.safety.init_tests()
+
+  def test_angle_cmd_when_enabled(self):
+    # when controls are allowed, angle cmd rate limit is enforced
+    speeds = [0., 1., 5., 10., 15., 50.]
+    # angles = [-300, -100, -10, 0, 10, 100, 300]
+    angles = [-23, -10, 0, 10, 23]
```
incognitojam, line 122: The DBC specifies that the range of angles we can command is only [-0.5, 0.5] radians... so this test needs to be updated for that. I also want to verify whether this is correct because 23.6 degrees isn't a huge range.

incognitojam:
Replaced by https://github.com/commaai/panda/pull/966

PR #815, opened 2021-12-22
Support for radarless honda

vanillagorillaa:
Requires https://github.com/commaai/opendbc/pull/481


sshane:
Moved to https://github.com/commaai/panda/pull/942

PR #817, opened 2022-01-03
Split linker files for STMs F2 and F4

### LINKED ISSUE #811: Split linker files for STMs F2 and F4
Make two separate linker files for F2(pedal) and F4(panda) as F2 has 128Kbytes SRAM and F4 has 320Kbytes SRAM. Extra SRAM might be used for CAN ring buffers.

briskspirit:
Closes #811


briskspirit:
Don't forget to revert Jenkins tests after testing on table.

PR #818, opened 2022-01-04
Add UNSAFE flags to panda safety declarations

spektor56:
Adding two unsafe flags, UNSAFE_SPLIT_LKAS_AND_ACC and UNSAFE_RESUME_LKAS_AFTER_BRAKE.

The implementations of these flags are used in some forks, so only the flag values will be up streamed as comma has no interest in implementing these features.  Reserving the flag values will enable forks to maintain safety with upstream.  If stock openpilot is run on a forks panda code these flags will stop the custom safety logic from running.

### DIFF
--- a/board/safety_declarations.h
+++ b/board/safety_declarations.h
@@ -142,6 +142,14 @@ struct sample_t angle_meas;         // last 3 steer angles
 // See ISO 15622:2018 for more information.
 #define UNSAFE_RAISE_LONGITUDINAL_LIMITS_TO_ISO_MAX 8
 
+// Enable the ability to toggle LKAS and ACC independently of each other. Allows openpilot to be engaged 
+// after pressing the LKAS button and doesn't block openpilot from engaging after pressing the cancel 
+// button.
+#define UNSAFE_SPLIT_LKAS_AND_ACC 16
+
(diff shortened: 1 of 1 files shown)


adeebshihadeh:
I don't understand the need for this. `pandad` doesn't launch `boardd` until the firmware matches.

spektor56:
this was the message from geo:

geohot — 11/14/2021
regardless of if it's upstreamed, it should be behind an unsafe flag. imagine changing forks w/o reflashing panda

And there are other cases where someone may try to combine forks and there are conflicts with safety flags, two unsafe flags may end up enabled at the same time unknowingly.

adeebshihadeh:
> this was the message from geo:
> 
> geohot — 11/14/2021 regardless of if it's upstreamed, it should be behind an unsafe flag. imagine changing forks w/o reflashing panda

I believe this was simply encouragement to develop your new experiences and features properly behind flags instead of changing the default behavior.

> And there are other cases where someone may try to combine forks and there are conflicts with safety flags, two unsafe flags may end up enabled at the same time unknowingly.

If you don't feel that you can confidently handle merge conflicts, I **strongly** recommend not modifying the panda code.

PR #821, opened 2022-01-09
don't play siren if ignition is low

adeebshihadeh:


board/main.c
```diff
@@ -695,13 +695,15 @@ void tick_handler(void) {
...
+
         // if the heartbeat has been gone for a while, go to SILENT safety mode and enter power save
-        if (heartbeat_counter >= (check_started() ? HEARTBEAT_IGNITION_CNT_ON : HEARTBEAT_IGNITION_CNT_OFF)) {
+        if (heartbeat_counter >= (ignition ? HEARTBEAT_IGNITION_CNT_ON : HEARTBEAT_IGNITION_CNT_OFF)) {
           puts("device hasn't sent a heartbeat for 0x");
           puth(heartbeat_counter);
           puts(" seconds. Safety is set to SILENT mode.\n");
 
-          if (controls_allowed_countdown > 0U) {
+          if (ignition && controls_allowed_countdown > 0U) {
```
adeebshihadeh, line 706:
```suggestion
          if (ignition && (controls_allowed_countdown > 0U)) {
```

pd0wm:
Is there also a cooldown on ignition? If the panda would loose al power we still want to sound the siren from the supercap, but in that case ignition would also go low.

PR #824, opened 2022-01-13
Gas resume support

jyoung8607:
Looks like we're going to have to fix this Chrysler gas resume thing right, so go ahead and fix it right-ish.

- Teach Panda safety how to deal with gas resume from standstill while engaged, disallowed by default
- Common tests for both the allowed and disallowed states
- Convert the Chrysler safety to use that instead of the speed hax, which never actually worked right
- Lose the custom Chrysler gas tests in favor of the updated common tests
- While here, clean up duplicate Chrysler brake disengage handling, taken care of in `generic_rx_checks()`

This does open a loophole wherein a Chrysler driver could gas-resume from engaged standstill and have openpilot start sending torque with gas still pressed, since (by designed intent here) there's no rising edge event until they release and reapply. I expect you may feel some kind of way about that. I started writing some "no torque with gas applied" conditions and tests, but ripped them back out because it seemed an ugly and incomplete hack.

**Long term:** safety and experience design for driver-triggered resume is something we need to think about anyway for E2E long, for resume from stop signs and stop lights. Perhaps `controls_allowed` needs to become a tri-state: disabled, enabled, resume/pre-enable? The third state could be used to permit only departing standstill, either by CC button spam or limited OP long, perhaps even OP long AEB when engaging in motion but gas not yet released.

**Short term:** are you okay with this PR as it stands? If not, what do you want to do about Chrysler?

### DIFF
--- a/board/safety.h
+++ b/board/safety.h
@@ -203,7 +203,7 @@ bool addr_safety_check(CANPacket_t *to_push,
 
 void generic_rx_checks(bool stock_ecu_detected) {
   // exit controls on rising edge of gas press
-  if (gas_pressed && !gas_pressed_prev && !(unsafe_mode & UNSAFE_DISABLE_DISENGAGE_ON_GAS)) {
+  if (gas_pressed && !gas_pressed_prev && !(gas_resume_allowed && !vehicle_moving) && !(unsafe_mode & UNSAFE_DISABLE_DISENGAGE_ON_GAS)) {
     controls_allowed = 0;
   }
   gas_pressed_prev = gas_pressed;
--- a/board/safety.h
+++ b/board/safety.h
@@ -267,6 +267,7 @@ int set_safety_hooks(uint16_t mode, int16_t param) {
   gas_interceptor_prev = 0;
   gas_pressed = false;
   gas_pressed_prev = false;
+  gas_resume_allowed = false;
   brake_pressed = false;
   brake_pressed_prev = false;
   cruise_engaged_prev = false;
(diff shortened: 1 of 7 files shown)


adeebshihadeh:
This is pretty clean, but I would prefer to remove this behavior. Looking back in the history, it seems that this just happened to slip through for the Chrysler port: https://github.com/commaai/panda/pull/130#issuecomment-436397047.

From the manuals, it seems that the resume button works as well. @adhintz is this true for all Chryslers that you know of?

2017 Pacifica:
```
If the target vehicle does not start moving within two seconds of your vehicle coming to a
standstill, the driver will either have to push the RES (resume) button, or apply the accelerator
pedal to reengage the ACC to the existing Set Speed
```

2021 Pacifica:
```
If your vehicle stays at a standstill for
longer than two seconds, then the driver
will either have to push the RES (resume)
button, or apply the accelerator pedal to
reengage Adaptive Cruise Control (ACC) to
the existing set speed.
```

adhintz:
> it seems that the resume button works as well. @adhintz is this true for all Chryslers that you know of?

Yes.

Resume button spamming should also be possible. We spam the cancel button to cancel ACC+OpenPilot and I've sent other button messages for other purposes as well. Would you merge a PR to spam the resume button on Chrysler? I've avoided spamming the resume button because I personally don't like it -- but if Comma likes it, I'm happy to send a PR for it.

jyoung8607:
> Resume button spamming should also be possible. We spam the cancel button to cancel ACC+OpenPilot and I've sent other button messages for other purposes as well. Would you merge a PR to spam the resume button on Chrysler? I've avoided spamming the resume button because I personally don't like it -- but if Comma likes it, I'm happy to send a PR for it.

If you can make resume spam work, that's absolutely the preferred solution. It's the experience on other openpilot cars, it would let us unwind some Chrysler specific gas resume hacks, and fix this EPS fault / controls mismatch at the same time.

adeebshihadeh:
> I've avoided spamming the resume button because I personally don't like it -- but if Comma likes it, I'm happy to send a PR for it.

Yes, this would be great. Let's start with removing the exception for gas press to resume. @jyoung8607 want to do that in this PR or I could do it too?

jyoung8607:
I can handle it. We'll normalize Chrysler gas handling in Panda and unwind the openpilot side as well, the gas rising edge speed check stuff was Chrysler specific as well.

PR #829, opened 2022-01-18
Add unsafe mode to health packet

sshane:


board/main.c
```diff
@@ -27,7 +27,7 @@
 extern int _app_start[0xc000]; // Only first 3 sectors of size 0x4000 are used
 
 // When changing this struct, boardd and python/__init__.py needs to be kept up to date!
-#define HEALTH_PACKET_VERSION 1
+#define HEALTH_PACKET_VERSION 2
```
sshane, line 30: I assume this is needed

sshane:
Is there anything else I should add to this PR?

pd0wm:
I'd like to add fields to end of the health packet, then older clients don't read the wrong data.

sshane:
That makes sense! I accidentally had the unsafe_mode_pkt as signed, I'll open a PR to fix that

PR #832, opened 2022-01-21
Deep sleep mode for pandas

### LINKED ISSUE #803: Deep sleep mode for panda
Look into adding sleep mode to panda(both F4 and H7).
Should wake up on interrupt from ignition line or CAN1 (for CAN based ignition)
Also if it is possible to wake up from USB.

briskspirit:
Use STOP mode for STMs to save some power. Wake up from CAN transceiver wake up(bus 0) or ignition line.

Deep sleep mode must be requested. Panda will go into deep sleep only if USB is not enumerated and there is no ignition.

Closes #803


board/drivers/rtc.h
```diff
@@ -21,9 +22,21 @@ uint16_t from_bcd(uint8_t value){
 }
 
 void rtc_init(void){
+  uint32_t bdcr_opts = 0U;
+  uint32_t bdcr_mask = 0U;
+  if ((hw_type == HW_TYPE_DOS) || (hw_type == HW_TYPE_UNO)) {
```
adeebshihadeh, line 27: think it makes more sense to keep the has_rtc field, and use it for this check

briskspirit: Nope, I need to choose between LSE and LSI clocks. Only Uno and Dos has external 32.768kHz.

adeebshihadeh: would still put that in the hw abstraction layer

board/main.c
```diff
@@ -758,6 +758,35 @@ void tick_handler(void) {
...
+void RTC_WKUP_IRQ_Handler(void) {
+  exti_irq_clear();
+  clock_init();
+
+  current_board->set_led(LED_RED, true);
+  delay(512000U);
+  current_board->set_led(LED_RED, false);
+  current_board->set_led(LED_BLUE, true);
+  delay(512000U);
+  current_board->set_led(LED_BLUE, false);
```
robbederks, line 787: wouldn't it make more sense to only toggle the LED state and immediately go back to sleep?

briskspirit: This PR is still draft ;) Testing different modes. And LED takes ~25mW, so not sure about toggling yet.

board/boards/black.h
```diff
@@ -206,7 +209,7 @@ const board board_black = {
   .has_hw_gmlan = false,
   .has_obd = true,
   .has_lin = false,
-  .has_rtc = false,
+  .has_rtc = true,
```
robbederks, line 212: would it make more sense to keep the old values and rename it to `has_rtc_battery`? Technically they all have RTC, so storing that in the struct isn't useful of course

robbederks: that way you can also check this field in the `rtc_init()` check adeeb was complaining about

briskspirit: Good point! +

PR #833, opened 2022-01-22
Reliable USB enumeration check

briskspirit:
This method is a lot more reliable than previous one(tested on BP and C3)


briskspirit:
> Did you also test this on C2? That's the main one it's useful for :)

Yes, works too.

PR #837, opened 2022-01-25
GM: Change from brake value threshold to Brake_Pressed signal

JasonJShuler:
Increases the threshold for a brake press in the safety code from 10 to 12 due to higher values in large-frame vehicles. This change is required to prevent false disengages. It must match the value in OP to prevent controls mismatch errors.

See corresponding OP PR for details: https://github.com/commaai/openpilot/pull/23617


JasonJShuler:
This one is updated to new signal and ready for review as well

JasonJShuler:
This needs to be merged as well - the two PRs are related

sshane:
Moved to https://github.com/commaai/panda/pull/846 for tweaks

PR #840, opened 2022-01-26
Honda: fix controls mismatch for stock PCM configs

adeebshihadeh:


board/safety/safety_honda.h
```diff
@@ -95,6 +95,9 @@ static int honda_rx_hook(CANPacket_t *to_push) {
   bool valid = addr_safety_check(to_push, &honda_rx_checks,
                                  honda_get_checksum, honda_compute_checksum, honda_get_counter);
 
+
```
adeebshihadeh, line 98:
```suggestion
```

PR #847, opened 2022-02-09
Remove CLOCK_SOURCE_MODE_EXTERNAL_SYNC

briskspirit:
Remove CLOCK_SOURCE_MODE_EXTERNAL_SYNC as it is not implemented and will interfere with deep sleep mode as deep sleep uses pin PC0 for EXTI (SBU1 line, ignition detection) and external sync uses pin PB0(GPS module time pulses)


robbederks (approved):
lgtm!

PR #853, opened 2022-02-14
Fix USB enumeration detection for C2

briskspirit:
Fix USB enumeration detection for C2


adeebshihadeh:
Nope, that would be #817

PR #855, opened 2022-02-15
Hyundai: do same enable check as openpilot

sshane:
On falling edge of enable button, set controls allowed

Should fix the Hyundai case here: https://github.com/commaai/openpilot/issues/22597


sshane:
Replaced by https://github.com/commaai/panda/pull/875

PR #863, opened 2022-03-01
Fix import paths in examples

2m:
Found this when trying out examples with Panda Red. I was running these examples from a virtual environment initialized with:

```
source ./venv/bin/activate.fish
python setup.py install
python examples/query_fw_versions.py
```

and the imports were not found with the current code in `master`. This PR fixes these imports.


2m:
Okay. I see that linter broke with my changes. Instead of the changes in my PR, I created a symlink `panda` that points to the current directory.

    ln -s . panda

This makes the examples work when running in virtual env as prepared with commands listed in my first post. Let me know if virtual env should be prepare differently.

PR #871, opened 2022-03-12
Subaru Torque limit increase. 

ClockeNessMnstr:
for the global platform the EPS enforces a maximum torque value and disables immediately if it's exceeded. 

This change enables the torque change on the '17-19 Impreza's and '18-19 Crosstreks (as well as the '20 Hybrid which is still in community) The max eps value is 3071, the value of 2047 was limiting the torque unintentionally. 

This change does not exceed torque already used on other models. 

2020+ impreza/crosstrek already has it's full torque using a value of 1439. 
2019 Forester has more torque to than either using it's value of 2047.


adeebshihadeh:
Leaving the openpilot PR open to track this, but will close this since it's a trivial diff.

PR #878, opened 2022-03-18
rename unsafeMode to alternativeExperience

sshane:


board/health.h
```diff
@@ -1,5 +1,5 @@
-// When changing this struct, boardd and python/__init__.py needs to be kept up to date!
-#define HEALTH_PACKET_VERSION 3
+// When changing this struct, python/__init__.py needs to be kept up to date!
+#define HEALTH_PACKET_VERSION 4
```
sshane, line 2: Do we need to update if only a name is changed?

board/health.h
```diff
@@ -1,4 +1,4 @@
-// When changing this struct, boardd and python/__init__.py needs to be kept up to date!
+// When changing this struct, python/__init__.py needs to be kept up to date!
 #define HEALTH_PACKET_VERSION 3
```
adeebshihadeh, line 2:
```suggestion
#define HEALTH_PACKET_VERSION 4
```

PR #880, opened 2022-03-18
Tesla: only allow cancel button

adeebshihadeh:
@robbederks any reason IDLE is also allowed? Seems openpilot only sends FWD. https://github.com/commaai/openpilot/blob/3d8537ab23197a383fd1dd7082329ddee1acfdb7/selfdrive/car/tesla/carcontroller.py#L56


robbederks:
Think this is left over from development, should be fine to remove!

PR #881, opened 2022-03-21
bump VW PQ ramp up rate to 10

actuallylemoncurd:
this is to allow better use of HCA 7 steering on PQ, also lets better control for normal HCA5 happen. all around just better to have, allows S curves to be taken easier


adeebshihadeh:
Looks like this limit is the same as MQB. @jyoung8607 I assume the PQ EPS is similar to MQB, or were the MQB limits just a starting point here?

In any case, we're currently developing a new procedure to set and validate these steering limits. Toyota is up first in #887, then we'll review and potentially increase the limits across the rest of the makes.

jyoung8607:
> Looks like this limit is the same as MQB. @jyoung8607 I assume the PQ EPS is similar to MQB, or were the MQB limits just a starting point here?

It's anecdotally comparable with the same limits. It has an alternate mode (the HCA 7 referred to above) that gives it a little more oomph by ignoring active return-to-center force. It's never been formally injection tested.

I'm all for increased available ramp-rate on both MQB and PQ, but I would like to test it for any adverse effects on tuning before it's rolled out. Last time I tried (a very long time ago) I had trouble with large oscillations at low speed. Might not be an issue anymore, and I'm sure it's fixable if it is, just want to make sure we don't accidentally regress.

PR #882, opened 2022-03-21
Hongqi HS5

jyoung8607:
Work in progress, mostly there. TODO:

- [ ] CC button (set/resume and cancel button handling), need to work this out in opendbc first
- [ ] verify ABS wheel speed signal scaling, current value is approximated
- [ ] confirm driver input torque threshold is reasonable
- [ ] explore factory EPS limits
- [ ] replay route

This passes CI testing locally, but won't pass in GitHub CI until the opendbc PR is merged and bumped in openpilot.


jyoung8607:
@robbederks @briskspirit It looks like part of the H7/Red Panda memory map is on the fine edge of overflowing. I've started having problems building this PR. Everything seems fine if I comment out a couple other unrelated safety modes. I don't know what happens if the H7 memory map is modified, or if it's even possible to modify.

https://github.com/commaai/panda/blob/f56ebf5b776b677bf12ec772b0223274dd798999/board/stm32h7/stm32h7x5_flash.ld#L64-L73.

image

briskspirit:
Will increase memory available size on Monday

briskspirit:
Should be fixed on master, "freed" ~28Kb

jyoung8607:
> Should be fixed on master, "freed" ~28Kb

Looks good now, thank you!

sshane:

PR #884, opened 2022-03-22
safety: disallow longitudinal actuation while gas is pressed

sshane:
For when UNSAFE_MODE disables disengage on gas

- [x] Toyota
- [x] Honda
  - [x] Nidec - https://github.com/commaai/panda/pull/889
  - [x] Bosch
- [x] GM
- [x] Hyundai
- [x] Tesla - https://github.com/commaai/panda/pull/890

<a></a>

- At a standstill: longitudinal will be allowed while user is either holding brake or not touching pedals
- While moving: longitudinal will only be allowed when user is not pressing either pedal
- Any: `longitudinal_allowed` is also gated on `controls_allowed` being True


sshane:
Think we need to expose `longitudinal_allowed`. The tests (ideally) test whether a certain message sets controls_allowed to false (in the rx hook), and then another test checks whether controls_allowed blocks or allows a certain message.

Usually a test doesn't test blocking behavior by having the panda receive a message that sets controls_allowed to false, but we would need to do that to test this new local variable computed in the generic tx hook.

board/safety.h
```diff
@@ -62,6 +62,10 @@ int safety_fwd_hook(int bus_num, CANPacket_t *to_fwd) {
   return (relay_malfunction ? -1 : current_hooks->fwd(bus_num, to_fwd));
 }
 
+bool get_longitudinal_allowed(void) {
+  return controls_allowed && !gas_pressed;
```
adeebshihadeh, line 66:
```suggestion
  // This enforce's openpilot's pre-enable state: controls_allowed is entered
  // immediately but actuation isn't allowed until both pedals are released.
  bool pedal_pressed = brake_pressed_prev && vehicle_moving;
  if (!(alternative_experience & ALT_EXP_DISABLE_DISENGAGE_ON_GAS)) {
    pedal_pressed = pedal_pressed || gas_pressed_prev;
  }
  return controls_allowed && !(pedal_pressed);
```

tests/safety/common.py
```diff
@@ -87,10 +87,13 @@ def test_alternative_experience_no_disengage_on_gas_interceptor(self):
...
     for g in range(0, 0x1000):
       self._rx(self._interceptor_msg(g, 0x201))
+      # Test we allow lateral, but not longitudinal
       self.assertTrue(self.safety.get_controls_allowed())
+      self.assertEqual(g <= self.INTERCEPTOR_THRESHOLD, self.safety.get_longitudinal_allowed())
+      # Make sure we can re-gain longitudinal actuation
       self._rx(self._interceptor_msg(0, 0x201))
       self.safety.set_gas_interceptor_detected(False)
-    self.safety.set_alternative_experience(ALTERNATIVE_EXPERIENCE.DEFAULT)
+      self.assertTrue(self.safety.get_longitudinal_allowed())
```
adeebshihadeh, line 96: the above line no longer makes sense

adeebshihadeh (commented):
Let's follow this up with passing `current_controls_allowed` to the TX hooks to properly enforce the pre-enable state everywhere. Let's also move the GM and Honda tests into the common tests in that PR. We'll also need to make sure this is properly handled on the openpilot side - I believe we only go into pre-enable when the gas is pressed.

board/safety.h
sshane: Don't we always want to consider both pedals actually? This isn't for lateral, just longitudinal. With this we'd allow longitudinal actuation when gas is pressed with no disengage on gas

adeebshihadeh: yeah, this was just missing the brake check

board/safety.h
```diff
@@ -62,6 +62,12 @@ int safety_fwd_hook(int bus_num, CANPacket_t *to_fwd) {
   return (relay_malfunction ? -1 : current_hooks->fwd(bus_num, to_fwd));
 }
 
+bool get_longitudinal_allowed(void) {
+  int pedal_pressed = brake_pressed_prev && vehicle_moving;
+  pedal_pressed = pedal_pressed || gas_pressed_prev;
+  return controls_allowed && !pedal_pressed;
```
sshane, line 68: Needs to be the same minus NO_DISENGAGE_ON_GAS check for the cars where we replace current_controls_allowed with this

PR #885, opened 2022-03-22
Honda: fix missing Honda interceptor tests

sshane:
We were calling `common.InterceptorSafetyTest.setUpClass()` from a Honda test, however its `setUpClass` checks if the class name is itself and if so, then skip the test. However, when calling from a child `cls.__name__` was always the parent, so it just skipped those tests.


sshane:
Closing for commit history preservation

PR #886, opened 2022-03-22
Honda: fix missing interceptor tests

sshane:
We were calling `common.InterceptorSafetyTest.setUpClass()` from a Honda test, however its `setUpClass` checks if the class name is itself and if so, then skip the test. However, when calling from a child `cls.__name__` was always the parent, so it just skipped those tests.

### DIFF
--- a/tests/safety/common.py
+++ b/tests/safety/common.py
@@ -60,9 +60,6 @@ class InterceptorSafetyTest(PandaSafetyTestBase):
       cls.safety = None
       raise unittest.SkipTest
 
-    # make sure interceptor is detected
-    cls._rx(cls._interceptor_msg(0, 0x201)) # pylint: disable=no-value-for-parameter
-
   @abc.abstractmethod
   def _interceptor_msg(self, gas, addr):
     pass
--- a/tests/safety/test_honda.py
+++ b/tests/safety/test_honda.py
@@ -379,7 +379,6 @@ class TestHondaNidecInterceptorSafety(TestHondaNidecSafety, common.InterceptorSa
   """
     Covers the Honda Nidec safety mode with a gas interceptor
   """
-  pass
 
 
 class TestHondaNidecAltSafety(TestHondaNidecSafety):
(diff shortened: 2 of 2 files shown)


tests/safety/common.py
```diff
@@ -60,9 +60,6 @@ def setUpClass(cls):
       cls.safety = None
       raise unittest.SkipTest
 
-    # make sure interceptor is detected
-    cls._rx(cls._interceptor_msg(0, 0x201)) # pylint: disable=no-value-for-parameter
```
sshane, line 64: Toyota never called this, and all tests call this at the start anyway

tests/safety/test_honda.py
```diff
@@ -379,9 +379,7 @@ class TestHondaNidecInterceptorSafety(TestHondaNidecSafety, common.InterceptorSa
   """
     Covers the Honda Nidec safety mode with a gas interceptor
   """
-  def setUp(self):
-    TestHondaNidecSafety.setUpClass()
-    common.InterceptorSafetyTest.setUpClass()
```
sshane, line 384: This caused these tests to skip

tests/safety/test_honda.py
```diff
@@ -379,6 +379,7 @@ class TestHondaNidecInterceptorSafety(TestHondaNidecSafety, common.InterceptorSa
   """
     Covers the Honda Nidec safety mode with a gas interceptor
   """
+  pass
```
sshane, line 382: Fails static analysis...

tests/safety/test_honda.py
```diff
@@ -379,9 +379,7 @@ class TestHondaNidecInterceptorSafety(TestHondaNidecSafety, common.InterceptorSa
   """
     Covers the Honda Nidec safety mode with a gas interceptor
   """
-  def setUp(self):
-    TestHondaNidecSafety.setUpClass()
-    common.InterceptorSafetyTest.setUpClass()
+  pass
```
adeebshihadeh, line 382:
```suggestion
```

Issue #888, opened 2022-03-24
Honda Nidec: add safety to ACC_HUD message


sshane:
This message is used to control the HUD as well as sends signals to the PCM to actuate longitudinal. Only applicable for Nidec Honda without pedal.

Currently there's no safety at all, and before https://github.com/commaai/openpilot/commit/c88ced1f10b939bf67fa63ca37adaf9adc39ee58, we were always filling the PCM signals with non-zero values while disengaged. *However*, on stock these PCM signals aren't always zero when disengaged either

PR #889, opened 2022-03-24
Honda Nidec: ACC_HUD safety

sshane:
Resolves: https://github.com/commaai/panda/issues/888


adeebshihadeh:
Let's make sure to add a test for this.

PR #891, opened 2022-03-26
safety: move common controls allowed check to safety_tx_hook

sshane:


sshane:
Looks like the safety replay never had any routes where we were in a pre-enable state (blocking messages while the exposed `controls_allowed` was 1. Do we want to expose `current_controls_allowed`? @adeebshihadeh

sshane:
We now allow steering in pre-enable: not needed

PR #894, opened 2022-03-26
tests: move common longitudinal tests to common.py

sshane:


tests/safety/common.py
```diff
@@ -391,10 +391,6 @@ def test_allow_brake_at_zero_speed(self):
     self.assertTrue(self.safety.get_controls_allowed())
     self._rx(self._brake_msg(0))
     self.assertTrue(self.safety.get_controls_allowed())
-    # rising edge of brake should disengage
```
sshane, line 394: test below does this

PR #895, opened 2022-03-26
tests: move common steering tests to common.py

sshane:
Most cars have very similar, if not identical functions, so move them all to `TorqueSteeringSafetyTest`, where each car passes on either `test_exceed_torque_error_limit` or `test_against_torque_driver` depending on if they limit torque by EPS torque (Toyota, Chrysler), or driver torque (rest).

I can also do individual makes so it makes it easier to review


sshane:
Done in https://github.com/commaai/panda/pull/951

PR #910, opened 2022-04-06
Make safety_param unsigned

sshane:
As we use it for flags and unsigned values, it would be better to have it unsigned.

- [x] change to uint32!


board/health.h
```diff
@@ -16,7 +16,7 @@ struct __attribute__((packed)) health_t {
   uint8_t car_harness_status_pkt;
   uint8_t usb_power_mode_pkt;
   uint8_t safety_mode_pkt;
-  int16_t safety_param_pkt;
+  uint16_t safety_param_pkt;
```
adeebshihadeh, line 19: think uint32_t is better here and that's what your cereal PR has

sshane: I have a TODO ;)

pd0wm:
We also need to bump `HEALTH_PACKET_VERSION`

PR #912, opened 2022-04-06
Dos: PWM based camera signals

robbederks:
This gets rid of the need for interrupt handling to generate the 20Hz camera signals.
Should improve the timing consistency, and should also limit the risk of interrupt starvation limiting the camera framerate.

### DIFF
--- a/board/boards/dos.h
+++ b/board/boards/dos.h
@@ -184,7 +184,7 @@ void dos_init(void) {
   }
 
   // Init clock source as internal free running
-  dos_set_clock_source_mode(CLOCK_SOURCE_MODE_FREE_RUNNING);
+  dos_set_clock_source_mode(CLOCK_SOURCE_MODE_PWM);
 }
 
 const harness_configuration dos_harness_config = {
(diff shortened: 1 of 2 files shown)


robbederks:
Important note: this does disable GPIOC5 (the sync signal output to BMX055). We don't use this anyways I think, so might not be a big deal.

We could also still let the interrupt fire to get this signal into the BMX, but not sure if that's worth it

pd0wm:
Sync signal is only to the BMX (not the LSM)? Then I'm fine with getting rid of it.

robbederks:
> Sync signal is only to the BMX (not the LSM)? Then I'm fine with getting rid of it.

Yes, that's the one which toggles the bit in the gyro reading

pd0wm:
And the LSM doesn't have a similar feature?

robbederks:
> And the LSM doesn't have a similar feature?

No

> Does the sync signal go to the SOM itself?

Yes, but so do the camera clock signals

pd0wm:
TODO (me):

- [ ] Write notebook to verify timings between frames and SOF between cameras
- [ ] Run in testing closet

pd0wm (approved):

PR #914, opened 2022-04-07
GM: allow active regen bit when gas is pressed

sshane:
Stock GM ACC keeps the active bit set (`GasRegenCmdActive`) when the user is overriding with gas, however with the addition of `longitudinal_allowed`, panda would always block these messages until a fault occurs.

Might need to use `controls_allowed` instead, as if we disengage on accelerator, we also won't allow the apply bit, which some users report causes faults as well. Will test

### DIFF
--- a/board/safety/safety_gm.h
+++ b/board/safety/safety_gm.h
@@ -202,12 +202,12 @@ static int gm_tx_hook(CANPacket_t *to_send, bool longitudinal_allowed) {
       if ((gas_regen != GM_MAX_REGEN)) {
         tx = 0;
       }
-      // Stock ECU keeps apply bit set when gas is pressed
-      if (!current_controls_allowed) {
-        bool apply = GET_BYTE(to_send, 0) & 1U;
-        if (apply) {
-          tx = 0;
-        }
+    }
+    // Need to allow apply bit in pre-enabled and overriding states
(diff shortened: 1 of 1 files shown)


board/safety/safety_gm.h
```diff
@@ -198,9 +198,15 @@ static int gm_tx_hook(CANPacket_t *to_send, bool longitudinal_allowed) {
...
-      bool apply = GET_BYTE(to_send, 0) & 1U;
       // Stock ECU sends max regen when not enabled
-      if (apply || (gas_regen != GM_MAX_REGEN)) {
+      if ((gas_regen != GM_MAX_REGEN)) {
+        tx = 0;
+      }
+    }
+    // Need to allow apply bit in pre-enabled and overriding states
+    if (!controls_allowed) {
+      bool apply = GET_BYTE(to_send, 0) & 1U;
```
adeebshihadeh, line 208: let's start moving this stuff to GET_BIT. it's nice since you can just copy the bit idx from the DBC

adeebshihadeh, line 208: let's start moving this stuff to GET_BIT. it's nice since you can just copy the bit idx from the DBC

adeebshihadeh (approved):

PR #917, opened 2022-04-15
Hyundai: enforce recent button press to allow controls with PCM

sshane:
- Safety code:

We just create two new `sample_t` structs; one for the main button and one for the cruise button, and only look at the last 4 values on rising edge of PCM cruise when not openpilot longitudinal. For main button, it's a single bit, so we check if it's 1, and for the cruise button we check that it's either CANCEL, RESUME, or SET. Any four of these actions can enable cruise control on newer model year-Hyundai.

- Safety tests:

`test_enable_control_allowed_from_cruise:`
```
Hyundai non-longitudinal only enables on PCM rising edge and recent button press. Tests PCM enabling with:
- disallowed: No buttons
- disallowed: Buttons that don't enable cruise
- allowed: Buttons that do enable cruise
- allowed: Main button with all above combinations
```

`test_sampling_cruise_buttons:`
```
Test that we allow controls on recent button press, but not as button leaves sliding window
```

### DIFF
--- a/board/safety/safety_hyundai.h
+++ b/board/safety/safety_hyundai.h
@@ -71,9 +71,8 @@ enum {
 
 // some newer HKG models can re-enable after spamming cancel button,
 // so keep track of user button presses to deny engagement if no interaction
-const int PREV_BUTTON_SAMPLES = 4;  // roughly 80 ms
-struct sample_t cruise_buttons;
-struct sample_t main_buttons;
+const uint8_t PREV_BUTTON_SAMPLES = 4;  // roughly 80 ms
+uint32_t last_button_interaction;
 
 bool hyundai_legacy = false;
 bool hyundai_ev_gas_signal = false;
@@ -176,9 +175,11 @@ static int hyundai_rx_hook(CANPacket_t *to_push) {
       int cruise_button = GET_BYTE(to_push, 0) & 0x7U;
       int main_button = GET_BIT(to_push, 3U);
 
-      // used for keeping track of user button interaction
-      update_sample(&cruise_buttons, cruise_button);
-      update_sample(&main_buttons, main_button);
+      if ((cruise_button == HYUNDAI_BTN_RESUME) || (cruise_button == HYUNDAI_BTN_SET) || (cruise_button == HYUNDAI_BTN_CANCEL) || (main_button != 0)) {
+        last_button_interaction = 0;
+      } else {
+        last_button_interaction += 1;
+      }
 
       if (hyundai_longitudinal) {
         // exit controls on cancel press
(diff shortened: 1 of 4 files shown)


board/safety/safety_hyundai.h
```diff
@@ -165,38 +171,62 @@ static int hyundai_rx_hook(CANPacket_t *to_push) {
...
+    }
+
+    // enter controls on rising edge of ACC and user button press, exit controls when ACC off
+    if (!hyundai_longitudinal && (addr == 1057)) {
+      // 2 bits: 13-14
+      int cruise_engaged = (GET_BYTES_04(to_push) >> 13) & 0x3U;
+      if (cruise_engaged && !cruise_engaged_prev) {
+        bool cruise_pressed_recent = false;
+        bool main_pressed_recent = false;
+        for (int i = 0; i < MIN(cruise_buttons.length, PREV_BUTTON_SAMPLES); i++) {
```
sshane, line 207: Can make a function for these two common loops, but it means a global function in safety.h that's make-specific and only ever used here.

sshane:
panda safety replayed before and after this PR with a re-engage: https://imgur.com/a/0QW9Hxr

board/safety/safety_hyundai.h
```diff
@@ -165,38 +170,48 @@ static int hyundai_rx_hook(CANPacket_t *to_push) {
...
-        int button = GET_BYTE(to_push, 0) & 0x7U;
+    // ACC steering wheel buttons
+    if (addr == 1265) {
+      int cruise_button = GET_BYTE(to_push, 0) & 0x7U;
+      int main_button = GET_BIT(to_push, 3U);
+
+      if ((cruise_button == HYUNDAI_BTN_RESUME) || (cruise_button == HYUNDAI_BTN_SET) || (cruise_button == HYUNDAI_BTN_CANCEL) || (main_button != 0)) {
+        last_button_interaction = 0U;
+      } else {
+        last_button_interaction = MIN(last_button_interaction + 1U, PREV_BUTTON_SAMPLES);
```
sshane, line 181: Do we care about using the smallest unsigned int and then clip like this for any optimization reasons?

board/safety/safety_hyundai.h
```diff
@@ -69,6 +69,11 @@ enum {
   HYUNDAI_BTN_CANCEL = 4,
 };
 
+// some newer HKG models can re-enable after spamming cancel button,
+// so keep track of user button presses to deny engagement if no interaction
+const uint8_t PREV_BUTTON_SAMPLES = 4;  // roughly 80 ms
+uint8_t last_button_interaction = PREV_BUTTON_SAMPLES;
```
adeebshihadeh, line 75: reset on this on init

adeebshihadeh: also prefix with hyundai_

tests/safety/test_hyundai.py
```diff
@@ -280,7 +319,7 @@ def test_buttons(self):
   def _pcm_status_msg(self, enable):
     raise NotImplementedError
 
-  def _button_msg(self, buttons):
+  def _button_msg(self, buttons, main_button=0):
```
adeebshihadeh, line 322: does this actually need to be redefined? should be the same msg right?

board/safety/safety_hyundai.h
```diff
@@ -165,38 +170,48 @@ static int hyundai_rx_hook(CANPacket_t *to_push) {
...
-        int cruise_engaged = (GET_BYTES_04(to_push) >> 13) & 0x3U;
-        if (cruise_engaged && !cruise_engaged_prev) {
+    }
+
+    // enter controls on rising edge of ACC and user button press, exit controls when ACC off
+    if (!hyundai_longitudinal && (addr == 1057)) {
+      // 2 bits: 13-14
+      int cruise_engaged = (GET_BYTES_04(to_push) >> 13) & 0x3U;
+      if (cruise_engaged && !cruise_engaged_prev) {
+        if (last_button_interaction < PREV_BUTTON_SAMPLES) {
```
adeebshihadeh, line 206: add this to the above if

tests/safety/test_hyundai.py
sshane: Yes, for linter reasons

board/safety/safety_hyundai.h
```diff
@@ -69,6 +69,11 @@ enum {
   HYUNDAI_BTN_CANCEL = 4,
 };
 
+// some newer HKG models can re-enable after spamming cancel button,
+// so keep track of user button presses to deny engagement if no interaction
+const uint8_t HYUNDAI_PREV_BUTTON_SAMPLES = 4;  // roughly 80 ms
+uint8_t hyundai_last_button_interaction;
```
adeebshihadeh, line 75:
```suggestion
uint8_t hyundai_last_button_interaction;   // button messages since the user pressed a button
```

PR #927, opened 2022-04-27
Toyota: stock longitudinal safety

sshane:
Only allow ACC_CONTROL messages which have the CANCEL_REQ bit set if not op longitudinal


tests/safety/common.py
```diff
@@ -450,7 +450,9 @@ def test_tx_hook_on_wrong_safety_mode(self):
             if 'Tesla' in attr and 'Tesla' in current_test:
               continue
 
-            if {attr, current_test} == {'TestToyotaSafety', 'TestToyotaAltBrakeSafety'}:
+            if {attr, current_test} in ({'TestToyotaSafety', 'TestToyotaAltBrakeSafety'},
+                                        {'TestToyotaSafety', 'TestToyotaStockLongitudinal'},
+                                        {'TestToyotaAltBrakeSafety', 'TestToyotaStockLongitudinal'}):
```
adeebshihadeh, line 455: just check if it's a subset

tests/safety/common.py
```diff
@@ -447,10 +447,10 @@ def test_tx_hook_on_wrong_safety_mode(self):
           tx = getattr(getattr(test, attr), "TX_MSGS")
           if tx is not None and not attr.endswith('Base'):
             # No point in comparing different Tesla safety modes
-            if 'Tesla' in attr and 'Tesla' in current_test:
+            if 'TestTesla' in attr and 'TestTesla' in current_test:
               continue
 
-            if {attr, current_test} == {'TestToyotaSafety', 'TestToyotaAltBrakeSafety'}:
+            if 'TestToyota' in attr and 'TestToyota' in current_test:
```
adeebshihadeh, line 453: just add the new test name to the set and check if `{attr, current_test}` is a subset. if you add a new Toyota test now, this test can be unintentionally skipped

PR #933, opened 2022-05-04
Hyundai: use GET_BIT and fix safety on incorrect bits

sshane:
This fixes an incorrect signal, always looking at `FCA_CmdAct` instead of `FCA_CmdAct`

### DIFF
--- a/board/safety/safety_hyundai.h
+++ b/board/safety/safety_hyundai.h
@@ -230,7 +230,7 @@ static int hyundai_rx_hook(CANPacket_t *to_push) {
     }
 
     if (addr == 916) {
-      brake_pressed = (GET_BYTE(to_push, 6) >> 7) != 0U;
+      brake_pressed = GET_BIT(to_push, 55U) != 0U;
     }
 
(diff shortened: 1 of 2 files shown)


board/safety/safety_hyundai.h
```diff
@@ -259,8 +259,8 @@ static int hyundai_tx_hook(CANPacket_t *to_send, bool longitudinal_allowed) {
   // FCA11: Block any potential actuation
   if (addr == 909) {
     int CR_VSM_DecCmd = GET_BYTE(to_send, 1);
-    int FCA_CmdAct = (GET_BYTE(to_send, 2) >> 5) & 1U;
```
sshane, line 262: thanks to @sunnyhaibin for the find, this was wrong, it was getting `FCA_StopReq`

adeebshihadeh: why didn't the tests catch this?

adeebshihadeh:
Is this good to merge? Also, functional changes should always be mentioned in commits like this which initially appear to just be a null-effect refactor.

PR #934, opened 2022-05-04
Toyota: forward ACC messages if stock longitudinal

eFiniLan:
we have toyota_stock_longitudinal in safety_toyota.h

can we not block 0x343 if toyota_stock_longitudinal is set to true?

### DIFF
--- a/board/safety/safety_toyota.h
+++ b/board/safety/safety_toyota.h
@@ -287,7 +287,7 @@ static int toyota_fwd_hook(int bus_num, CANPacket_t *to_fwd) {
     // in TSS2, 0x191 is LTA which we need to block to avoid controls collision
     int is_lkas_msg = ((addr == 0x2E4) || (addr == 0x412) || (addr == 0x191));
     // in TSS2 the camera does ACC as well, so filter 0x343
-    int is_acc_msg = (addr == 0x343);
+    int is_acc_msg = !toyota_stock_longitudinal && (addr == 0x343);
     int block_msg = is_lkas_msg || is_acc_msg;
     if (!block_msg) {
       bus_fwd = 0;
--- a/board/safety/safety_toyota.h
+++ b/board/safety/safety_toyota.h
@@ -288,7 +288,7 @@ static int toyota_fwd_hook(int bus_num, CANPacket_t *to_fwd) {
     int is_lkas_msg = ((addr == 0x2E4) || (addr == 0x412) || (addr == 0x191));
     // in TSS2 the camera does ACC as well, so filter 0x343
     int is_acc_msg = (addr == 0x343);
-    int block_msg = is_lkas_msg || (is_acc_msg && !toyota_stock_long);
+    int block_msg = is_lkas_msg || (is_acc_msg && !toyota_stock_longitudinal);
     if (!block_msg) {
       bus_fwd = 0;
     }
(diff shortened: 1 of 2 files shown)


sshane:
I had this on a branch with safety tests before I realized the new 2022 RAV4's don't send 0x343 from the camera (https://github.com/commaai/panda/pull/873).

Not sure we want safety that's not doing anything, but it does allow forks to use stock Toyota longitudinal with upstream/stock panda fw. @adeebshihadeh what do you think?

eFiniLan:
> I had this on a branch before I realized the new 2022 RAV4's don't send 0x343 from the camera (#873).
> 
> Not sure we want safety that's not doing anything, but it does allow forks to use stock Toyota longitudinal with upstream/stock panda fw. @adeebshihadeh what do you think?

Thanks Shane!

I think the cases are:
1. for whoever wants to use stock long, they can just set `ret.openpilotLongitudinalControl = False`.
2. for the vehicle that has unparsable radar data + ACC coming from the camera (potentially).

board/safety/safety_toyota.h
```diff
@@ -287,7 +287,7 @@ static int toyota_fwd_hook(int bus_num, CANPacket_t *to_fwd) {
     // in TSS2, 0x191 is LTA which we need to block to avoid controls collision
     int is_lkas_msg = ((addr == 0x2E4) || (addr == 0x412) || (addr == 0x191));
     // in TSS2 the camera does ACC as well, so filter 0x343
-    int is_acc_msg = (addr == 0x343);
+    int is_acc_msg = !toyota_stock_longitudinal && (addr == 0x343);
     int block_msg = is_lkas_msg || is_acc_msg;
```
sshane, line 291:
```suggestion
    int is_acc_msg = (addr == 0x343);
    int block_msg = is_lkas_msg || (is_acc_msg && !toyota_stock_long);
```

sshane: name of is_acc_msg still makes sense here

board/safety/safety_toyota.h
```diff
@@ -288,7 +288,7 @@ static int toyota_fwd_hook(int bus_num, CANPacket_t *to_fwd) {
     int is_lkas_msg = ((addr == 0x2E4) || (addr == 0x412) || (addr == 0x191));
     // in TSS2 the camera does ACC as well, so filter 0x343
     int is_acc_msg = (addr == 0x343);
-    int block_msg = is_lkas_msg || is_acc_msg;
+    int block_msg = is_lkas_msg || (is_acc_msg && !toyota_stock_long);
```
sshane, line 291:
```suggestion
    int block_msg = is_lkas_msg || (is_acc_msg && !toyota_stock_longitudinal);
```

sshane:
Think is a good PR for the case of adding a new car that we don't have a radar parser for, but sends ACC messages from the camera, going to merge

tests/safety/test_toyota.py
```diff
@@ -191,6 +191,11 @@ def test_acc_cancel(self):
         should_tx = np.isclose(accel, 0, atol=0.0001)
         self.assertEqual(should_tx, self._tx(self._accel_msg(accel, cancel_req=1)))
 
+  def test_fwd_hook(self):
+    # remove 0x343
```
sshane, line 195:
```suggestion
    # remove ACC_CONTROL
```

tests/safety/test_toyota.py
```diff
@@ -191,6 +191,11 @@ def test_acc_cancel(self):
         should_tx = np.isclose(accel, 0, atol=0.0001)
         self.assertEqual(should_tx, self._tx(self._accel_msg(accel, cancel_req=1)))
 
+  def test_fwd_hook(self):
+    # remove ACC_CONTROL
```
sshane, line 195:
```suggestion
    # forward ACC_CONTROL
```

PR #936, opened 2022-05-11
Toyota: limit using driver torque

sshane:
Will verify all cars have similar driver torque scales, but shouldn't need an EPS factor

### DIFF
--- a/board/safety/safety_toyota.h
+++ b/board/safety/safety_toyota.h
@@ -5,7 +5,7 @@ const int TOYOTA_MAX_TORQUE = 1500;       // max torque cmd allowed ever
 // packet is sent at 100hz, so this limit is 1500/sec
 const int TOYOTA_MAX_RATE_UP = 15;        // ramp up slow
 const int TOYOTA_MAX_RATE_DOWN = 25;      // ramp down fast
-const int TOYOTA_DRIVER_TORQUE_ALLOWANCE = 10;
+const int TOYOTA_DRIVER_TORQUE_ALLOWANCE = 100;
 const int TOYOTA_DRIVER_TORQUE_FACTOR = 10;
 
 // real time torque limit to prevent controls spamming
--- a/board/safety/safety_toyota.h
+++ b/board/safety/safety_toyota.h
@@ -5,7 +5,8 @@ const int TOYOTA_MAX_TORQUE = 1500;       // max torque cmd allowed ever
 // packet is sent at 100hz, so this limit is 1500/sec
 const int TOYOTA_MAX_RATE_UP = 15;        // ramp up slow
 const int TOYOTA_MAX_RATE_DOWN = 25;      // ramp down fast
-const int TOYOTA_MAX_TORQUE_ERROR = 350;  // max torque cmd in excess of torque motor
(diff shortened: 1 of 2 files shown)


rbiasini:
Just totally out of curiosity: is the check on driver torque preferred over motor torque comparison?
Check on motor torque seems a bit more comprehensive, as it might include discrepancies between commanded and actual that are unrelated to the steer torque applied by the driver.

sshane:
Yeah, I'm not too sure if this is a huge improvement now. Originally I believed that the motor torque limit was causing loss of control on turns when hitting bumps, as that would limit torque when the driver isn't touching the wheel, but with good tuning I haven't really seen any issues with lack of control in normal turns.

The two torque limiting methods are enforcing different things though, with most of our brands using the driver torque limits. The driver torque limits solely enforces that the user can easily and safely take over without the wheel fighting them, while the motor torque limits enforces overriding as well as making sure the commanded torque is close to the already applied EPS torque. Which, under normal scenarios the ramp limits should take care of. If the motor isn't responding to our torque commands then it sounds like we have another problem.

It's worth exploring more though, and I'll definitely look at data and segments before any other work is done. What do you think @rbiasini?

rbiasini:
> Originally I believed that the motor torque limit was causing loss of control on turns when hitting bumps, as that would limit torque when the driver isn't touching the wheel, but with good tuning I haven't really seen any of these issues recently.

mmm... putting torque motor and torque cmd on the same plot should tell you if this is the case. Definitely not ideal if torque drop happens due to unexpected discrepancies between the 2 torques, regardless the tuning.

> with most of our brands using the driver torque limits

yeah, motor_torque isn't available on all brands. :) Or maybe it is?

> while the motor torque limits enforces overriding as well as making sure the commanded torque is close to the already applied EPS torque

Exactly

> Which, under normal scenarios I believe the ramp limits should take care of this. If the motor isn't responding to our torque commands then it sounds like we have another problem.

It's a very remote scenario, but imagine that the EPS is not responding to the commanded torque: the controller would wind up the command to the max allowed value in case safety is enforced based on the driver's torque (assume it's 0). Then, if all of a sudden the EPS starts responding again, then the rate limit might be violated. The check on the diff between motor torque and commanded torque would catch this by disallowing the commanded torque to wind up.

Personal 2 cents: the check on driver's torque is good. The check on motor torque is a bit more comprehensive. A third over complex option is to enforce both.

sshane:
Thanks for the reply! Yeah, I suspected we might not be using the motor torque limiter on other brands because a lack of motor signal.

> Definitely not ideal if torque drop happens due to unexpected discrepancies between the 2 torques, regardless the tuning.

I think this is definitely possible in some situations, where we aren't able to apply the desired torque because of mismatched EPS scaling or hitting bumps/potholes, lag in the EPS or signal, etc., but I'm now not convinced I'm hitting it very often, perhaps it was worse with a bad PID tune that constantly overshot.

I do think that having this driver torque blending limiter also implicitly check we don't exceed the effective torque rate limits due to the rare case the EPS isn't responding to our commands isn't the clearest way to do that, when none of the other makes do this (except one), so maybe there can be some other method implemented in the future that ensures we don't limit torque when we don't need to.

For now, I'll close this unless I find some cases where we're significantly limiting torque for no reason.

PR #938, opened 2022-05-13
Toyota steer fault workaround safety

sshane:
I did all my testing on vl, not vl_all (message based), and I believe that the fault is also time based (faults in the same time sending at 50hz vs 100hz), so I'm not sure how the car will respond to doing the torque cutting on messages vs. time/frames

Will make another PR that just allows one message with mismatched steer_req bit as that will allow the correct torque cutting from openpilot, and also be easier for the Hyundai fault workaround as well

In short, do we want to allow the behavior of a mismatched torque request bit from openpilot at any time, or do we want to do an identical check on openpilot's side?


sshane:
This safety is much simpler and shouldn't have any possibilities for race conditions: https://github.com/commaai/panda/pull/939/files

PR #940, opened 2022-05-14
Toyota: check STEER_REQUEST bit

sshane:
We could previously ramp up while the bit is 0 and then set to 1 and effectively bypass the openpilot and panda torque rate limits (although not too much as the max torque would be imposed to 350 away from the current due to the torque blending)

### DIFF
--- a/tests/safety/test_toyota.py
+++ b/tests/safety/test_toyota.py
@@ -141,9 +141,8 @@ class TestToyotaSafety(common.PandaSafetyTest, common.InterceptorSafetyTest,
     """
       On Toyota, you can ramp up torque and then set the steer_req bit and the EPS will ramp up faster than
       the effective panda safety limits. This tests:
-        - Nothing is sent when cutting torque after the first message
+        - Nothing is sent when cutting torque
         - Nothing is blocked when sending torque normally
-        - We can recover after a mismatch
     """
     self.safety.set_controls_allowed(True)
     for _ in range(100):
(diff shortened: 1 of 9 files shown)


tests/safety/test_toyota.py
```diff
@@ -137,6 +137,23 @@ def test_lta_steer_cmd(self):
...
         self.assertEqual(should_tx, self._tx(self._lta_msg(req, req2, angle)))
 
+  def test_steer_req_bit(self):
+    """
+      On Toyota, you can ramp up torque and then set the steer_req bit and the EPS will ramp up faster than
+      the effective panda safety limits. This tests:
+        - Nothing is sent when cutting torque after the first message
+        - Nothing is blocked when sending torque normally
+        - We can recover after a mismatch
+    """
```
sshane, line 147:
```suggestion
  def test_steer_req_bit(self):
    """
      On Toyota, you can ramp up torque and then set the steer_req bit and the EPS will ramp up faster than
      the effective panda safety limits. This tests:
        - Nothing is sent when cutting torque
        - Nothing is blocked when sending torque normally
    """
```

tests/safety/test_toyota.py
```diff
@@ -137,6 +137,22 @@ def test_lta_steer_cmd(self):
...
         self.assertEqual(should_tx, self._tx(self._lta_msg(req, req2, angle)))
 
+  def test_steer_req_bit(self):
+    """
+      On Toyota, you can ramp up torque and then set the steer_req bit and the EPS will ramp up faster than
+      the effective panda safety limits. This tests:
+        - Nothing is sent when cutting torque
+        - Nothing is blocked when sending torque normally
+    """
+    self.safety.set_controls_allowed(True)
```
sshane, line 147:
```suggestion
    """
      On Toyota, you can ramp up torque and then set the STEER_REQUEST bit and the
      EPS will ramp up faster than the effective panda safety limits. This tests:
        - Nothing is sent when cutting torque
        - Nothing is blocked when sending torque normally
    """
    self.safety.set_controls_allowed(True)
```

PR #952, opened 2022-05-26
Mazda: fix steering safety and add tests

sshane:
In openpilot, the max steer is only 800 units, but in panda we were allowing up to 2048 with a super high realtime allowance (allowed 800 torque in 0.2 seconds and 2048 torque in 0.55 seconds, while the rate up limit only allowed reaching 800 torque in 0.8 seconds).

Mazda safety changes:
- Runs all driver torque tests
- Change max torque constant to openpilot's
- Lower realtime delta limits to be 20% higher than max rate up

### DIFF
--- a/board/safety/safety_mazda.h
+++ b/board/safety/safety_mazda.h
@@ -95,7 +95,7 @@ static int mazda_tx_hook(CANPacket_t *to_send, bool longitudinal_allowed) {
 
     // steer cmd checks
     if (addr == MAZDA_LKAS) {
-      int desired_torque = (((GET_BYTE(to_send, 0) & 0x0FU) << 8) | GET_BYTE(to_send, 1)) - 2048;
+      int desired_torque = (((GET_BYTE(to_send, 0) & 0x0FU) << 8) | GET_BYTE(to_send, 1)) - 2048U;
       bool violation = 0;
       uint32_t ts = microsecond_timer_get();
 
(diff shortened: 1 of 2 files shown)


board/safety/safety_mazda.h
```diff
@@ -12,12 +12,11 @@
 #define MAZDA_AUX  1
 #define MAZDA_CAM  2
 
-#define MAZDA_MAX_STEER 2048U
+#define MAZDA_MAX_STEER 800U
```
sshane, line 15: @Jafaral Do you know why this was set to 2048 if the limit in openpilot is only 800? I read here https://github.com/commaai/openpilot/pull/20864 you said newer models have more torque, but still don't respond above certain values, so think this is okay to correctly set?

Jafaral: I did see values north of 1500 in newer  models. @adeebshihadeh  suggested we keep the max at 800 in OP. Matching that in panda is probably fine.

sshane: Sounds good! If we ever want to verify these higher values are safe in openpilot we can always raise the limits and fix the `test_steer_safety_check` check (double 1500 would put it above the max signal value, wrapping back around to negative values)

Jafaral: The 2048 is the theoretical maximum. Sending high torques values is OK, the car trims these to whatever is allowed in the EPS without causing faults/errors.

sshane: I think the range is actually -2048 to 2047 here, 2048 would be -2048

Jafaral: you are correct about the range. I  just used a proper power of two so that I can use the constant as-is in the code which you updated also in the PR. All looking good.

PR #958, opened 2022-06-05
add missing type information to /python module, fix incorrect types, minor code simplifications

cmyui:
hey, just been reading through the code and adding/correcting typing information where possible

one question - which python versions are supported by this project? the [numpy version used in requirements.txt](https://numpy.org/doc/stable/release/1.21.0-notes.html) is >=python3.7 but the [setup.py file has 2 & 3 listed](https://github.com/commaai/panda/blob/master/setup.py#L61-L62).

if only >=3.7 is supported, i can use quite a few newer typing features

### DIFF
--- a/python/__init__.py
+++ b/python/__init__.py
@@ -8,13 +8,17 @@ import time
 import traceback
 import sys
 from functools import wraps
-from typing import Optional
+from typing import List, Optional, Sequence, Tuple, Union
+from typing_extensions import TypedDict
+from . import Panda
 from .dfu import PandaDFU, MCU_TYPE_F2, MCU_TYPE_F4, MCU_TYPE_H7  # pylint: disable=import-error
 from .flash_release import flash_release  # noqa pylint: disable=import-error
 from .update import ensure_st_up_to_date  # noqa pylint: disable=import-error
 from .serial import PandaSerial  # noqa pylint: disable=import-error
 from .isotp import isotp_send, isotp_recv  # pylint: disable=import-error
 from .config import DEFAULT_FW_FN, DEFAULT_H7_FW_FN  # noqa pylint: disable=import-error
+from ._typing import CAN_MSG
+
 
 __version__ = '0.0.10'
 
@@ -26,7 +30,7 @@ CANPACKET_HEAD_SIZE = 0x5
 DLC_TO_LEN = [0, 1, 2, 3, 4, 5, 6, 7, 8, 12, 16, 20, 24, 32, 48, 64]
 LEN_TO_DLC = {length: dlc for (dlc, length) in enumerate(DLC_TO_LEN)}
 
-def pack_can_buffer(arr):
+def pack_can_buffer(arr: Sequence[CAN_MSG]) -> List[bytes]:
   snds = [b'']
   idx = 0
   for address, _, dat, bus in arr:
@@ -57,7 +61,7 @@ def pack_can_buffer(arr):
     snds[idx] = tx
   return snds
 
-def unpack_can_buffer(dat):
+def unpack_can_buffer(dat: bytes) -> List[CAN_MSG]:
   ret = []
   counter = 0
   tail = bytearray()
@@ -95,6 +99,8 @@ def unpack_can_buffer(dat):
         break
   return ret
 
+# TODO: type hints for these decorators
+
 def ensure_health_packet_version(fn):
   @wraps(fn)
   def wrapper(self, *args, **kwargs):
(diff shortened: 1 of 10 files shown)


python/uds.py
```diff
@@ -875,19 +885,19 @@ def request_upload(self, memory_address: int, memory_size: int, memory_address_b
     resp = self._uds_request(SERVICE_TYPE.REQUEST_UPLOAD, subfunction=None, data=data)
     max_num_bytes_len = resp[0] >> 4 if len(resp) > 0 else 0
     if max_num_bytes_len >= 1 and max_num_bytes_len <= 4:
-      max_num_bytes = struct.unpack('!I', (b"\x00" * (4 - max_num_bytes_len)) + resp[1:max_num_bytes_len + 1])[0]
+      max_num_bytes: int = struct.unpack('!I', (b"\x00" * (4 - max_num_bytes_len)) + resp[1:max_num_bytes_len + 1])[0]
```
cmyui, line 888: struct.unpack returns `tuple[Any, ...]`

python/uds.py
```diff
@@ -641,17 +641,17 @@ def access_timing_parameter(self, timing_parameter_type: TIMING_PARAMETER_TYPE,
...
     # TODO: parse response into multiple output values?
     return resp
 
-  def control_dtc_setting(self, dtc_setting_type: DTC_SETTING_TYPE):
+  def control_dtc_setting(self, dtc_setting_type: DTC_SETTING_TYPE) -> None:
     self._uds_request(SERVICE_TYPE.CONTROL_DTC_SETTING, subfunction=dtc_setting_type)
 
   def response_on_event(self, response_event_type: RESPONSE_EVENT_TYPE, store_event: bool, window_time: int,
-                        event_type_record: int, service_response_record: int):
+                        event_type_record: int, service_response_record: int) -> Mapping[str, Any]:
```
cmyui, line 654: this might actually be a good usecase for a [TypedDict](https://peps.python.org/pep-0589) to make the return type less generic

cmyui: would it be okay to return `"event_window_time": None` in the case `response_event_type == RESPONSE_EVENT_TYPE.REPORT_ACTIVATED_EVENTS` below? that way, the TypedDict can preserve [totality](https://peps.python.org/pep-0589/#totality)

cmyui: let me know if you'd like me to implement either

python/ccp.py
```diff
@@ -3,6 +3,33 @@
...
+  status: int
+  info: Optional[int]
+
+class DiagnosticServiceResponse(TypedDict):
+  length: int
+  type: int
+
+class ActionServiceResponse(TypedDict):
+  length: int
+  type: int
```
cmyui, line 31: If anyone has better names for these - i'm not too familiar with the protocol yet 😅

python/_typing.py
```diff
@@ -0,0 +1,3 @@
+from typing import Any, Tuple
+
+CAN_MSG = Tuple[int, Any, bytes, int]
```
cmyui, line 3: not necessarily required, just wasn't sure where to put this. seems better to have a name assigned to it rather than inlining `Tuple[int, Any, bytes, int]` everywhere.

cmyui:
i've mostly completed adding type information to the `/python` module

python/__init__.py
```diff
@@ -473,84 +513,102 @@ def has_canfd(self) -> bool:
...
     self._handle.controlWrite(Panda.REQUEST_OUT, 0xe5, int(enable), 0, b'')
 
-  def set_can_enable(self, bus_num, enable):
+  def set_can_enable(self, bus_num: int, enable: bool) -> None:
+    assert self._handle is not None
     # sets the can transceiver enable pin
     self._handle.controlWrite(Panda.REQUEST_OUT, 0xf4, int(bus_num), int(enable), b'')
 
-  def set_can_speed_kbps(self, bus, speed):
+  def set_can_speed_kbps(self, bus: int, speed: Union[int, float]) -> None:
```
cmyui, line 583: there are a few cases where i type `Union[int, float]` since the code will technically convert it to int - search for that type & let me know if you'd like these changed to `int` only

python/__init__.py
```diff
@@ -95,6 +99,8 @@ def unpack_can_buffer(dat):
         break
   return ret
 
+# TODO: type hints for these decorators
```
cmyui, line 102: i'm not exactly sure how to type these correctly, so i've left them out for now.
i'll figure them out and either add to this PR, or submit it in another one.

i think it's something along the lines of
```py
from typing import TypeVar
from typing_extension import ParamSpec

P = ParamSpec("P")
R = TypeVar("R")


def ensure_health_packet_version(fn: Callable[Concatenate[Panda, P], R]) -> Callable[Concatenate[Panda, P], R]:
  @wraps(fn)
  def wrapper(self, *args: P.args, **kwargs: P.kwargs) -> R:
    if self.health_version < self.HEALTH_PACKET_VERSION:
      raise RuntimeError("Panda firmware has outdated health packet definition. Reflash panda firmware.")
    elif self.health_version > self.HEALTH_PACKET_VERSION:
      raise RuntimeError("Panda python library has outdated health packet definition. Update panda python library.")
    return fn(self, *args, **kwargs)
  return wrapper

def ensure_can_packet_version(fn: Callable[Concatenate[Panda, P], R]) -> Callable[Concatenate[Panda, P], R]:
  @wraps(fn)
  def wrapper(self, *args: P.args, **kwargs: P.kwargs) -> R:
    if self.can_version < self.CAN_PACKET_VERSION:
      raise RuntimeError("Panda firmware has outdated CAN packet definition. Reflash panda firmware.")
    elif self.can_version > self.CAN_PACKET_VERSION:
      raise RuntimeError("Panda python library has outdated CAN packet definition. Update panda python library.")
    return fn(self, *args, **kwargs)
  return wrapper
```

adeebshihadeh:
Thanks for the PR, but closing since it's been sitting a while and has a bunch of conflicts. If you want to try again, it's best to break it up into the smallest chunks possible.

PR #962, opened 2022-06-08
GM: Add support for cam harness & stock ACC

JasonJShuler:
In preparation for non-ASCM based Silverado, Suburban and Tahoe support, safety model options
have been added to enable forwarding between the camera and powertrain bus (with appropriate
message filtering), and to optionally allow the use of stock ACC.
(Note that these changes should apply to any GM vehicle with ACC + LKAS that doesn't use an ASCM)

The friction brake command is permitted for tx on powertrain, and included in ACC command fwd filter.

Regen paddle updated to set brake_pressed instead of directly unsetting controls_allowed.

Contains WIP LKAS timing enforcement wrapped in an #ifdef for easy removal - 
may be necessary until the OP side EPS fix is completed.

TODO:
* [x] REMINDER: carcontroller.py should omit ASCM keepalive messages if they can be forwarded from the cam.
* [x] Testing of new scenarios and messages must be implemented
* ~~[ ] Work on preventing remaining EPS faults in OP~~ This will be done in OP
* ~~[ ] Determine why steering limit code is being triggered and resolve if possible~~ This will be a separate issue
* [x] Regression testing
* [x] Determine if any changes are necessary in safety code for self-contained ACC modules vs stock camera-based VOACC
* [x] Determine if changes are required if the front camera isn't connected (nothing to forward)
* ~~[ ] Reminder: Current port doesn't monitor the status of the brake controller on the chassis bus - might be good to add~~ Not in scope for this PR
* ~~[ ] Reminder: Investigate brake controller status for cars that use the PT bus for friction brakes~~ Not in scope for this PR
* [x] Resolve any MISRA violations introduced


board/safety/safety_gm.h
```diff
@@ -42,6 +53,18 @@ enum {
   GM_BTN_CANCEL = 6,
 };
 
+int gm_cam_bus = 2;
+
+enum {GM_OBD2, GM_CAM} gm_harness = GM_OBD2;
+bool gm_stock_long = false;
+
+#ifdef GM_EPS_TIMING_WORKAROUND
```
sshane, line 61: Let's do all this stuff in another PR for now

JasonJShuler: I didn't expect the timing stuff would be in the final pr - I'm hoping something can be done OP side.

tests/safety/test_gm.py
```diff
@@ -196,6 +213,10 @@ def setUp(self):
     self.safety.set_safety_hooks(Panda.SAFETY_GM, Panda.FLAG_GM_HW_CAM)
     self.safety.init_tests()
 
+  # TODO: implement this
```
sshane, line 216: @JasonJShuler Do you know what message this is?

JasonJShuler: On cars that engage/disengage entirely based on PCM status rather than using buttons, I suspect this would be implemented.
Recall when you had to roll back the use of the boolean brake_pressed signal because the PCM uses a different threshold? The PCM would disengage over bumps and such, and I think it caused a controls mismatch or something...

In my opinion it warrants investigation: if cruise status is being managed by some ECU, I don't know why we aren't using the ECU state for engage/disengage, rather than buttons. (It might be related to improving the stop/go experience?)

Especially where these new cars will be initially dependent on the stock ACC...

JasonJShuler:
@sshane is helping to update the tests, otherwise I think the main code is ready for review - I'm removing this from draft status

JasonJShuler:
@sshane are you ok if I rebase from commaai/master and force push to clean things up a tad?

adeebshihadeh:
> @sshane are you ok if I rebase from commaai/master and force push to clean things up a tad?

Yeah, that's fine. We always squash merge, so we don't care how dirty your branch is.

tests/safety/test_gm.py
sshane: I think it was because we assume we can control the ACC state, including enabling and disabling without spamming buttons, and controlling the ACC set speed. For example we can't enable cruise on Toyota and we don't have cruise set speed control, but we can with Hyundai.

Do you know if the PCM manages these things or does the camera?

tests/safety/test_gm.py
```diff
@@ -18,7 +18,7 @@ class Buttons:
   CANCEL = 6
 
 
-class TestGmSafety(common.PandaSafetyTest, common.DriverTorqueSteeringSafetyTest):
+class TestGmSafetyBase(common.PandaSafetyTest, common.DriverTorqueSteeringSafetyTest):
   TX_MSGS = [[384, 0], [1033, 0], [1034, 0], [715, 0], [880, 0],  # pt bus
```
sshane, line 22: Self note:

- One base test class (only subclassed from)
- A class for ASCM with op long
- A class for GM CAM (only stock long, we can subclass that to add op long tests)

sshane:
- [x] Made GM CAM use the PCM cruise state, just like openpilot

sshane:
Let's do the longitudinal support in a PR after this is merged

sshane:
- [ ] Remove resume safety

board/safety/safety_gm.h
```diff
@@ -218,12 +236,44 @@ static int gm_tx_hook(CANPacket_t *to_send, bool longitudinal_allowed) {
     }
   }
 
+  // BUTTONS: used for resume spamming and cruise cancellation
```
sshane, line 239:
```suggestion
  // BUTTONS: used for resume spamming and cruise cancellation with stock longitudinal
```

board/safety/safety_gm.h
```diff
@@ -218,12 +236,44 @@ static int gm_tx_hook(CANPacket_t *to_send, bool longitudinal_allowed) {
...
+
+  int bus_fwd = -1;
+
+  if (gm_hw != GM_ASCM) {
+    if (bus_num == 0) {
+      bus_fwd = 2;
+    }
+
+    if (bus_num == 2) {
+      // block lkas message, forward all others (including ACC messages)
```
sshane, line 263:
```suggestion
      // block lkas message, forward all others
```

board/safety/safety_gm.h
```diff
@@ -218,12 +236,44 @@ static int gm_tx_hook(CANPacket_t *to_send, bool longitudinal_allowed) {
...
+
   // 1 allows the message through
   return tx;
 }
 
+static int gm_fwd_hook(int bus_num, CANPacket_t *to_fwd) {
+
+  int bus_fwd = -1;
+
+  if (gm_hw != GM_ASCM) {
```
adeebshihadeh, line 257: any reason this is written like this? reads better as:
```suggestion
  if (gm_hw != GM_CAM) {
```

tests/safety/test_gm.py
```diff
@@ -180,5 +147,88 @@ def test_tx_hook_on_pedal_pressed_on_alternative_gas_experience(self):
...
+        self._rx(self._button_msg(Buttons.UNPRESS))
+        self.assertTrue(self.safety.get_controls_allowed())
+
+  def test_cancel_button(self):
+    self.safety.set_controls_allowed(1)
+    self._rx(self._button_msg(Buttons.CANCEL))
+    self.assertFalse(self.safety.get_controls_allowed())
+
+
+class TestGmCameraSafety(TestGmSafetyBase):
```
adeebshihadeh, line 203: missing a button test case

tests/safety/test_gm.py
```diff
@@ -180,5 +147,88 @@ def test_tx_hook_on_pedal_pressed_on_alternative_gas_experience(self):
...
+    self.packer = CANPackerPanda("gm_global_a_powertrain_generated")
+    self.packer_chassis = CANPackerPanda("gm_global_a_chassis")
+    self.safety = libpandasafety_py.libpandasafety
+    self.safety.set_safety_hooks(Panda.SAFETY_GM, Panda.FLAG_GM_HW_CAM)
+    self.safety.init_tests()
+
+  def _user_gas_msg(self, gas):
+    # FIXME: be careful with tests that use _user_gas_msg, we set CruiseState=1 so that
+    # certain gas tests aren't impacted by this safety mode's use of cruise state in the same message
+    values = {"AcceleratorPedal2": 1 if gas else 0, "CruiseState": 1}
```
adeebshihadeh, line 218: you can just read the cruise state and pass that in

board/safety/safety_gm.h
sshane: I think it was because there was going to be a GM_CAM_LONG, but we can figure that out later in another PR. That should probably be another variable like the other makes

tests/safety/test_gm.py
sshane: Also need to set pcm_cruise in all the tests then, and the global tests are ran on button enable and PCM cars, so we'd get missing implementation errors.

PR #963, opened 2022-06-09
Honda Bosch: consider correct bus for buttons

sshane:
Fixes a bug uncovered by https://github.com/commaai/panda/pull/874. We get quite a few messages from bus 2 where the first byte is 0xff for SCM_BUTTONS, we got unlucky as it parses out to 7 and considering buttons with no past didn't make this an issue (7 is not a recognized button) before. But now it's pretty much always guaranteed the last button is 7, so neither set or resume, causing controls_allowed never to rise.

This is a Bosch Civic: https://user-images.githubusercontent.com/25857203/172751996-27a04991-9232-4591-bdf1-70e13b5c409e.png

What panda sees: https://user-images.githubusercontent.com/25857203/172752445-56369f9e-712b-404a-8d62-3ddf91678646.png

Checked a Honda CR-V (Nidec), didn't have the button message on 0x1A6 on any other bus in normal operation, so only a bosch problem likely.

### DIFF
--- a/board/safety/safety_honda.h
+++ b/board/safety/safety_honda.h
@@ -145,7 +145,7 @@ static int honda_rx_hook(CANPacket_t *to_push) {
 
     // state machine to enter and exit controls for button enabling
     // 0x1A6 for the ILX, 0x296 for the Civic Touring
-    if (((addr == 0x1A6) || (addr == 0x296)) && bus == pt_bus) {
+    if (((addr == 0x1A6) || (addr == 0x296)) && (bus == pt_bus)) {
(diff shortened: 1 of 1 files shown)


sshane:
@adeebshihadeh Maybe it would be nice to change the addr check struct to make it so messages with incorrect length or bus aren't valid so we don't parse them? Right now they're just filters to find the counter and checksum checks, and if not it's always valid.

This would probably need changes across openpilot though, as we do existance checks for stock ecus for example, and those aren't in the addr checks.

PR #970, opened 2022-06-17
Chrysler: use universal gas and brake signals

sshane:


sshane:
Tests will pass when updating the pinned openpilot version

PR #976, opened 2022-06-22
Dockerfile: pin opendbc commit

sshane:
For cases where you need to merge a new signal to opendbc, however you can't bump opendbc on openpilot master until the panda PR is ready and tests pass. Previously you needed to make a second commit in panda to bump openpilot once it was merged (with failing panda tests).


Dockerfile.panda
```diff
@@ -65,6 +65,7 @@ RUN cd /tmp && \
     git fetch && \
     git checkout 1b0167ce24afb037b36464c40f9c5e0d657e77d9 && \
     git submodule update --init cereal opendbc rednose_repo && \
+    git -C opendbc checkout b8b3adfd9d64413642201954808810ea853782d7 && \
```
adeebshihadeh, line 68: always do a fetch, reset, and clean as well

PR #978, opened 2022-07-04
Pedal FW: Auto-detect installation in GM vehicle and apply ADC transform

JasonJShuler:
Pedal has required custom firmware to work in GM vehicles due to different electrical characteristics (the ECM has different internal resistance than (any) other brand, causing the ADC to read values lower than they should be)

Without this mitigation, the Pedal writes the incorrect ADC values to the DAC, which causes the vehicle to throw a DTC and show an error.

This version of the firmware automatically detects if it is in a GM vehicle vs other vehicles. Once it has detected being in a GM vehicle, the Pedal is latched into GM mode. The test is most likely performed before the car is fully powered on, and should appear transparent to openpilot.

If the user is pressing the pedal while the interceptor is starting up, it could cause a false-negative.

The scale factors were calculated by comparing the voltages at the throttle position sensor without a Pedal, and with a pedal attached.

Note: This transform is completely unrelated to the transform defined in the dbc file. Because the resistance is different in GM's ECM, the voltage values returned by the ADC are incorrect (too low). When the Pedal attempts to write the ADC value to the DAC, the output voltage is out of spec and the vehicle throws errors.

Also, the numbers used for the transform were selected to be able to use integer math.

### DIFF
--- a/board/pedal/main.c
+++ b/board/pedal/main.c
@@ -97,6 +97,11 @@ int usb_cb_control_msg(USB_Setup_TypeDef *setup, uint8_t *resp) {
 #define CAN_GAS_SIZE 6
 #define COUNTER_CYCLE 0xFU
 
+#define XFORM_MODE_UNKNOWN 0U
+#define XFORM_MODE_NORMAL 1U
+#define XFORM_MODE_GM 2U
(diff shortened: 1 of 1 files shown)


JasonJShuler:
After testing, this didn't work.

PR #986, opened 2022-07-12
compatible with lower version arm-none-eabi-gcc

maimaixiong:


adeebshihadeh:
What exactly does this fix and what version are you targeting?

PR #989, opened 2022-07-18
VW PQ: Longitudinal control safety, lateral torque update

jyoung8607:
* Prep for E2E longitudinal control, tested and working with current-state openpilot VOACC
* Open lateral control ramp-rates a little for use with openpilot's lateral acceleration torque controller


jyoung8607:
Ready for consideration. Notes:

* `3cfdec54aa035f3f|2022-07-19--23-45-10` needs to be uploaded for longitudinal-mode Panda safety replay testing. I've tested locally with the first two segments concatenated, so we see two full engages and disengages, and it looks good.

* Based on subjective comparison with my Golf R, and the low noise values from the test NMS Passat, I wanted to bring the driver torque threshold from 80 down to 60. Unfortunately it breaks the current stock ACC Panda safety replay route, and I don't have an immediate replacement. Will do this again in the future.

* For use with the new lateral torque tune, I bumped the max ramp rate a little to avoid clipping. Before we'd go to max in 1.5 seconds, now we can go to max in 1.0 seconds. The factory EPS lets us go to max in 0.6 seconds. I'd rather take it to factory max, but I figured this would be easier to justify for the first round.

adeebshihadeh:
Can you rebase? Seems there's some conflicts from my recent refactors.

jyoung8607:
> Can you rebase? Seems there's some conflicts from my recent refactors.

Done and passing.

PR #1004, opened 2022-08-01
CAN chunking refactor

robbederks:
The current chunking code relies on the following assumptions from the USB bulk transfers. This PR removes those assumptions, which aren't true for SPI comms:

- `max_len` is always 0x40 (the bulk transfer chunk size)
- the chunks are always requested in a max-size bulk transfer
- the chunk counter always starts at 0 for a new transfer, with the overflow buffer empty

### DIFF
--- a/python/__init__.py
+++ b/python/__init__.py
@@ -68,17 +68,30 @@ def unpack_can_buffer(dat, prev_rx_counter=None):
   else:
     can_rx_counter = dat[0]
 
-  for i in range(0, len(dat), 64):
-    if can_rx_counter != dat[i]:
-      print("CAN: LOST RECV PACKET COUNTER", can_rx_counter, dat[i])
-      break
+  i = 0
+  while i < len(dat):
+    # unpack chunk header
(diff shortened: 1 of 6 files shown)


robbederks:
Still needs loopback testing!

python/__init__.py
```diff
@@ -42,33 +50,73 @@ def pack_can_buffer(arr):
...
+  chunks[-1] = CAN_CHUNK_HEADER_STRUCT.pack(counter, prev_overflow_len, len(chunks[-1])) + chunks[-1]
+  transfers.append(b"".join(chunks))
+
+  return transfers
 
-def unpack_can_buffer(dat):
+def unpack_can_buffer(dat, prev_rx_counter=None):
   ret = []
-  counter = 0
   tail = bytearray()
```
robbederks, line 88: Resetting the tail prevents having overflow data between different unpack buffers (should never happen on USB). We just discard this data for now, might still want to fix this for the SPI implementation

PR #1008, opened 2022-08-04
Honda Bosch Radarless longitudinal safety

royjr:
Works great when the openpilotLongitudinalControl toggle is set to true, but lights up the dash otherwise. May have something to do with these lines..

```
int is_acc_msg = honda_bosch_radarless && (addr == 0x1C8);
if (!is_lkas_msg && !is_acc_msg) {
```

[See openpilot PR](https://github.com/commaai/openpilot/pull/25364)


sshane:
We'll likely need a new safety param or modify the existing one to support enabling and disabling longitudinal

board/safety/safety_honda.h
```diff
@@ -8,7 +8,7 @@
 //      brake > 0mph
 const CanMsg HONDA_N_TX_MSGS[] = {{0xE4, 0, 5}, {0x194, 0, 4}, {0x1FA, 0, 8}, {0x200, 0, 6}, {0x30C, 0, 8}, {0x33D, 0, 5}};
 const CanMsg HONDA_BOSCH_TX_MSGS[] = {{0xE4, 0, 5}, {0xE5, 0, 8}, {0x296, 1, 4}, {0x33D, 0, 5}, {0x33DA, 0, 5}, {0x33DB, 0, 8}};  // Bosch
-const CanMsg HONDA_RADARLESS_TX_MSGS[] = {{0xE4, 0, 5}, {0x296, 2, 4}, {0x33D, 0, 8}};  // Bosch radarless
+const CanMsg HONDA_RADARLESS_TX_MSGS[] = {{0xE4, 0, 5}, {0x296, 2, 4}, {0x33D, 0, 8}, {0x1C8, 0, 8}};  // Bosch radarless
```
adeebshihadeh, line 11: need a second set of TX msgs for this

royjr: Something like this?

```suggestion
const CanMsg HONDA_RADARLESS_TX_MSGS[] = {{0xE4, 0, 5}, {0x296, 2, 4}, {0x33D, 0, 8}};  // Bosch radarless
const CanMsg HONDA_RADARLESS_LONG_TX_MSGS[] = {{0xE4, 0, 5}, {0x296, 2, 4}, {0x33D, 0, 8}, {0x1C8, 0, 8}};  // Bosch radarless w/ gas and brakes
```

sshane: Yep

board/safety/safety_honda.h
```diff
@@ -440,7 +440,8 @@ static int honda_bosch_fwd_hook(int bus_num, CANPacket_t *to_fwd) {
   if (bus_num == 2)  {
     int addr = GET_ADDR(to_fwd);
     int is_lkas_msg = (addr == 0xE4) || (addr == 0xE5) || (addr == 0x33D) || (addr == 0x33DA) || (addr == 0x33DB);
-    if (!is_lkas_msg) {
+    int is_acc_msg = honda_bosch_radarless && (addr == 0x1C8);
+    if (!is_lkas_msg && !is_acc_msg) {
```
sshane, line 444: Then need to check the longitudinal flag here to conditionally block the ACC message.

royjr:
Seems to work with the toggle now without faults.

DisableRadar: 0 - 1a5d045d2c531a6d|2022-08-08--00-17-08
DisableRadar: 1 - 1a5d045d2c531a6d|2022-08-08--00-18-59

adeebshihadeh (changes requested):
Can you also add the safety checks? https://github.com/commaai/panda/blob/f0d04b724882313ded3cc11bad0f4368fbd0dea9/board/safety/safety_honda.h#L308

royjr:
Steering seems to be using the existing check. Started on the accel check, but not too familiar with GET_BYTE.
https://github.com/royjr/panda/blob/civic22_long/board/safety/safety_honda.h#L336

tests/safety/test_honda.py
```diff
@@ -536,8 +541,43 @@ def setUp(self):
...
+    self.packer = CANPackerPanda("honda_civic_ex_2022_can_generated")
+    self.safety = libpanda_py.libpanda
+    self.safety.set_safety_hooks(Panda.SAFETY_HONDA_BOSCH, Panda.FLAG_HONDA_RADARLESS | Panda.FLAG_HONDA_BOSCH_LONG)
+    self.safety.init_tests()
+
+  def _send_accel_msg(self, accel):
+    values = {
+      "ACCEL_COMMAND": accel,
+    }
+    return self.packer.make_can_msg_panda("ACC_CONTROL", self.PT_BUS, values)
```
sshane, line 573: this isn't really doing anything. it can be super confusing how to test this properly though, so i'll make a PR to add common longitudinal acceleration tests

sshane: https://github.com/commaai/panda/pull/1298

PR #1009, opened 2022-08-05
Ram HD

realfast:
This implements the RAM_HD trucks into the standard Chrysler brand.


adeebshihadeh:
Continued in #1010, I wasn't able to push to this branch for some reason.

PR #1022, opened 2022-08-11
Simple integrating fan controller

robbederks:


robbederks:
TODO:
- [ ] do we need to expose the error integral / actual pwm output?
- [ ] board-specific max fan RPM?
- [ ] do we need a D-controller as well to reduce the alternator issue (commaai/openpilot#24688)?

adeebshihadeh:
> * [ ]  do we need to expose the error integral / actual pwm output?

I think it would be nice to have in the health packet.

PR #1023, opened 2022-08-11
Genesis gv70 support

zunichky:


zunichky:
Need to make this generic for all HDA2 vehicles.

PR #1024, opened 2022-08-11
bootstub: check MCU before launching app

briskspirit:
Couldn't find a way to do that without changing bootstub, difference in isr_vector won't let run even early_init() function.
Backwards compatible.


briskspirit:
Will restore CI file after more tests

tests/automated/1_program.py
```diff
@@ -13,13 +14,40 @@ def check_signature(p):
...
+
+  t_start = time.time()
+  while len(PandaDFU.list()) == 0:
+    print("waiting for DFU...")
+    time.sleep(0.1)
+    if timeout is not None and (time.time() - t_start) > timeout:
+      return False
+
+  dfu = PandaDFU(PandaDFU.st_serial_to_dfu_serial(p._serial, p._mcu_type))
+  fn = KNOWN_H7_BOOTSTUB_FN if p._mcu_type == MCU_TYPE_H7 else KNOWN_BOOTSTUB_FN
```
adeebshihadeh, line 33: this is exactly the kind of thing we need to clean up. this is now done in the panda python lib, pandad, and here. the panda python lib might not even respect the filename you pass in: https://github.com/commaai/panda/blob/508ee90f8ecf1ce70fc9d002bbc14b47a50a459a/python/__init__.py#L340

tests/automated/1_program.py
```diff
@@ -13,13 +14,40 @@ def check_signature(p):
 
 @test_all_pandas
 @panda_connect_and_init(full_reset=False)
-def test_a_recover(p):
+def test_a_known_bootstub(p):
```
adeebshihadeh, line 17: new test should go in its own PR before this one

adeebshihadeh: also can you add a description for the test?

tests/automated/1_program.py
```diff
@@ -13,13 +14,40 @@ def check_signature(p):
...
+
+  p.reset(enter_bootstub=True)
+  p.reset(enter_bootloader=True)
+
+  t_start = time.time()
+  while len(PandaDFU.list()) == 0:
+    print("waiting for DFU...")
+    time.sleep(0.1)
+    if timeout is not None and (time.time() - t_start) > timeout:
+      return False
```
adeebshihadeh, line 30: same loop is in Panda.recover, perhaps pull it out into a helper

tests/automated/1_program.py
```diff
@@ -13,13 +14,40 @@ def check_signature(p):
...
+  timeout = 30
+
+  p.reset(enter_bootstub=True)
+  p.reset(enter_bootloader=True)
+
+  t_start = time.time()
+  while len(PandaDFU.list()) == 0:
+    print("waiting for DFU...")
+    time.sleep(0.1)
+    if timeout is not None and (time.time() - t_start) > timeout:
```
adeebshihadeh, line 29: time.time can jump
```suggestion
    if timeout is not None and (time.monotonic() - t_start) > timeout:
```

tests/automated/1_program.py
```diff
@@ -13,13 +14,40 @@ def check_signature(p):
...
+      return False
+
+  dfu = PandaDFU(PandaDFU.st_serial_to_dfu_serial(p._serial, p._mcu_type))
```
adeebshihadeh, line 32: since we already have the loop, we should wait for the specific serial

crypto/sign.py
```diff
@@ -18,6 +19,8 @@
 
 with open(sys.argv[2], "wb") as f:
   if os.getenv("SETLEN") is not None:
+    # add mcu id to signature
```
adeebshihadeh, line 22: hmm, this stuff doesn't really seem like it belongs here

board/bootstub.c
```diff
@@ -56,9 +56,9 @@ int main(void) {
   SHA_hash(&_app_start[1], len-4, digest);
 
   // verify version, last bytes in the signed area
-  uint32_t vers[2] = {0};
+  uint32_t vers[3] = {0};
   memcpy(&vers, ((void*)&_app_start[0]) + len - sizeof(vers), sizeof(vers));
-  if (vers[0] != VERS_TAG || vers[1] < MIN_VERSION) {
+  if (vers[0] != MCU_IDCODE || vers[1] != VERS_TAG || vers[2] < MIN_VERSION) {
```
adeebshihadeh, line 61: we should make a test for each of these cases, e.g. flash pedal fw to black panda, F4 fw to a red panda, etc.

tests/automated/1_program.py
adeebshihadeh: the panda python lib is a pretty high level library, but we don't consistently treat it that way

tests/automated/1_program.py
adeebshihadeh: seems like a nice to do in Panda.recover as well :)

tests/automated/1_program.py
briskspirit: agree about panda lib but don't understand how it is related to test. I am selecting known good bootstub file and not passing it to panda lib but flashing it.

tests/automated/1_program.py
briskspirit: added description, will split PR

board/bootstub.c
briskspirit: added test for flashing F4 with H7 app and H7 with F4.

briskspirit: added flashing with F2 for both

adeebshihadeh:
With the 3X, we've had to make pandad reflash the bootstub in lots of different cases. Don't think this is worth it anymore.

PR #1025, opened 2022-08-11
safety: common cruise checks

sshane:
On the way to making the safety modes more like openpilot (parsers), per recent conversation. Reducing the setting of controls_allowed in each car-specific mode also seems like a good move.

Hyundai and Honda add special logic to this check, so those aren't moved over yet.

### DIFF
--- a/board/safety.h
+++ b/board/safety.h
@@ -519,13 +519,13 @@ bool steer_torque_cmd_checks(int desired_torque, int steer_req, const SteeringLi
   return violation;
 }
 
-void pcm_cruise_check(int cruise_engaged) {
+void pcm_cruise_check(bool cruise_engaged) {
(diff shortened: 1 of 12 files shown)


board/safety.h
```diff
@@ -518,3 +518,14 @@ bool steer_torque_cmd_checks(int desired_torque, int steer_req, const SteeringLi
 
   return violation;
 }
+
+void pcm_cruise_check(int cruise_engaged) {
```
sshane, line 522: We can also create a global pcm_cruise boolean, and just check this in generic_rx_checks, even more similar to openpilot's carstate.

adeebshihadeh: this is a good start. we can see what makes sense after doing these obvious, copy/paste cleanups

PR #1026, opened 2022-08-12
GM: same regen check as openpilot

sshane:
Needed for Bolt PR

- Checks entire `EBCMRegenPaddle->RegenPaddle` signal like openpilot, not just one bit
- Create common `regen_braking` so we don't have to set controls_allowed manually


board/safety/safety_gm.h
```diff
@@ -85,19 +86,16 @@ static int gm_rx_hook(CANPacket_t *to_push) {
     if (addr == 241) {
       // Brake pedal's potentiometer returns near-zero reading
       // even when pedal is not pressed
-      brake_pressed = GET_BYTE(to_push, 1) >= 10U;
+      brake_pressed = (GET_BYTE(to_push, 1) >= 10U) || (gm_regen_braking != 0);
```
sshane, line 89: In test_models we do a brake_pressed comparison, and in GM's CarState it ORs brake and regen together. We were unlucky to never test any segments with regen braking, so we never hit a mismatch.

sshane:
Hmm, there can still be a mismatch as we're using two different messages with two different rates.

Here's a falling edge with CS.brakePressed on the left and panda on the right. We likely get a regen message with no regen braking, but we haven't gotten a normal brake message to update brake_pressed yet:

```
True True
True True
True True
True True
True True
True True
False True
False False
False False
False False
False False
```

board/safety/safety_gm.h
```diff
@@ -85,19 +86,16 @@ static int gm_rx_hook(CANPacket_t *to_push) {
...
     }
 
-    // exit controls on regen paddle
+    // Regen braking is braking
     if (addr == 189) {
-      bool regen = GET_BYTE(to_push, 0) & 0x20U;
-      if (regen) {
-        controls_allowed = 0;
-      }
+      gm_regen_braking = GET_BYTE(to_push, 0) >> 4;
```
adeebshihadeh, line 98: you can just make a global regen_braking and add it to the check in generic_rx_checks

sshane:
Look good now @adeebshihadeh? Chose to allow regen braking at a standstill we do in openpilot

adeebshihadeh (commented):
LGTM, but let's write tests before merging.

tests/safety/common.py
```diff
@@ -584,3 +584,57 @@ def test_tx_hook_on_wrong_safety_mode(self):
...
+
+  def test_prev_regen(self):
+    self.assertFalse(self.safety.get_regen_braking_prev())
+    for regen in [True, False]:
+      self._rx(self._user_regen_msg(not regen))
+      self.assertEqual(not regen, self.safety.get_regen_braking_prev())
+      self._rx(self._user_regen_msg(regen))
+      self.assertEqual(regen, self.safety.get_regen_braking_prev())
+
+  def test_allow_regen_at_zero_speed(self):
```
adeebshihadeh, line 611: re-use the brake test code instead of copy/pasting it

tests/safety/common.py
```diff
@@ -474,13 +478,19 @@ def test_alternative_experience_no_disengage_on_gas(self):
...
 
-  def test_prev_brake(self):
-    self.assertFalse(self.safety.get_brake_pressed_prev())
-    for pressed in [True, False]:
-      self._rx(self._user_brake_msg(not pressed))
-      self.assertEqual(not pressed, self.safety.get_brake_pressed_prev())
-      self._rx(self._user_brake_msg(pressed))
-      self.assertEqual(pressed, self.safety.get_brake_pressed_prev())
+  def test_prev_brake(self, test="brake"):
+    def test_prev_brake(_user_brake_msg, get_brake_pressed_prev):
```
adeebshihadeh, line 482: these have the same name

tests/safety/test_gm.py
```diff
@@ -90,6 +94,18 @@ def _torque_cmd_msg(self, torque, steer_req=1):
     values = {"LKASteeringCmd": torque}
     return self.packer.make_can_msg_panda("ASCMLKASteeringCmd", 0, values)
 
+  def test_prev_brake(self, test="brake"):
+    super().test_prev_brake(test="brake")
+    super().test_prev_brake(test="regen")
```
adeebshihadeh, line 99: these tests should be common

tests/safety/common.py
```diff
@@ -474,13 +478,19 @@ def test_alternative_experience_no_disengage_on_gas(self):
...
+      self.assertFalse(self.safety.get_brake_pressed_prev())
+      for pressed in [True, False]:
+        self._rx(_user_brake_msg(not pressed))
+        self.assertEqual(not pressed, get_brake_pressed_prev())
+        self._rx(_user_brake_msg(pressed))
+        self.assertEqual(pressed, get_brake_pressed_prev())
+
+    if test == "brake":
+      test_prev_brake(self._user_brake_msg, self.safety.get_brake_pressed_prev)
+    else:  # test with regen message and global variable
```
adeebshihadeh, line 492: this comment shouldn't be necessary. it would be more self-explanatory if the function had a better name

tests/safety/test_gm.py
sshane: What do you want `_user_regen_msg` to return for the other cars?

sshane: Maybe that should be split up so the common test tests behavior using the state of the variable, and then there's a car-specific test for GM that tests that the CAN message affects the state of the variable.

This sounds like a good clean up for all of the tests

sshane: This now tests all previous brake tests for all safety models, and only the regen tests for safety models that implement `_user_regen_msg` (it's impossible to test anything without it)

sshane: Just made it detect the definition of the regen message, as there's nothing to test if the safety model test does not use it

tests/safety/common.py
```diff
@@ -542,13 +545,17 @@ def test_alternative_experience_no_disengage_on_gas(self):
     self._rx(self._user_gas_msg(0))
     self.assertTrue(self.safety.get_longitudinal_allowed())
 
-  def test_prev_brake(self):
-    self.assertFalse(self.safety.get_brake_pressed_prev())
+  def test_prev_brake(self, user_brake_msg=None, get_brake_pressed_prev=None):
+    if user_brake_msg is None or get_brake_pressed_prev is None:
```
sshane, line 549:
```suggestion
    if user_brake_msg is None:
```

PR #1031, opened 2022-08-14
Hyundai CAN-FD: support for ICE cars

sunnyhaibin:
**Requires**
- https://github.com/commaai/panda/pull/1106

**Prerequisite of**
- https://github.com/commaai/openpilot/pull/25434


python/__init__.py
```diff
@@ -199,6 +199,7 @@ class Panda:
 
   FLAG_HYUNDAI_CANFD_HDA2 = 1
   FLAG_HYUNDAI_CANFD_ALT_BUTTONS = 2
+  FLAG_HYUNDAI_CANFD_ALT_PEDALS = 4
```
adeebshihadeh, line 202: let's match what we do for the normal hyundai safety mode: https://github.com/commaai/panda/blob/1776165a3d902d6320965b6b6b1715bb9a25915b/python/__init__.py#L196

board/safety/safety_hyundai_canfd.h
```diff
@@ -26,9 +26,10 @@ const CanMsg HYUNDAI_CANFD_HDA1_TX_MSGS[] = {
 
 AddrCheckStruct hyundai_canfd_addr_checks[] = {
   {.msg = {{0x35, 1, 32, .check_checksum = true, .max_counter = 0xffU, .expected_timestep = 10000U},
-           {0x105, 0, 32, .check_checksum = true, .max_counter = 0xffU, .expected_timestep = 10000U}, { 0 }}},
-  {.msg = {{0x65, 1, 32, .check_checksum = true, .max_counter = 0xffU, .expected_timestep = 10000U},
-           {0x65, 0, 32, .check_checksum = true, .max_counter = 0xffU, .expected_timestep = 10000U}, { 0 }}},
+           {0x105, 0, 32, .check_checksum = true, .max_counter = 0xffU, .expected_timestep = 10000U},
+           {0x100, 0, 32, .check_checksum = true, .max_counter = 0xffU, .expected_timestep = 10000U}}},
+  //{.msg = {{0x65, 1, 32, .check_checksum = true, .max_counter = 0xffU, .expected_timestep = 10000U},
+  //         {0x65, 0, 32, .check_checksum = true, .max_counter = 0xffU, .expected_timestep = 10000U}, { 0 }}},
```
adeebshihadeh, line 32: you can just make another struct for this, but have you checked for a common brake pressed yet? normal hyundai has one. I can check our EV6

adeebshihadeh (changes requested):

board/safety/safety_hyundai_canfd.h
sunnyhaibin: I've looked through all the messages, 0x100 is the one that the ICE versions use. I'll make a struct specifically for ICE CAN-FD and we can refine at a later time :)

sshane:
@sunnyhaibin going to remove all longitudinal related stuff and split some stuff out. feel free to make a copy of your branch at this commit to retain your long changes for a future PR

PR #1037, opened 2022-08-18
Subaru longitudinal safety

martinl:
Minimal panda safety implementation with required messages for Subaru Global longitudinal support


adeebshihadeh (changes requested):
Can you add a flag for Subaru long? See Hyundai as an example

We can merge this after:
- [ ] safety flag for Subaru long, gated behind `ALLOW_DEBUG` https://github.com/commaai/panda/blob/36b317185cd1ece71025a1551c178cc1317602d4/board/safety/safety_hyundai.h#L371
- [ ] separate set of TX msgs https://github.com/commaai/panda/blob/36b317185cd1ece71025a1551c178cc1317602d4/board/safety/safety_hyundai.h#L23
- [ ] safety tests for TX msgs https://github.com/commaai/panda/blob/36b317185cd1ece71025a1551c178cc1317602d4/tests/safety/test_hyundai.py#L240
- [ ] check flag in fwd hook

martinl:
I'll update the openpilot branch also, this should make it easier to debug

adeebshihadeh:
Looks good. Anything else you want to do before merging?

martinl:
I updated the longitudinal signal checks a bit and I think it should be ok to merge. 

Next step would be figuring out how to check the acceleration limits - maybe using G_Sensor Longitudinal or tracking the speed signal changes over some time?

martinl:
I updated the long safety to check the three long signals min and max values and added the safety tests. It is tested and working.

I'm currently looking into finding AEB/PCB status signal but have not found a good candidate.  For example LKAS_Alert = 2 (FCW_Repeated_Beep) also triggers braking in addition to multiple PCB dash alerts. Since stock ACC is inactive when openpilot long is used then I think it would make sense to pass through AEB (ES_Brake Brake_Pressure) in all cases when ES_Brake Cruise_Brake_Active becomes high in bus 2 messages instead of checking multiple dash alert values.

martinl:
subaru_aeb is using now ES_LKAS_State LKAS_Alert signals to match openpilot stockAeb. I also updated longitudinal checks to apply only when controls are allowed. Turns out Eyesight is generating longitudinal output signals also when idle and they sometimes exceed openpilot limits so it's better to pass them through as is.

martinl:
Superseded by https://github.com/commaai/panda/pull/1500 - closing

PR #1039, opened 2022-08-23
GM Safety: Add support for auto-resume via button spamming

JasonJShuler:
Resume spamming needs to be on PT bus rather than camera bus used for cancel button spamming.

Updated the TX allowed messages to permit the button spamming on the PT bus for all configurations

Updated the TX safety code to allow the resume button if cruise is off. Cancel functionality should be unchanged


JasonJShuler:
I'm gonna just close this one myself...

PR #1041, opened 2022-08-24
CI: add hardware reflasher

briskspirit:
Will reset all pandas in CI with DFU flash

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -4,7 +4,7 @@ pipeline {
     DOCKER_IMAGE_TAG = "panda:build-${env.GIT_COMMIT}"
   }
   stages {
-    stage ('check locks') {
+    stage ('Acquire resource locks') {
       options {
         lock(resource: "pandas")
       }
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -49,7 +49,7 @@ pipeline {
         }
         stage('HITL tests') {
           steps {
-            timeout(time: 20, unit: 'MINUTES') {
+            timeout(time: 30, unit: 'MINUTES') {
               script {
                 sh "docker run --rm --privileged \
                       --volume /dev/bus/usb:/dev/bus/usb \
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -14,6 +14,22 @@ pipeline {
         }
       }
     }
+    stage('reset hardware') {
+      steps {
+        lock(extra: [[resource: "pedal"]], resource: "pandas") {
+          timeout(time: 10, unit: 'MINUTES') {
+            script {
+              sh "docker run --rm --privileged \
+                    --volume /dev/bus/usb:/dev/bus/usb \
+                    --volume /var/run/dbus:/var/run/dbus \
+                    --net host \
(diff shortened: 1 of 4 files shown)


Jenkinsfile
```diff
@@ -14,6 +14,22 @@ pipeline {
         }
       }
     }
+    stage('reset hardware') {
```
adeebshihadeh, line 17: need to lock all the resources and do everything at once now

tests/ci_resetter.py
```diff
@@ -0,0 +1,85 @@
...
+    self.enable_power(3, False)
+    time.sleep(1)
+    self.enable_power(1, True)
+    self.enable_power(2, True)
+    self.enable_power(3, True)
+    if delay > 0:
+      time.sleep(delay)
+
+if __name__ == "__main__":
+  r = Resetter()
```
adeebshihadeh, line 58: this part should be split out from the file with the class. there should be a file for the support lib and another to get the CI zoo in a good state. I would also move things like the test jungle and panda serials in the prep script so they're all in one place

tests/ci_resetter.py
```diff
@@ -0,0 +1,85 @@
...
+  print(pandas)
+  assert len(pandas) == 7
+
+  for serial in pandas:
+    p = PandaDFU(serial)
+    p.recover()
+
+  r.cycle_power(5)
+
+  pandas = [x for x in Panda.list() if x != 'none'] # Exclude pedal
```
adeebshihadeh, line 74: should we just exclude pedals by default in Panda.list()? this none thing is a pretty bad convention

Jenkinsfile
briskspirit: I am already locking those with extra, no?

tests/ci_resetter.py
briskspirit: agree, add here or separate PR?

Jenkinsfile
briskspirit: `lock(extra: [[resource: "pedal"]], resource: "pandas") {`

adeebshihadeh: that's only for the reset stage. the rest of the stages should all be under one lock now

tests/ci_resetter.py
adeebshihadeh: another PR

tests/ci_reset_hw.py
```diff
@@ -0,0 +1,31 @@
+from panda import Panda, PandaDFU
+from libs.resetter import Resetter  # pylint: disable=import-error
+
+r = Resetter()
```
adeebshihadeh, line 4: always guard with `if __name__ == '__main__':`. you don't want this to run at import time

tests/ci_reset_hw.py
```diff
@@ -0,0 +1,31 @@
+from panda import Panda, PandaDFU
+from libs.resetter import Resetter  # pylint: disable=import-error
```
adeebshihadeh, line 2: shouldn't need a pylint exception here

briskspirit: pre-commit hook and pylint tells that I need it

PR #1042, opened 2022-08-24
python lib: check serial length for Panda.list()

briskspirit:
Serial length for real panda is always == 24

### DIFF
--- a/python/__init__.py
+++ b/python/__init__.py
@@ -410,7 +410,9 @@ class Panda:
       for device in context.getDeviceList(skip_on_error=True):
         if device.getVendorID() == 0xbbaa and device.getProductID() in (0xddcc, 0xddee):
           try:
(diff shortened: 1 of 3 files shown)


python/__init__.py
```diff
@@ -410,7 +410,9 @@ def list():
       for device in context.getDeviceList(skip_on_error=True):
         if device.getVendorID() == 0xbbaa and device.getProductID() in (0xddcc, 0xddee):
           try:
-            ret.append(device.getSerialNumber())
+            serial = device.getSerialNumber()
+            if len(serial) == 24:
```
adeebshihadeh, line 414: would probably print a warning here https://github.com/commaai/openpilot/blob/70f7340c173f0a4b0d3bb98dda9032cd304f890e/tools/lib/logreader.py#L101

PR #1044, opened 2022-08-24
debug build: remove direct jump to DFU

briskspirit:
For release build to go into DFU mode we need to jump to bootstub first.
I see no reason for this, just adds complexity.
Also we might want to check if safety is set to car mode before jumping to either bootloader or softloader.

### DIFF
--- a/board/main.c
+++ b/board/main.c
@@ -38,15 +38,18 @@ void debug_ring_callback(uart_ring *ring) {
   while (getc(ring, &rcv)) {
     (void)putc(ring, rcv);  // misra-c2012-17.7: cast to void is ok: debug function
 
-    // jump to DFU flash
-    if (rcv == 'z') {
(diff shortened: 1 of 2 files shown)


adeebshihadeh:
This is done intentionally. Why do you want to remove it?

briskspirit:
We still can enter DFU on release through bootstub, so what's the point then?

jyoung8607:
> Also we might want to check if safety is set to car mode before jumping to either bootloader or softloader.

This is both mandatory for SIL2 and still shown as open (#369). Not sure if it's connected to the hop-skip-jump requirement for DFU, but it's either that or protection from flashing unsigned or differently-signed firmware.

robbederks:
Was thinking of doing the opposite: only allowing to enter DFU from the bootstub, never from the application code. Right now you can still enter DFU directly when on a non-release build, which adds confusion while testing. Think we should always force you to go through the bootstub.

PR #1045, opened 2022-08-26
Toyota: match openpilot standstill check

sshane:


sshane:
@adeebshihadeh should be safe, lowest wheel speeds ever reported were 0.01 kph which is 1 here.

board/safety/safety_toyota.h
```diff
@@ -106,7 +104,7 @@ static int toyota_rx_hook(CANPacket_t *to_push) {
         int wheel_speed = (GET_BYTE(to_push, i) << 8U) + GET_BYTE(to_push, (i+1U));
         speed += wheel_speed - 0x1a6f;
       }
-      vehicle_moving = ABS(speed / 4) > TOYOTA_STANDSTILL_THRSLD;
+      vehicle_moving = ABS(speed) > 0;
```
adeebshihadeh, line 107: no need to add them up anymore, just check this in the for loop

board/safety/safety_toyota.h
```diff
@@ -98,15 +96,13 @@ static int toyota_rx_hook(CANPacket_t *to_push) {
...
 
-    // sample speed
     if (addr == 0xaa) {
-      int speed = 0;
-      // sum 4 wheel speeds
-      for (uint8_t i=0U; i<8U; i+=2U) {
-        int wheel_speed = (GET_BYTE(to_push, i) << 8U) + GET_BYTE(to_push, (i+1U));
-        speed += wheel_speed - 0x1a6f;
+      // check any wheel speed for non-zero value
+      vehicle_moving = false;
```
adeebshihadeh, line 101: make sure this is done atomically

board/safety/safety_toyota.h
```diff
@@ -98,15 +96,13 @@ static int toyota_rx_hook(CANPacket_t *to_push) {
...
+      for (uint8_t i = 0U; i < 8U; i += 2U) {
+        int wheel_speed = (GET_BYTE(to_push, i) << 8U) + GET_BYTE(to_push, (i + 1U));
+        vehicle_moving |= wheel_speed != 0x1a6f;
```
adeebshihadeh, line 104: use a logical or

adeebshihadeh (changes requested):

board/safety/safety_toyota.h
sshane: Tried to just use `GET_BYTES_04`, but it weirdly returned `0x6f1a6f00` for 0 speed, not `0x6f1a6f1a`.

Issue #843, opened 2022-01-28
Block actuation while in the pre-enable state


adeebshihadeh:
Currently, only the Honda safety handles this.

adeebshihadeh:
@sshane is this done?

sshane:
I believe the topic of this PR was to enforce pre-enable safety when enabling with gas pressed, however we now allow steering in pre-enable state and longitudinal pre-enabling is done with the common checks in safety.h. Relevant PRs: https://github.com/commaai/panda/pull/884 and https://github.com/commaai/panda/pull/1121

We still need to block actuation when pre-enabling at a stop in is this draft PR which I'm trying to clean up: https://github.com/commaai/panda/pull/1123

PR #1046, opened 2022-08-27
refactor torque measurement reset on violation

ntegan1:
* removes some duplication
* allows for cleaner custom steer request mismatch handling (useful when updating toyota-steer-faults2)


adeebshihadeh:
This isn't necessary; GM will use the common steer checks once #843 is done.

PR #1050, opened 2022-08-29
python lib: cleanup hw type check

briskspirit:
Think it was old plan to remove those


python/__init__.py
```diff
@@ -530,13 +511,7 @@ def get_mcu_type(self):
     return None
 
   def has_obd(self):
-    return (self.is_uno() or self.is_dos() or self.is_black() or self.is_red())
-
-  def has_canfd(self) -> bool:
-    return self.get_type() in Panda.H7_DEVICES
-
-  def is_internal(self) -> bool:
```
sshane, line 538: pandad.py uses is_internal, so I can't bump panda on master, can I revert?

briskspirit: done

PR #1051, opened 2022-08-31
USB power mode + bootkick cleanup

adeebshihadeh:


board/main.c
```diff
@@ -401,7 +380,7 @@ int main(void) {
       #ifdef DEBUG_FAULTS
       if(fault_status == FAULT_STATUS_NONE){
       #endif
-        uint32_t div_mode = ((usb_power_mode == USB_POWER_DCP) ? 4U : 1U);
+        const uint32_t div_mode = 1U;
```
robbederks, line 383: want to just get rid of this?

board/boards/uno.h
```diff
@@ -145,9 +126,11 @@ void uno_set_can_mode(uint8_t mode){
...
 }
 
-void uno_usb_power_mode_tick(uint32_t uptime){
-  UNUSED(uptime);
-  if(bootkick_timer != 0U){
+void uno_board_tick(bool ignition, bool usb_enum, bool heartbeat_seen) {
+  UNUSED(ignition);
+  UNUSED(usb_enum);
+  UNUSED(heartbeat_seen);
+  if (bootkick_timer != 0U) {
```
robbederks, line 133: might want to increase `BOOTKICK_TIME` since the tick is now called at 8Hz instead of 1

adeebshihadeh: moved to the 1Hz loop

PR #1052, opened 2022-08-31
Ford safety updates

incognitojam:
- Changed the message name style to match the DBC
- Button press checks
  - Allow sending the resume button press while openpilot controls are allowed
  - Also allow sending various other signals at any time (so that we can forward signals e.g. blinkers) which are not safety relevant
- Added ACCDATA_3 message which we use for the "lane centering" UI (it was blocked before, that's why it didn't show up on the dash!)

Checksums:
- `EngBrakeData`
  - There are no checksums/counters
- `EngVehicleSpThrottle`
  - There are signals for a checksum/counter on the gas pedal position signal, but they look like they're unused
- `DesiredTorqBrk`
  - There are no checksums/counters


board/safety/safety_ford.h
```diff
@@ -82,16 +82,16 @@ static int ford_tx_hook(CANPacket_t *to_send, bool longitudinal_allowed) {
...
+  // Cruise button check to only allow cancel button press when controls are off
+  if ((addr == MSG_STEERING_DATA_FD1) && !controls_allowed) {
+    // Violation if any set/resume button is pressed while controls not allowed
+    //   CcAslButtnCnclPress is 0x01 in byte 1 (not violation)
+    //   CcAsllButtnResPress is 0x02 in byte 3 (violation)
+    //   TjaButtnOnOffPress is 0x01 in byte 5 (not violation)
+    bool violation = ((GET_BYTE(to_send, 1) & 0x06U) |
+                      (GET_BYTE(to_send, 2) & 0x21U) |
+                      (GET_BYTE(to_send, 3) & 0xFEU) |
+                      (GET_BYTE(to_send, 4) & 0xE0U)) != 0U;
```
sshane, line 94: If they're 1 bit signals, you can use GET_BIT, with the msb from the DBC

incognitojam: decided to just check the signals that we send in OP instead

PR #1054, opened 2022-08-31
Hyundai: enforce ACC main state with controls allowed

sunnyhaibin:
Enforce ACC main state with stock/openpilot longitudinal control in openpilot and panda safety. Enforce `controls_allowed = 0` when the signal is `0`.

Stock ACC: `SCC11|MainMode_ACC`

openpilot ACC: `EMS16|CRUISE_LAMP_M`
- Available on all currently supported openpilot longitudinal HKG cars
- Sync with ACC MAIN button press
- Will be copied and sent with `SCC11|MainMode_ACC` to the car when ACC MAIN is pressed

**Prerequisite of:**
- https://github.com/commaai/openpilot/pull/25624


sunnyhaibin:
> Upon testing on some older models, the signal does not do anything when the CRUISE MAIN button is pressed. Need to investigate and see if there is another common signal to use for those cars.
* https://github.com/commaai/openpilot/pull/25624#issuecomment-1264820144

PR #1055, opened 2022-08-31
Hyundai CAN-FD: common brake pressed signal

adeebshihadeh:


sshane:
duplicate? https://github.com/commaai/panda/pull/1113

PR #1056, opened 2022-09-01
Hyundai: Car Port for Kia EV6 (HDA1) 2022

hoomoose:
**Prerequisite for:**
[#25629 #Kia: Car Port for EV6 (HDA1) 2022](https://github.com/commaai/openpilot/pull/25629)

**Need Help**
1. I could not figure out how to add multiple 0x35 signal checks like you do for 0x1cf without getting errors
2. Resume from stop is not working for my HDA1 EV6; does 0x1cf need to be added to forward hook since it is on bus0 instead of bus1?


hoomoose:
No longer needed

PR #1014, opened 2022-08-09
GM: longitudinal support for camera-integrated cars

JasonJShuler:
Add safetyparam and code to enable OP longitudinal on GM vehicles with factory VOACC.

Important: Contains code from PR #962 - There are actually just a handful of changes. Once the PR is merged this can be refreshed from master and it should simmer down to the actual changes.

On cars with camera-based ACC, many users have found OP's VOACC to be superior; additionally without OP long, e2e is not possible.

This provides a parameter that will cause the Panda to block ACC messages from the camera and adjust various conditions (disable controls if ACC messages are present on PT bus, use buttons for enable / disable / speed adjust; allow tx of ACC messages and brake command on PT bus) that enable OP to source the ACC.

The parameter is implemented as a 3rd hardware type for simplicity, but (for now) it only apples to cars with stock VOACC.

PR #1057, opened 2022-09-01
GM camera ACC: long safety

sshane:


sshane:
Will merge into https://github.com/commaai/panda/pull/1014 later

PR #1065, opened 2022-09-13
Chrysler: match openpilot standstill check

sshane:


board/safety/safety_chrysler.h
```diff
@@ -206,14 +203,12 @@ static int chrysler_rx_hook(CANPacket_t *to_push) {
     // TODO: use the same message for both
     // update speed
     if ((chrysler_platform != CHRYSLER_PACIFICA) && (bus == 0) && (addr == chrysler_addrs->ESP_8)) {
-      vehicle_speed = (((GET_BYTE(to_push, 4) & 0x3U) << 8) + GET_BYTE(to_push, 5))*0.0078125;
-      vehicle_moving = (int)vehicle_speed > CHRYSLER_RAM_STANDSTILL_THRSLD;
+      vehicle_moving = ((GET_BYTE(to_push, 4) << 8) + GET_BYTE(to_push, 5)) != 0U;
```
sshane, line 206: @realfast any reason we were only parsing the first two bits of the first byte of the signal? Made it parse the whole two bytes now

PR #1076, opened 2022-09-21
examples: choose panda for fw query

incognitojam:
improves the car port experience for those needing multiple pandas/external red panda

https://discord.com/channels/469524606043160576/539096103468007424/1022224299093921913

### DIFF
--- a/python/__init__.py
+++ b/python/__init__.py
@@ -183,17 +183,6 @@ class Panda:
   HW_TYPE_RED_PANDA_V2 = b'\x08'
   HW_TYPE_TRES = b'\x09'
 
-  HW_TYPES = {
-    HW_TYPE_WHITE_PANDA: "white",
-    HW_TYPE_GREY_PANDA: "grey",
-    HW_TYPE_BLACK_PANDA: "black",
-    HW_TYPE_PEDAL: "pedal",
-    HW_TYPE_UNO: "uno",
-    HW_TYPE_DOS: "dos",
-    HW_TYPE_RED_PANDA: "red",
-    HW_TYPE_RED_PANDA_V2: "red v2",
-  }
-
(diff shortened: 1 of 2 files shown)


incognitojam:
maybe it should translate `--bus 4` and others to the correct panda?

python/__init__.py
```diff
@@ -541,6 +541,27 @@ def get_type(self):
 
     return ret
 
+  def get_type_name(self):
+    hw_type = self.get_type()
+    if hw_type == Panda.HW_TYPE_WHITE_PANDA:
```
adeebshihadeh, line 546: use a dict

incognitojam: I tried but it said it was unhashable

incognitojam: ah I have an idea

incognitojam:
better to improve fw_versions.py in openpilot instead  https://github.com/commaai/openpilot/blob/master/selfdrive/car/fw_versions.py

sshane:
query_fw_versions.py is useful for debugging, would this fix red panda querying?

python/__init__.py
```diff
@@ -183,6 +183,18 @@ class Panda:
   HW_TYPE_RED_PANDA_V2 = b'\x08'
   HW_TYPE_TRES = b'\x09'
 
+  HW_TYPES = [HW_TYPE_WHITE_PANDA, HW_TYPE_GREY_PANDA, HW_TYPE_BLACK_PANDA, HW_TYPE_PEDAL, HW_TYPE_UNO, HW_TYPE_DOS, HW_TYPE_RED_PANDA, HW_TYPE_RED_PANDA_V2]
```
sshane, line 186:
```suggestion
```

PR #1079, opened 2022-09-28
IsoTpMessage: param for single frame flow control

sshane:
Before the fix in https://github.com/commaai/panda/pull/1078, we'd time out on certain ecus that take longer than 100 ms to finish a full FW query response. Then, when a brand has multiple requests with no ecu whitelists and we'd receive frames from the previous request, throwing invalid consecutive frame errors.

Extending the timeouts on each received frame fixed most of those issues, however the 10 ms delay we have in the flow control message is a minimum, and it's quite possible that an ECU takes 100+ ms to send a frame and we'd time out on it.

This PR removes the frame interval we added to fix certain Hyundai ECUs and replaces it by sending the block size parameter. `IsoTpMessage` now needs to send a flow control message after each consecutive frame it gets from the ECU. This way, if we time out on an ECU, it won't send any more frames on its own.

### DIFF
--- a/python/uds.py
+++ b/python/uds.py
@@ -480,7 +480,7 @@ class IsoTpMessage():
         self.rx_done = True
       elif self.single_frame_mode:
         # notify ECU to send next frame
-        msg = b"\x30\x01\x00".ljust(self.max_len, b"\x00")
+        msg = b"\x30\x01\x0a".ljust(self.max_len, b"\x00")
         self._can_client.send([msg])
       if self.debug:
         print(f"ISO-TP: RX - consecutive frame - {hex(self._can_client.rx_addr)} idx={self.rx_idx} done={self.rx_done}")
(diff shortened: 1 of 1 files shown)


sshane:
Requested for review of how I did this, but will test on cars first before merging

gregjhogan (changes requested):
I don't like hard-coding it to be slow like this, lets add a parameter to `IsoTpMessage` constructor called something like `single_frame_mode` (for something easy) and default to the current mode

gregjhogan:
> Also, this may not be perfect, as in that one instance where a frame takes longer than our timeout to respond, it would still respond eventually. The difference is we would get one extra unwanted frame after timing out, not the rest of the response.

yeah I think the only sane way to try to cover all these edge cases is just retry the whole fingerprinting process, ideally locking into a brand if we received any exact fw version matches and rolling forward/skipping any addrs we received an exact fw version match on.

sshane:
This PR strictly adds an option for enabling single-frame mode, not changes behavior

Issue #916, opened 2022-04-14
add support for CAN-FD 1.0


gregjhogan:
currently we only support CAN-FD 2.0 (which is the ISO standard) but the red panda can support CAN-FD 1.0 which is used by Hyundai Palisade blind spot radars.

enabled by setting NISO bit in FDCAN_CCCR register (see https://github.com/gregjhogan/panda/commit/f801919dd858a0935e502f2ce7a4c0585f2e4478)

adeebshihadeh:
If you're in a non-car safety mode, I think it's fine to switch between ISO and non-ISO CAN FD until valid messages are seen.

PR #1082, opened 2022-10-01
CAN-FD non-ISO support

gregjhogan:
add support #916

### DIFF
--- a/board/stm32h7/llfdcan.h
+++ b/board/stm32h7/llfdcan.h
@@ -84,7 +84,7 @@ bool fdcan_exit_init(FDCAN_GlobalTypeDef *CANx) {
   return ret;
 }
 
-bool llcan_set_speed(FDCAN_GlobalTypeDef *CANx, uint32_t speed, uint32_t data_speed, bool loopback, bool silent) {
+bool llcan_set_speed(FDCAN_GlobalTypeDef *CANx, uint32_t speed, uint32_t data_speed, bool non_iso, bool loopback, bool silent) {
   UNUSED(speed);
   bool ret = fdcan_request_init(CANx);
 
@@ -97,6 +97,7 @@ bool llcan_set_speed(FDCAN_GlobalTypeDef *CANx, uint32_t speed, uint32_t data_sp
     CANx->TEST &= ~(FDCAN_TEST_LBCK);
     CANx->CCCR &= ~(FDCAN_CCCR_MON);
     CANx->CCCR &= ~(FDCAN_CCCR_ASM);
+    CANx->CCCR &= ~(FDCAN_CCCR_NISO);
 
     // TODO: add as a separate safety mode
     // Enable ASM restricted operation(for debug or automatic bitrate switching)
@@ -129,6 +130,11 @@ bool llcan_set_speed(FDCAN_GlobalTypeDef *CANx, uint32_t speed, uint32_t data_sp
     sjw = MIN(15U, seg2);
 
     CANx->DBTP = (((sjw & 0xFU)-1U)<<FDCAN_DBTP_DSJW_Pos) | (((seg1 & 0x1FU)-1U)<<FDCAN_DBTP_DTSEG1_Pos) | (((seg2 & 0xFU)-1U)<<FDCAN_DBTP_DTSEG2_Pos) | (((prescaler & 0x1FU)-1U)<<FDCAN_DBTP_DBRP_Pos);
+
+    if (non_iso) {
+      // FD non-ISO mode
+      CANx->CCCR |= FDCAN_CCCR_NISO;
+    }
 
     // Silent loopback is known as internal loopback in the docs
     if (loopback) {
--- a/python/__init__.py
+++ b/python/__init__.py
@@ -186,9 +186,9 @@ class Panda:
 
   CAN_PACKET_VERSION = 2
   HEALTH_PACKET_VERSION = 11
-  CAN_HEALTH_PACKET_VERSION = 2
+  CAN_HEALTH_PACKET_VERSION = 3
   HEALTH_STRUCT = struct.Struct("<IIIIIIIIIBBBBBBHBBBHfBB")
-  CAN_HEALTH_STRUCT = struct.Struct("<BIBBBBBBBBIIIIIIHHBB")
+  CAN_HEALTH_STRUCT = struct.Struct("<BIBBBBBBBBIIIIIIHHBBB")
 
   F2_DEVICES = (HW_TYPE_PEDAL, )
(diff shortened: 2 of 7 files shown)


adeebshihadeh:
It would be nice if this was auto-detected, like we already do with CAN-FD.

gregjhogan:
> It would be nice if this was auto-detected, like we already do with CAN-FD.

no way to do it based on the data in the messages like we did for non-ISO (it isn't backward or forward compatible in any way, sadly)

adeebshihadeh:
> > It would be nice if this was auto-detected, like we already do with CAN-FD.
> 
> no way to do it based on the data in the messages like we did for ISO (v2.0) mode (it isn't backward or forward compatible in any way, sadly)

We can just flip back and forth until we see a valid message though, obviously with some constraints like only do once, must be in the silent safety mode, etc.

gregjhogan:
I am not sure how to do that, and on top of that I am the only person using it now and the default speed is wrong, so I have to change the speed and may as well set this flag at the same time.

Don't we also want the explicit control?  If yes, what you are describing sounds like it should be a separate PR.

gregjhogan:
I am also not 100% certain it will work to just look for valid messages.  Sometimes I see a few messages on non-sense addresses and I am suspicious what is happening is sometimes the CAN-FD 1.0 messages parse successfully as valid CAN-FD 2.0 messages.  I don't have an example right now, but will post one if I see it happen again.  Since CAN-FD 2.0 introduced breaking changes it definitely seems possible.

adeebshihadeh:
> Don't we also want the explicit control? If yes, what you are describing sounds like it should be a separate PR.

Yes, we want manual control and this would be another PR; just bringing it up.

gregjhogan:
another problem that would need to be addressed is that boardd doesn't start in silent mode (currently it initially puts all panda devices in ELM) which causes faults if you don't have the correct mode selected, but perhaps we can change it to only put the first panda in ELM mode for fingerprinting.

board/stm32h7/llfdcan.h
```diff
@@ -150,6 +150,29 @@ bool llcan_set_speed(FDCAN_GlobalTypeDef *CANx, uint32_t speed, uint32_t data_sp
   return ret;
 }
 
+bool llcan_set_non_iso(FDCAN_GlobalTypeDef *CANx, bool non_iso) {
```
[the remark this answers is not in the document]
gregjhogan: I see `llcan_set_speed` already does more than just set the speed for both CAN and CAN-FD, so I will move setting this new option into there and someone can do a name refactor in a future PR if the name is an issue (I feel like the code would be easier to navigate for example if the CAN-FD only functions had `canfd` in the name like `llcan_set_speed` -> `llcanfd_set_speed`, for example).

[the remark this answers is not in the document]
briskspirit: Reason behind this is that I need to have the same function name for both F4 and H7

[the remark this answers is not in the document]
gregjhogan: I don't think that is true for all `ll` functions?  The `llcan_set_speed` functions each have different signatures so it isn't relying on a shared name.

python/__init__.py
```diff
@@ -624,14 +625,17 @@ def set_can_speed_kbps(self, bus, speed):
   def set_can_data_speed_kbps(self, bus, speed):
     self._handle.controlWrite(Panda.REQUEST_OUT, 0xf9, bus, int(speed * 10), b'')
 
-  # CAN FD and BRS status
+  # CAN FD, BRS, non-ISO mode status
   def get_canfd_status(self, bus):
-    dat = self._handle.controlRead(Panda.REQUEST_IN, 0xfa, bus, 0, 2)
+    dat = self._handle.controlRead(Panda.REQUEST_IN, 0xfa, bus, 0, 3)
     if dat:
```
briskspirit, line 631: Would add length check too as python lib will get exception from struct.unpack if old firmware panda will be connected.

gregjhogan: sure, can pad with zero byte and it will be backwards compatible (was always non-ISO in the past)

board/stm32h7/llfdcan.h
```diff
@@ -130,6 +130,14 @@ bool llcan_set_speed(FDCAN_GlobalTypeDef *CANx, uint32_t speed, uint32_t data_sp
 
     CANx->DBTP = (((sjw & 0xFU)-1U)<<FDCAN_DBTP_DSJW_Pos) | (((seg1 & 0x1FU)-1U)<<FDCAN_DBTP_DTSEG1_Pos) | (((seg2 & 0xFU)-1U)<<FDCAN_DBTP_DTSEG2_Pos) | (((prescaler & 0x1FU)-1U)<<FDCAN_DBTP_DBRP_Pos);
 
+    if (non_iso) {
+      // FD non-ISO mode
+      CANx->CCCR |= FDCAN_CCCR_NISO;
+    } else {
+      // FD ISO mode
+      CANx->CCCR &= ~(FDCAN_CCCR_NISO);
```
briskspirit, line 138: It's better to move FD ISO reset to beginning of llcan_set_speed() (where we "reset operation mode to Normal" to be consistent with the code.

briskspirit (approved):

python/__init__.py
adeebshihadeh: I don't think backwards compatibility is a good goal here; it introduces a lot of complexity: when should we care about backwards compatibility? how long should I keep this code around? etc. I would just add `ensure_can_packet_version` to this function.

gregjhogan: Does tying it to `ensure_can_packet_version` make sense?  That seems confusing, too, since it has no impact on what is being returned here (we didn't change that version for this PR).

Adding a new version number isn't trivial either.  If we care that much about adding `.ljust(3, b"\x00")` here I would recommend just making it a breaking change; no one is using this method anyway.

gregjhogan: we could also make it a new method with a new ID, but I didn't think it was worth the extra lines of code required.

briskspirit: Meh, then let's just break it.

adeebshihadeh: I feel strongly that priorities should be: communicating why something doesn't work > doesn't work (uncommunicated) > half works because we added some backwards compatibility. In this case, I'd probably just remove `get_canfd_status`; isn't it now just a subset of the can health stuff?

gregjhogan: Looks like it is a subset so it could be removed, but if we remove it any old scripts using it get `doesn't work (uncommunicated)` and if we instead add a length check as originally suggested you get `communicating why something doesn't work` so should we keep it and add a length check?

adeebshihadeh: realistically, no one is using this in a script. removing is the best option in this case

gregjhogan:
I think this is good to merge now

PR #1085, opened 2022-10-03
Ford: allow sending buttons msg to camera

incognitojam:
The DBC lists both the IPMA and PCM as receiving modules for the `CcAslButtnCnclPress` and `CcAsllButtnResPress` signals used to cancel/resume ACC.

Sending resume to both buses is confirmed to work by three testers, now it's just down to tweaking openpilot to get high reliability.


incognitojam:
Actually the DBC tells us that the msg should be sent to both the IPMA and PCM
https://github.com/commaai/opendbc/blob/master/ford_lincoln_base_pt.dbc#L2954

PR #1089, opened 2022-10-07
GM: change brake pressed signal

sshane:
For PR: https://github.com/commaai/openpilot/pull/25970


board/safety/safety_gm.h
```diff
@@ -89,10 +89,12 @@ static int gm_rx_hook(CANPacket_t *to_push) {
...
 
-    if (addr == 241) {
-      // Brake pedal's potentiometer returns near-zero reading
-      // even when pedal is not pressed
-      brake_pressed = GET_BYTE(to_push, 1) >= 10U;
+    if (addr == 190) {
+      // Some Volt 2016-17 have loose brake pedal push rod retainers which causes the ECM to believe
+      // that the brake is being intermittently pressed without user interaction.
+      // To avoid a cruise fault we need to match the ECM's brake pressed signal and threshold
+      // https://static.nhtsa.gov/odi/tsbs/2017/MC-10137629-9999.pdf
```
adeebshihadeh, line 96: don't duplicate this comment. you can just leave one here that says reference gm/carstate.py

sshane:
This other signal might actually be better for all cars besides the Volt. Here are some examples of the current signal we use having an offset while this new signal does not: https://imgur.com/a/Xf7As1R

PR #1094, opened 2022-10-11
Torque request bit: support multiple invalid frames

sshane:


sshane:
This is just the generic PR that adds support for multiple invalid frames btw, nothing to do with Hyundai

board/safety.h
```diff
@@ -522,36 +523,46 @@ bool steer_torque_cmd_checks(int desired_torque, int steer_req, const SteeringLi
...
       }
 
       valid_steer_req_count = 0;
       ts_steer_req_mismatch_last = ts;
+      invalid_steer_req_count = MIN(invalid_steer_req_count + 1, limits.max_invalid_request_frames);
     }
   } else {
     valid_steer_req_count = MIN(valid_steer_req_count + 1, limits.min_valid_request_frames);
+    invalid_steer_req_count = 0;
   }
```
adeebshihadeh, line 560: can you refactor this a bit to check `limits.has_steer_req_tolerance` earlier? I think it would simplify this a bit

tests/safety/common.py
```diff
@@ -219,6 +226,33 @@ def test_steer_req_bit_frames(self):
...
       self.assertTrue(self._tx(self._torque_cmd_msg(self.MAX_TORQUE, steer_req=1)))
 
+  def test_steer_req_bit_multi_invalid(self):
+    """
+      For safety modes allowing multiple consecutive invalid frames, this ensures that once a valid frame
+      is sent after an invalid frame (even without sending the max number of allowed invalid frames),
+      all counters are reset.
+    """
+    if self.MIN_VALID_STEERING_FRAMES == 0:
+      raise unittest.SkipTest("Safety mode does not implement tolerance for steer request bit safety")
```
adeebshihadeh, line 236: can we just run this test with everything set to zero? if not, why not?

board/safety.h
```diff
@@ -522,36 +523,46 @@ bool steer_torque_cmd_checks(int desired_torque, int steer_req, const SteeringLi
...
+    // no torque if request bit isn't high and safety mode doesn't allow tolerance
     if (!limits.has_steer_req_tolerance) {
       violation = true;
 
     } else {
-      // disallow torque cut if not enough recent matching steer_req messages
-      if (valid_steer_req_count < limits.min_valid_request_frames) {
-        violation = true;
-      }
+      if (invalid_steer_req_count == 0) {
```
adeebshihadeh, line 535: is this really needed? seems like all three checks can just always be run, though might have to account for that in the RT interval

sshane: We reset `valid_steer_req_count` and the rt timer every time we get an invalid frame, so we'd need to conditionally reset those once we sent `max_invalid_request_frames` frames. But then if you send less than what's allowed, it will never be reset, allowing you to oscillate back and forth.

board/safety.h
sshane: flipped around, to me it looks a bit messier, but let me know what you think

tests/safety/common.py
sshane: It would work if they were set to 0, however most safety modes don't have safety around the steer request bit, and we assume it does with these tests

adeebshihadeh: yeah, let's add a comment. I'd like to do this in the future

tests/safety/common.py
```diff
@@ -219,6 +226,33 @@ def test_steer_req_bit_frames(self):
...
+    self.safety.set_timer(self.MIN_VALID_STEERING_RT_INTERVAL)
+
+    # Allow torque cut
+    self.safety.set_controls_allowed(True)
+    self._set_prev_torque(self.MAX_TORQUE)
+    for _ in range(self.MIN_VALID_STEERING_FRAMES):
+      self.assertTrue(self._tx(self._torque_cmd_msg(self.MAX_TORQUE, steer_req=1)))
+
+    # Send at least 1 or partial amount of allowed invalid frames
+    for _ in range(max(self.MAX_INVALID_STEERING_FRAMES - 1, 1)):
```
adeebshihadeh, line 248: why can't this just be `self.MAX_INVALID_STEERING_FRAMES`?

board/safety.h
adeebshihadeh: I think it's easier to follow now

tests/safety/common.py
sshane: I wanted this test to ensure that if we don't send the full amount of invalid frames, sending a valid frame again would reset the counters, disallowing oscillating back and forth between valid and invalid, like the comment above talks about

sshane: Cleaned up to test varying numbers of frames

PR #1098, opened 2022-10-12
Hyundai CAN-FD: support 90deg angle limit bypass

sshane:


board/safety/safety_hyundai_canfd.h
```diff
@@ -9,6 +9,13 @@ const SteeringLimits HYUNDAI_CANFD_STEERING_LIMITS = {
   .driver_torque_allowance = 250,
   .driver_torque_factor = 2,
   .type = TorqueDriverLimited,
+
+  // the EPS faults when the steering angle is above a certain threshold for too long. to prevent this,
+  // we allow setting torque actuation bit to 0 while maintaining the requested torque value for two consecutive frames
+  .min_valid_request_frames = 89,
+  .max_invalid_request_frames = 2,
```
adeebshihadeh, line 16: make sure to check these numbers again on the EV6; we should be using as many minimum valid frames as we can

sshane: The EV6 actually faulted at 2 occasionally, we might need more

sshane: I don't know where I saw that. I just did a prolonged test, no faulting at 2 invalid frames (many faults with only 1). I also tested 95 valid frames with lots of faults, so I don't want to go too much higher there either

PR #1101, opened 2022-10-13
Hyundai CAN-FD: ignore LKAS steering on HDA2 long mode

adeebshihadeh:
In HDA2 long mode, we use LFA to steer instead of LKAS.


sshane:
No more steering faults!

PR #1103, opened 2022-10-14
query_fw_versions: auto-select panda with bus

sshane:
so we can query with external pandas like red panda


adeebshihadeh:
This doesn't really make sense since the panda list isn't sorted. I'd just change it to take a specific panda serial number.

sshane:
@incognitojam want to re-open your PR? It did pretty much that

PR #1104, opened 2022-10-14
GM: different brake threshold by platform

sshane:


board/safety/safety_gm.h
```diff
@@ -95,7 +95,11 @@ static int gm_rx_hook(CANPacket_t *to_push) {
     if (addr == 190) {
       // Reference for signal and threshold:
```
sshane, line 96:
```suggestion
      // Reference for signal and thresholds:
```

PR #1105, opened 2022-10-14
GM: match ECM standstill check

sshane:
also matches openpilot

- [ ] Do safety test

### DIFF
--- a/board/safety/safety_gm.h
+++ b/board/safety/safety_gm.h
@@ -16,6 +16,8 @@ const int GM_MAX_RATE_DOWN = 17;
 const int GM_DRIVER_TORQUE_ALLOWANCE = 50;
 const int GM_DRIVER_TORQUE_FACTOR = 4;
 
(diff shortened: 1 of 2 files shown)


board/safety/safety_gm.h
```diff
@@ -16,6 +16,8 @@ const int GM_MAX_RATE_DOWN = 17;
 const int GM_DRIVER_TORQUE_ALLOWANCE = 50;
 const int GM_DRIVER_TORQUE_FACTOR = 4;
 
+const int GM_STANDSTILL_THRSLD = 11;  // 0.3421kph
+
```
sshane, line 20: All wheel speeds are 0.0311 (10 in raw units) as the highest speed you can engage with your foot on the brake, however safety tests work with `<` not `<=`, so use 11

sshane: And on other cars (like Silverado, the minimum wheel speed is around 0.6 kph, so the standstill for them is not affected

PR #1114, opened 2022-10-21
Hyundai: max torque in safety param

sshane:


board/safety/safety_hyundai_common.h
```diff
@@ -19,11 +23,13 @@ enum {
 bool hyundai_ev_gas_signal = false;
 bool hyundai_hybrid_gas_signal = false;
 bool hyundai_longitudinal = false;
+int hyundai_max_steer = 0;
 uint8_t hyundai_last_button_interaction;  // button messages since the user pressed an enable button
 
 void hyundai_common_init(uint16_t param) {
   hyundai_ev_gas_signal = GET_FLAG(param, HYUNDAI_PARAM_EV_GAS);
   hyundai_hybrid_gas_signal = !hyundai_ev_gas_signal && GET_FLAG(param, HYUNDAI_PARAM_HYBRID_GAS);
+  hyundai_max_steer = param & HYUNDAI_PARAM_MAX_STEER;
```
sshane, line 32: @adeebshihadeh do you want to create the steering safety struct on init, or modify it somehow?

python/__init__.py
```diff
@@ -209,12 +209,13 @@ class Panda:
...
-  FLAG_HYUNDAI_CAMERA_SCC = 8
-  FLAG_HYUNDAI_CANFD_HDA2 = 8
-  FLAG_HYUNDAI_CANFD_ALT_BUTTONS = 16
+  # first byte is for max actuation torque
+  FLAG_HYUNDAI_EV_GAS = (1 << 8)
+  FLAG_HYUNDAI_HYBRID_GAS = (2 << 8)
+  FLAG_HYUNDAI_LONG = (4 << 8)
+  FLAG_HYUNDAI_CAMERA_SCC = (8 << 8)
+  FLAG_HYUNDAI_CANFD_HDA2 = (8 << 8)
+  FLAG_HYUNDAI_CANFD_ALT_BUTTONS = (16 << 8)
```
adeebshihadeh, line 218: why not leave these and use second byte for torque? seems cleaner

sshane: oh that's what I did for the toyota one but I remember you flipped it

board/safety/safety_hyundai_common.h
adeebshihadeh: we have to be able to modify the max torque to whatever is passed in

PR #1124, opened 2022-11-01
Nissan: check two wheel speeds for standstill

sshane:


adeebshihadeh:
Let's check all like carState.

sshane:
They're split across two messages. I was going to make openpilot check two like GM

PR #1126, opened 2022-11-02
F4: more RAM

briskspirit:
RAM in linker script was set to 128Kb (to support F205) long time ago.
Split linker scripts for F2(pedal) and F4


briskspirit:
Also F413 has minimum 1024Kb flash, fixed in linker script

PR #1133, opened 2022-11-04
Hyundai CAN-FD: more complete tests

sshane:
Tests HDA1 EV and HEV (previously only HEV)
Tests alt buttons EV and HEV (previously only HEV)

in the vein of https://github.com/commaai/panda/issues/1110


sshane:
in the vein of https://github.com/commaai/panda/issues/1110

PR #1137, opened 2022-11-07
VW MQB: Redundant brake pressed signals

jyoung8607:
The brake signal we're using today, `ESP_05.ESP_Fahrer_bremst` is from the ABS/ESP controller. Documentation reveals this signal indicates driver brake pressure over a specified threshold for that platform. It does not reflect the state of Brake Light Switch (F) or Brake Pedal/Test Switch (F63), the physically redundant mechanical switches built into the brake pedal assembly.

The pressure signal turns out to be a little lazy on MQB-A0. Very light braking would not set `ESP_05.ESP_Fahrer_bremst` but would set `Motor_14.MO_BLS` and/or `Motor_14.MO_Fahrer_bremst`. The drivetrain coordinator would fault about 400ms after onset of the Motor_14 signals if openpilot was continuing to send acceleration commands.

**My vehicle:**
`3cfdec54aa035f3f|2022-10-13--14-58-58`

Screenshot from 2022-11-07 12-26-17

**Polo with light braking:**
`0bbe367c98fa1538|2022-10-04--14-04-50`

Screenshot from 2022-11-07 12-29-10

This PR makes brake_pressed the OR of `MO_Fahrer_bremst` and `ESP_Fahrer_bremst`. This incorporates the Brake Pedal Switch (F63) into brake-pressed detection without losing anything that was detected before. Documentation implies the stock ACC radar watches both of these *and* the Brake Light Switch (F) in `MO_BLS`, but also implies `MO_BLS` may be unfiltered/noisy or even potentially not installed on some vehicles, so I'm not including that for now.

Requires a corresponding openpilot PR.

**Route:** `0bbe367c98fa1538|2022-11-08--14-22-14`


jyoung8607:
Also lengthened the safety test CI timeout due to an earlier CI failure.

Screenshot from 2022-11-07 15-52-36

PR #1142, opened 2022-11-08
Steer request bit safety: fix zero steer bug

sshane:
This bug occurs in this segment `eb5a50f9e98f50b9|2022-10-31--12-44-32--1` from https://github.com/commaai/openpilot/issues/26176.

The problem is, panda doesn't know the difference between openpilot turning off the bit to disengage lateral, or if it's performing a steer req bit cut ***if the desired torque is 0***, which often can be the case when user torque is high on driver torque limited cars.

A fix: continue to run through the allowed max invalid bit frame flow, so we don't reset `valid_steer_req_count` early

We're cutting for 3 frames here: you can see that the desired torque happens to drop to 0 from high user torque (but lat is still active), so panda resets the min valid frames instead of staying in the code that continues to allow a bit drop/invalid frames, so that when the third and final message is sent with a low bit, the torque becomes non-zero again and it blocks it.

Screenshot from 2022-11-08 15-12-17

### DIFF
--- a/board/safety.h
+++ b/board/safety.h
@@ -535,7 +535,7 @@ bool steer_torque_cmd_checks(int desired_torque, int steer_req, const SteeringLi
     if (steer_req_mismatch || (invalid_steer_req_count != -1 && invalid_steer_req_count < limits.max_invalid_request_frames)) {
       if (invalid_steer_req_count == 0) {
         // disallow torque cut if not enough recent matching steer_req messages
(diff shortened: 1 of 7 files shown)


sshane:
Before, after: https://imgur.com/a/EcnRowH

sshane:
Unclear how to handle this properly, this isn't the right way to do it. Will re-open when I have something

PR #1144, opened 2022-11-10
safety: fix regen controls mismatch

sshane:
How to reproduce:

1. Come to stop, press brake, engage.
2. With foot on brake, depress regen paddle.

To openpilot, brakePressed does not change. To panda, it independently checks the rising edges of the two signals.


sshane:
Assuming I add tests, is this good to merge @adeebshihadeh? It matches openpilot behavior

PR #1148, opened 2022-11-11
H7: enable USB LDO and fix RGB LED gpios

briskspirit:


board/boards/tres.h
```diff
@@ -2,6 +2,15 @@
 // Tres + Harness //
 // /////////////////
 
+void red_tres_init(void) {
+  // Enable USB 3.3V LDO for USB block
+  register_set_bits(&(PWR->CR3), PWR_CR3_USBREGEN); // Only for tres
+  register_set_bits(&(PWR->CR3), PWR_CR3_USB33DEN);
+  while ((PWR->CR3 & PWR_CR3_USB33RDY) == 0);
+
+  red_chiplet_init();
```
adeebshihadeh, line 11: think it makes more sense to do this first

board/boards/tres.h
```diff
@@ -2,6 +2,15 @@
 // Tres + Harness //
 // /////////////////
 
+void red_tres_init(void) {
```
adeebshihadeh, line 5:
```suggestion
void tres_init(void) {
```

board/boards/tres.h
briskspirit: I am confused, what to do first?

PR #1151, opened 2022-11-11
setup UART for SOM debugging

robbederks:
Ideally we would just enumerate as a USB CDC-ACM device to show up as a COM-port natively, but for now our uart interface will do the job


robbederks:
To use:
```
PYTHONUNBUFFERED=1 NO_COLOR=1 CLAIM=1 PORT=4 ./debug_console.py
```

PR #1160, opened 2022-11-17
F4: cleanup clock source

adeebshihadeh:


adeebshihadeh:
@robbederks Not sure if we had a use case for this before, but like this is all unused now.

robbederks (approved):

PR #1164, opened 2022-11-22
Mazda gen 4 port

MoreTore:
Tested on OP version 0.8.13. Ported to master.


MoreTore:
Related PR. https://github.com/commaai/cereal/pull/386

gregjhogan:
other brands have parameterized what looks like very similar changes (no new safety mode)
e.g. https://github.com/commaai/panda/blob/master/board/safety/safety_honda.h

MoreTore:
So I should change the code to use the safety param passed to the init function instead of using separate safetyModel

PR #1171, opened 2022-11-29
Safety: common acceleration cmd checks

sshane:


board/safety.h
```diff
@@ -486,6 +486,16 @@ float interpolate(struct lookup_t xy, float x) {
   return ret;
 }
 
+// Safety checks for longitudinal actuation
+bool long_accel_checks(int desired_accel, const LongitudinalLimits limits, const bool longitudinal_allowed) {
```
adeebshihadeh, line 490: prefer a little verbosity here for readability
```suggestion
bool longitudinal_accel_checks(int desired_accel, const LongitudinalLimits limits, const bool longitudinal_allowed) {
```

PR #1180, opened 2022-12-01
scons: put targets in global SConscript

sshane:


SConstruct
```diff
@@ -1,5 +1,15 @@
-# panda fw
-SConscript('board/SConscript')
+import os
+
+TICI = os.path.isfile('/TICI')
+
+AddOption('--test',
+          action='store_true',
+#           default=not TICI,
```
adeebshihadeh, line 7: can just default true here

SConstruct
```diff
@@ -1,5 +1,7 @@
-# panda fw
-SConscript('board/SConscript')
+AddOption('--test',
+          action='store_true',
+          default=True,
+          help='build test files')
```
sshane, line 4: Verified that openpilot's option overrides panda's here when running `scons` from the root directory, so test still won't be built on device.

PR #1182, opened 2022-12-01
safety: common angle checks

sshane:
doing what was planned 3 years ago today! https://github.com/commaai/panda/pull/244


sshane:
@robbederks let me know if you have any other clean ups in mind related to this!

adeebshihadeh (changes requested):
Can you split up the test cleanups from the actual safety code refactors?

PR #1187, opened 2022-12-04
Subaru Forester 2022 safety

martinl:
This PR adds support for Subaru Forester 2022 safety mode which is using new LKAS message (ES_LKAS_2)
Prerequisite: https://github.com/commaai/opendbc/pull/758


martinl:
I updated the safety mode to use ES_Status Cruise_Activated signal which is a better representation of cruise state than previously used ES_DashStatus signal.

jnewb1:
need a route that doesn't have the SnG stuff from martinls branch

jnewb1:
https://github.com/commaai/panda/pull/1520

board/safety/safety_subaru.h
```diff
@@ -186,9 +222,12 @@ static int subaru_fwd_hook(int bus_num, int addr) {
 
 static const addr_checks* subaru_init(uint16_t param) {
   subaru_gen2 = GET_FLAG(param, SUBARU_PARAM_GEN2);
+  subaru_gen3 = GET_FLAG(param, SUBARU_PARAM_GEN3);
 
   if (subaru_gen2) {
     subaru_rx_checks = (addr_checks){subaru_gen2_addr_checks, SUBARU_GEN2_ADDR_CHECK_LEN};
+  } else if (subaru_gen3) {
+    subaru_rx_checks = (addr_checks){subaru_gen3_addr_checks, SUBARU_GEN3_ADDR_CHECK_LEN};
```
sshane, line 230: Do we know if Subaru is calling this car having a gen 2 EyeSight system? Where did we get the current gen 2 name? Would like to make the name appropriate. Right now there's just a few differences. Perhaps these are still gen 2 cars, just with alternate LKAS and cruise activated signals, like we handle Hyundai's differences.

jnewb1: I think they are similar enough to where it doesn't make sense to make new files, but flags might make sense

SUBARU_ALT_LKAS
(Only this car has this arm, but I suspect it will also be on the other cars with the new harnesses)

and

SUBARU_USE_ES_STATUS
(Cruise_Activated was removed from CruiseControl, need to use ES_STATUS instead, unless we can find another bit from the PCM that matches our old CruiseControl bit. Other cars have this problem too, such as the hybrids.)

martinl: gen2 refers more to new canbus architecture with some critical
messages on 2 redundant buses. Eyesight can messages have not changed much in the process and can use messsage/feature flags

martinl: https://github.com/martinl/openpilot/wiki/CAN-Network-diagrams

jnewb1:
The forester seems more like gen1 than gen2, the messages we care about are on bus0/2. I think flags makes more sense, especially since we want to support the hybrid cars which will also have small differences like this, but without a new "generation"

sshane:
Have we actually controlled a Forester yet? This LKAS_Output in ES_LKAS_ALT is actually a steering angle, not torque.

image

jnewb1:
> Have we actually controlled a Forester yet? This LKAS_Output in ES_LKAS_ALT is actually a steering angle, not torque... That's why it's non-zero and follows the angle while it's not steering, angle commands usually do (LTA, Tesla, etc)
> 
> image

martinl has the 22 forester on subaru-community and at least 2 people have been running with it. I suppose angle and torque might seem similar on the highway, but I will look into this further

sshane:
Yes, that's what Toyota did when they transitioned to LTA. You could still send the old torque message for multiple model years before they stopped responding to it (2023 RAV4).

sshane:
Also, ES_Status doesn't seem to capture everything. Here is the stock system requesting gas, but that bit is off. I'd like to understand why and figure out which bit is actually "cruise activated" before we move forward. From the stock route 7fd1e4f3a33c1673|2022-12-07--20-16-23

image

sshane:
btw there's another ES_Distance message, seems to be near duplicate

image

martinl:
> Also, ES_Status doesn't seem to capture everything. Here is the stock system requesting gas, but that bit is off. I'd like to understand why and figure out which bit is actually "cruise activated" before we move forward. From the stock route 7fd1e4f3a33c1673|2022-12-07--20-16-23
> 
> image
> 

Eyesight generates the gas request and rpm signals in the background even when it’s not engaged

> Also, ES_Status doesn't seem to capture everything. Here is the stock system requesting gas, but that bit is off. I'd like to understand why and figure out which bit is actually "cruise activated" before we move forward. From the stock route 7fd1e4f3a33c1673|2022-12-07--20-16-23
> 
> image
> 

Eyesight generates cruise throttle and rpm signals also in the background when it is not engaged. You can check that other models that have CruiseControl message behave the same

jnewb1:
superseded by https://github.com/commaai/panda/pull/1533/files

PR #1191, opened 2022-12-12
No more magic for the can chunks

robbederks:


board/can_comms.h
```diff
@@ -1,15 +1,15 @@
...
   into multiple transfers or chunks.
 
   * comms_can_read outputs this buffer in chunks of a specified length.
     chunks are always the given length, except the last one.
-  * comms_can_write reads in this buffer in chunks. start of the transaction
-    is denoted by the CAN_TRANSACTION_MAGIC value.
+  * comms_can_write reads in this buffer in chunks.
   * both functions maintain an overflow buffer for a partial CANPacket_t that
     spans multiple transfers/chunks.
+  * the overflow buffers are reset by a dedicated control transfer (0xc0),
```
adeebshihadeh, line 11: best to avoid specific values in comments since it's not obvious when they break
```suggestion
  * the overflow buffers are reset by a dedicated control transfer endpoint,
```

tests/usbprotocol/test_comms.py
```diff
@@ -28,6 +28,9 @@ def random_can_messages(n, bus=None):
 
 
 class TestPandaComms(unittest.TestCase):
+  def setUp(self):
+    lpp.comms_can_reset()
```
adeebshihadeh, line 32: we should have a test case for this

python/__init__.py
```diff
@@ -655,6 +649,10 @@ def set_uart_callback(self, uart, install):
   # Timeout is in ms. If set to 0, the timeout is infinite.
   CAN_SEND_TIMEOUT_MS = 10
 
+  @ensure_can_packet_version
```
adeebshihadeh, line 652: if we call reset on init, this won't allow old FW to connect

robbederks: not sure what you are suggesting?

PR #1204, opened 2023-01-09
fix debug uart buffer size

pd0wm:
Buffer size was changed in https://github.com/commaai/panda/pull/1151. Since `rx_dma` is false I assume the buffer size should also be `FIFO_SIZE_INT`.

For some reason using a larger buffer causes the firmware to stop working on f4 devices when compiling on MacOS (`arm-none-eabi-gcc (Arm GNU Toolchain 12.2.Rel1 (Build arm-12.24)) 12.2.1 20221205`) or with the latest compiler from the arch repos ( `arm-none-eabi-gcc (Arch Repository) 12.2.0`). Since this is unexpected I assume you'd like to figure out why. 

`.bss` size is the same between both compilers, so it's not suddenly creating larger global variables.

### DIFF
--- a/board/drivers/uart.h
+++ b/board/drivers/uart.h
@@ -51,7 +51,7 @@ UART_BUFFER(lin1, FIFO_SIZE_INT, FIFO_SIZE_INT, UART5, NULL, false)
 UART_BUFFER(lin2, FIFO_SIZE_INT, FIFO_SIZE_INT, USART3, NULL, false)
 
 // debug = USART2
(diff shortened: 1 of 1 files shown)


robbederks:
Interesting. Already kinda ran into this issue while making that PR (which is why I did https://github.com/commaai/panda/pull/1151/commits/8f99a153ea68bd3117f38294d02454edefc72d8c).

Can't remember why I increased the debug buffer size, but should be fine to revert like you did.

At some point we should investigate why this breaks the code, we must just have something configured wrong somewhere (overflowing into a bank we haven't enabled the power to or something like that?).

PR #1209, opened 2023-01-13
Panda: context manager support

sshane:
- Now also closes the panda when an exception occurs inside the context manager


sshane:
@briskspirit do you have any desire for this type of PR?

PR #1218, opened 2023-01-19
Compile panda on apple silicon

tomtom-95:
On my Mac one of the first problem I get when I try to build openpilot is fatal error: initializer element is not a compile-time constant. This fix the error


tomtom-95:
Thanks for the answer @adeebshihadeh but can you be a bit more specific on why it is not a good fix (or point me on some material to understand better the issue)?

PR #1227, opened 2023-01-27
SPI support in bootstub

adeebshihadeh:


board/stm32fx/peripherals.h
```diff
@@ -5,6 +5,15 @@ void gpio_usb_init(void) {
   GPIOA->OSPEEDR = GPIO_OSPEEDER_OSPEEDR11 | GPIO_OSPEEDER_OSPEEDR12;
 }
 
+void gpio_spi_init(void) {
```
robbederks, line 8: This might be more board-specific, why not just have this in `dos` and `red_chiplet`?

robbederks (approved):

board/stm32fx/peripherals.h
adeebshihadeh: because we don’t do the whole board init in the flasher, but perhaps that’s a legacy thing?

robbederks: Ah, I see.
Should be fine to keep it as-is then

python/__init__.py
```diff
@@ -440,7 +449,7 @@ def flash_static(handle, code, mcu_type):
     # erase sectors
     logging.warning(f"flash: erasing sectors 1 - {last_sector}")
     for i in range(1, last_sector + 1):
-      handle.controlWrite(Panda.REQUEST_IN, 0xb2, i, 0, b'')
+      handle.controlWrite(Panda.REQUEST_IN, 0xb2, i, 0, b'', timeout=int(15*1e4))
```
robbederks, line 452: why 1e4? libusb's timeout is in ms, which would need 1e3

python/spi.py
```diff
@@ -96,9 +96,10 @@ def _calc_checksum(self, data: List[int]) -> int:
       cksum ^= b
     return cksum
 
-  def _wait_for_ack(self, spi, ack_val: int) -> None:
+  def _wait_for_ack(self, spi, timeout: int, ack_val: int) -> None:
     start = time.monotonic()
-    while (time.monotonic() - start) < ACK_TIMEOUT_SECONDS:
+    timeout_seconds = max(timeout * 1e-4, MIN_ACK_TIMEOUT_SECONDS)
```
robbederks, line 101: same here, 1e3

PR #1123, opened 2022-10-28
safety: add longitudinal pre-enable state for brake at a standstill

sshane:
Adds new pre-enable state when enabling at a standstill with brake depressed. Allows brake and negative accel to 0 accel, blocks positive accel, gas, interceptor.

Pre-reqs (so this change isn't unmaintainable):
- [x] https://github.com/commaai/panda/pull/1171
- [x] https://github.com/commaai/panda/pull/1168
- [x] https://github.com/commaai/panda/pull/1178
- [x] https://github.com/commaai/panda/pull/1179
- [x] https://github.com/commaai/panda/pull/1176
- [x] Common angle-based steer checks

PR #1230, opened 2023-01-28
Disallow pedal interceptor while braking

nworb-cire:
Certain edge cases (primarily enabling at a stop) can lead to panda allowing gas interceptor commands while the brake pedal is pressed. This behavior causes the throttle to fight the brakes and is in violation of the safety policy.


sshane:
We should first fix this in openpilot, but I agree this isn't great. Do you have a segment where openpilot is commanding pedal actuation with the brakes pressed?

nworb-cire:
GM doesn't yet have pedal interceptor support in stock openpilot, so if you're asking about stock, we'd have to check Honda or Toyota. Looking through the code there, I think it may still be possible for this behavior to occur. 

If you're able to access an unpublished route, here's a dirty one of me using joystick mode: `806907cd33ef2647|2023-01-28--15-32-39--0`. I can get a clean, published route for this if necessary.

sshane:
Can you also do this for the gas (non interceptor) check and positive accel check? You might want to make a helper

nworb-cire:
@sshane any other changes you'd like to see on this?

board/safety.h
```diff
@@ -500,7 +504,7 @@ int ROUND(float val) {
 
 // Safety checks for longitudinal actuation
 bool longitudinal_accel_checks(int desired_accel, const LongitudinalLimits limits) {
-  bool accel_valid = get_longitudinal_allowed() && !max_limit_check(desired_accel, limits.max_accel, limits.min_accel);
+  bool accel_valid = get_accel_allowed() && !max_limit_check(desired_accel, limits.max_accel, limits.min_accel);
```
sshane, line 507: If you enable at a stop with foot on brake, we treat this state as pre-enabled in openpilot. We want to block just gas in this state, since requesting brake to stay stopped is fine (and needed for some brake/non-accel cars like GM and Honda), and we switched to allowing steering in pre-enable as well.

Currently, this should block those messages. Can we change this to allow only brake?

board/safety.h
```diff
@@ -81,6 +81,10 @@ int safety_fwd_hook(int bus_num, int addr) {
   return (relay_malfunction ? -1 : current_hooks->fwd(bus_num, addr));
 }
 
+bool get_accel_allowed(void) {
+  return controls_allowed && !gas_pressed_prev && !brake_pressed_prev;
+}
+
 bool get_longitudinal_allowed(void) {
   return controls_allowed && !gas_pressed_prev;
 }
```
sshane, line 90: if this is a superset, maybe use it in get_accel_allowed?

sshane (changes requested):

sshane:
Thanks but found we already have a PR in progress to do this, but adds the pre-enable state for all longitudinal control modes (gas/brake, accel, interceptor): https://github.com/commaai/panda/pull/1123

PR #1231, opened 2023-01-28
python: common base handle for SPI + USB

adeebshihadeh:


sshane:
Fails CI bumped on openpilot:

```
panda/python/usb.py:19: error: Returning Any from function declared to return "int"
panda/python/usb.py:22: error: Returning Any from function declared to return "bytes"
Found 2 errors in 1 file (checked 134 source files)
panda/python/usb.py:19: error: Returning Any from function declared to return "int"
panda/python/usb.py:22: error: Returning Any from function declared to return "bytes"
Found 2 errors in 1 file (checked 134 source files)
```

Trying to see why mypy doesn't catch it in panda first

PR #1232, opened 2023-01-29
GM: lower steering RT buffer to 20%

sshane:
Matches other makes like Toyota


sshane:
nvm, seems like not all cars use 20%. no real need to alter just one make right now

PR #1233, opened 2023-02-08
Same mypy checks as openpilot

sshane:


adeebshihadeh:
Nice catch. We may want to set up a standalone repo with common linter/pre-commit/GH action configs to use in all our repos.

PR #1236, opened 2023-02-09
Microsecond timer fix

robbederks:
@sshane noticed the microsecond timer is actually running as a half-microsecond timer

### DIFF
--- a/board/drivers/clock_source.h
+++ b/board/drivers/clock_source.h
@@ -3,7 +3,7 @@
 
 void clock_source_init(void) {
   // Setup timer
-  register_set(&(TIM1->PSC), ((APB2_FREQ*100U)-1U), 0xFFFFU);                 // Tick on 0.1 ms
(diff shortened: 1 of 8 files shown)


robbederks:
Seems to have been broken by changing the prescaler in https://github.com/commaai/panda/pull/1061

@briskspirit can you double check that you didn't break anything else with that change?

briskspirit:
> Seems to have been broken by changing the prescaler in #1061
> 
> @briskspirit can you double check that you didn't break anything else with that change?

Those values were used for 3 timers only, see no anything else broken.

briskspirit (approved):

robbederks:
> > Seems to have been broken by changing the prescaler in #1061
> > 
> > @briskspirit can you double check that you didn't break anything else with that change?
> 
> Those values were used for 3 timers only, see no anything else broken.

It's not only the values, in the PR that broke this you changed the APB1 frequency. Are there any other peripherals that could quietly be affected by that change, apart from the microsecond timer?

briskspirit:
By clock configuration map for both F4 and H7 everything else looks good.

PR #1247, opened 2023-02-21
VW: Allow inactive accel values at all times

jyoung8607:
Fix a bug with VW when using openpilot longitudinal: upon hitting Set or Resume (transitioning Panda to controls allowed) current safety code incorrectly starts dropping ACC control frames with the inactive accel value.

The current code only checks for and allows the inactive accel value when `get_longitudinal_allowed()` is False. The VW inactive accel value falls outside the normally valid accel range, which appears to be unique to VW. 

This left open several scenarios wherein openpilot would fault the VW drivetrain coordinator after a rejected engagement. The most common case was Resume without a prior Set, but it would also happen for engaging without a seatbelt on, or any other noEntry case that still passes Panda safety checks.

- [x] Augment the VW MQB and VW PQ longitudinal safety tests to probe this specific case
- [x] Refactor `longitudinal_accel_checks()` to always allow the car port's inactive accel value

**Route with bug:** `d5eb604aa6ae46b7|2023-02-20--19-52-14`
**Route with fix:** `d5eb604aa6ae46b7|2023-02-20--23-22-21`


board/safety.h
```diff
@@ -485,13 +485,9 @@ float interpolate(struct lookup_t xy, float x) {
 
 // Safety checks for longitudinal actuation
 bool longitudinal_accel_checks(int desired_accel, const LongitudinalLimits limits) {
-  bool violation = false;
```
adeebshihadeh, line 488: @sshane can you merge this and make the rest of the tests fail?

sshane: would definitely fail with old safety and updated tests now

```python
FAIL: test_accel_safety_check (__main__.TestVolkswagenMqbLongSafety)
----------------------------------------------------------------------
  File "./test_volkswagen_mqb.py", line 218, in test_accel_safety_check
    self.assertEqual(send, self._tx(self._acc_06_msg(accel)), (controls_allowed, accel))
AssertionError: True != 0 : (True, 3.01)
```

```
FAIL: test_accel_actuation_limits (__main__.TestVolkswagenPqLongSafety)
----------------------------------------------------------------------
  File "/home/batman/openpilot/panda/tests/safety/common.py", line 169, in test_accel_actuation_limits
    self.assertEqual(should_tx, self._tx(self._accel_msg(accel)))
AssertionError: True != 0
```

sshane:
Longitudinal safety now
- allows acceleration message if controls are allowed and it is within limits
- OR if inactive accel in any state

PR #1250, opened 2023-02-22
safety: reset torque_meas on init properly

sshane:
fixes 3 year old bug :slightly_smiling_face: https://github.com/commaai/panda/pull/542


adeebshihadeh:
Nice fix, let's add a failing test case too.

sshane:
The tests for the sample_t behavior is all over the place. There's only one test in the EPS torque safety class that asserts specific behavior of the EPS safety + behavior of updating sample_t (min/max), when that could be common and each class just asserts that the meas values are being parsed correctly into the class

PR #1252, opened 2023-02-22
RTC: increase LSE drive strength on H7

robbederks:


board/stm32fx/llrtc.h
```diff
@@ -1,6 +1,8 @@
 #define RCC_BDCR_MASK_LSE (RCC_BDCR_RTCEN | RCC_BDCR_RTCSEL | RCC_BDCR_LSEMOD | RCC_BDCR_LSEBYP | RCC_BDCR_LSEON)
 #define RCC_BDCR_MASK_LSI (RCC_BDCR_RTCEN | RCC_BDCR_RTCSEL)
 
+#define RCC_BDCR_LSEDRV_1 0U
```
adeebshihadeh, line 4: this is pretty hacky, we should just do a better job of splitting up the H7 and F4 drivers

robbederks: yeah that's the reason it's still in draft, needs some cleanup

PR #1253, opened 2023-02-23
Angle control: fix limiting bug

sshane:
For https://github.com/commaai/openpilot/pull/27428

Bug prevents tests in https://github.com/commaai/panda/pull/1251 from working properly


sshane:
Turns out using the down limits (which are usually higher) near zero fixes a bug where the angle is too small to be represented in the CAN signal, so it gets converted to 0 for the panda.

An example for Nissan:

```
# Previous real commanded angle is 0.002, in safety it gets int rounded to 0, now we're trying to send -0.398 (utilizing down limits, 0.4)
set angle: 0.002, in safety: 0, sending angle: -0.398
```

However, since panda thinks the previous angle is 0, it switches to using the up limits (.15 instead of 0.4), and then blocks our message. We can instead fix openpilot to use the up limits (https://github.com/commaai/openpilot/pull/27428), but keep panda the same (using down limits).

I will add a test to assert that the down limits are always above or equal to up and add a comment describing this nuance.

sshane:
The test in https://github.com/commaai/panda/pull/1257/files using the same openpilot angle limiting function shows that without this change, it's possible for panda to block messages if the angle is close enough to zero.

PR #1254, opened 2023-02-23
Angle control: match openpilot limits

sshane:
Instead of adding an implicit fudge to the raw speed bp lists, add an explicit comment and offset in the angle limiting function. Will make it easy to keep the limits matching OP (some items in the v lists also weren't matching, which is unnecessary with a speed offset. fixed)

We can probably reduce the speed fudging even more, but I left a TODO comment to look at data to determine what a reasonable factor would be for now.

TLDR, what does this PR do?
- Change speed fudging from 2 m/s to 1 m/s (reduce tolerance)
- Fixes one point of the angle limit lookup tables not matching OP
- Makes the lookup tables match the tests' and OP's visually and by value


sshane:
Screenshot from 2023-02-22 20-01-47

PR #1255, opened 2023-02-23
Nissan: use averaged left and right wheel speeds

sshane:
The error from the speed openpilot uses can be up to at least 0.6 m/s away using just one side of the car. Allows us to reduce the speed fudge for the angle rate limits: https://github.com/commaai/panda/pull/1254

Screenshot from 2023-02-22 19-30-04


sshane:
@robbederks do you want to update Tesla to use the same speed message openpilot uses? Looks like panda uses `DI_torque2->DI_vehicleSpeed`, and openpilot uses `ESP_B->ESP_vehicleSpeed`. Any reason?

robbederks:
> @robbederks do you want to update Tesla to use the same speed message openpilot uses? Looks like panda uses `DI_torque2->DI_vehicleSpeed`, and openpilot uses `ESP_B->ESP_vehicleSpeed`. Then we can ensure that standstill matches exactly as well

Looks like it used to match until I implemented longitudinal safety (second panda). I doubt the ESP message is on the bus my second panda uses, but it might be possible to switch over openpilot to use the DI message instead. Will make an issue

PR #1249, opened 2023-02-22
Ford: driver torque limit safety

sshane:

PR #1264, opened 2023-02-25
Ford: check quality flag

sshane:
Generically checks the quality flag signal for the yaw rate and speed signal to be used in the driver torque limiting PR: https://github.com/commaai/panda/pull/1249/files

### DIFF
--- a/board/safety.h
+++ b/board/safety.h
@@ -239,8 +239,9 @@ bool addr_safety_check(CANPacket_t *to_push,
       rx_checks->check[index].wrong_counters = 0U;
     }
 
-    // quality flag check
-    if ((get_quality_flag_valid != NULL) && rx_checks->check[index].msg[rx_checks->check[index].index].check_quality_flag) {
+    // quality flag check. warning: this is not yet checked in openpilot's CAN parser,
+    // this may be the cause of blocked messages with some safety modes
+    if ((get_quality_flag_valid != NULL) && rx_checks->check[index].msg[rx_checks->check[index].index].quality_flag) {
       rx_checks->check[index].valid_quality_flag = get_quality_flag_valid(to_push);
     } else {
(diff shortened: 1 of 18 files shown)


sshane:
@adeebshihadeh do we want some sort of test for this? I don't recall an explicit counter or checksum test.

board/safety.h
```diff
@@ -237,6 +238,14 @@ bool addr_safety_check(CANPacket_t *to_push,
     } else {
       rx_checks->check[index].wrong_counters = 0U;
     }
+
+    // quality flag check. warning: this is not yet checked in openpilot's CAN parser,
+    // this may be the cause of blocked messages with some safety modes
```
adeebshihadeh, line 243: this is a per car port thing. should go in the ford safety mode

board/safety_declarations.h
```diff
@@ -97,6 +98,7 @@ typedef struct {
   int index;                         // if multiple messages are allowed to be checked, this stores the index of the first one seen. only msg[msg_index] will be used
   bool valid_checksum;               // true if and only if checksum check is passed
   int wrong_counters;                // counter of wrong counters, saturated between 0 and MAX_WRONG_COUNTERS
+  bool valid_quality_flag;           // true if the message's quality flag signifies signals are valid
```
adeebshihadeh, line 101:
```suggestion
  bool valid_quality_flag;           // true if the message's quality/health/status signals are valid
```

adeebshihadeh (approved):
> @adeebshihadeh do we want some sort of test for this?

Yes, we definitely want tests for this!

> I don't recall an explicit counter or checksum test.

Some safety modes have checksum/counter tests (e.g.  [toyota](https://github.com/commaai/panda/blob/master/tests/safety/test_toyota.py#L146)), but they're not in the common tests yet.

sshane:
Ah I see, will update!

PR #1270, opened 2023-03-03
PandaDFU: SPI support

adeebshihadeh:


python/spi.py
```diff
@@ -172,3 +174,144 @@ def bulkRead(self, endpoint: int, length: int, timeout: int = 0) -> bytes:
...
+
+  def _erase_sector(self, sector: int):
+    p = struct.pack('>H', 0)  # number of sectors to erase
+    d = struct.pack('>H', sector)
+    self._cmd(0x44, data=[d, ], predata=p)
+
+  # *** PandaDFU API ***
+
+  def erase_app(self):
+    self._erase_sector(1)
```
adeebshihadeh, line 287: @robbederks noticed the USB DFU doesn't erase all the necessary sectors like we do in the normal library. fine to just erase sector 1 here?

robbederks: iirc on the H7 the sectors are large enough such that the whole binary fits into one sector, but that's not the case on F4.

The normal flashing routine calculates the amount of sectors you need to erase, maybe we should use that here as well?

adeebshihadeh: we don't necessarily know that though, since we're not flashing the app section. I guess we could just erase all the relevant sectors instead, but not sure that's preferable?

PR #1284, opened 2023-03-09
Raise max steer limits for Ram DT

vincentw56:
We have tested for over a year that the Ram DT can handle 350 on the max steer. Any higher and the EPS will error out. Raising this here and then a PR can be made for the SelfDrive files.


adeebshihadeh:
Thanks for the PR. We've verified this with both our Ram and the data from the Ram community fork.

PR #1287, opened 2023-03-14
Tres: fan cooldown +  no stall recovery

robbederks:
Wait a few seconds before turning the fan power off. If we turn it off with the rotor still spinning, the back EMF creates enough power to slightly power the fan controller and generate spurious tach edges.

### DIFF
--- a/board/boards/tres.h
+++ b/board/boards/tres.h
@@ -4,14 +4,8 @@
 
 bool tres_ir_enabled;
 bool tres_fan_enabled;
-uint8_t tres_fan_cooldown = 0U;
 void tres_update_fan_ir_power(void) {
-  if (tres_fan_enabled) {
-    // keep power on a few seconds after disabling to allow the fan to spin down
-    tres_fan_cooldown = 3U;
-  }
-
-  red_chiplet_set_fan_or_usb_load_switch(tres_ir_enabled || tres_fan_enabled || (tres_fan_cooldown > 0U));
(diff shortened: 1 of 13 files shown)


board/boards/tres.h
```diff
@@ -4,8 +4,14 @@
 
 bool tres_ir_enabled;
 bool tres_fan_enabled;
+uint8_t tres_fan_cooldown = 0U;
 void tres_update_fan_ir_power(void) {
-  red_chiplet_set_fan_or_usb_load_switch(tres_ir_enabled || tres_fan_enabled);
+  if (tres_fan_enabled) {
```
adeebshihadeh, line 9: seems a bit messy to have fan logic in so many places. if we start seeing stalls on these fans, will we have to account for this cooldown in the fan driver now?

briskspirit:
This is not gonna work if panda will be reset.
turn panda on, spin fan, reset panda - will get same interrupt errors.

robbederks:
> This is not gonna work if panda will be reset.
> turn panda on, spin fan, reset panda - will get same interrupt errors.

Good point, didn't think of that. Let's just init the cooldown at non-zero then

board/boards/tres.h
robbederks: Agree that it feels a bit messy now, will see if there's a good way to clean it up.
Don't expect to see any issues with these fans, so don't think we need any of the stall logic for the tres

robbederks: Moved the logic inside the fan driver, and also disabled stall recovery for all boards except the dos.

PR #1288, opened 2023-03-14
Fix different ADC scales

briskspirit:


adeebshihadeh:
Let's add a HITL test for this too.

briskspirit:
> Let's add a HITL test for this too.

Seems like we always had it

robbederks:
> Seems like we always had it

How was this not caught then? The scale change seems too big to still have it in a 10-14V range no?

PR #1293, opened 2023-03-17
Make Toyota safety tests more modular

sshane:
Toyota will have a base class that's not tested, but provides a bunch of helper rx message functions, and some base tests for all safety modes/params. This is similar to other brands

Think this will be separate from the extra LTA safety in this PR, WIP

Safety tests:

- Torque + Interceptor
- Torque + Interceptor + Alt brake
- Torque + Interceptor + Stock long
- Angle + Interceptor
- Angle + Interceptor + Stock long


sshane:
Superseded by https://github.com/commaai/panda/pull/1295

PR #1294, opened 2023-03-17
Make Toyota safety tests more modular

sshane:
Toyota will have a base class that's not tested, but provides a bunch of helper rx message functions, and some base tests for all safety modes/params. This is similar to other brands

Safety tests before and after this PR (121 tests):

- Torque + Interceptor
- Torque + Interceptor + Alt brake
- Torque + Interceptor + Stock long


sshane:
Basically:
- Moved torque params into `TestToyotaSafetyTorque`, rest is now in untested `TestToyotaSafetyBase` which every other class stems from.

PR #1298, opened 2023-03-21
safety tests: add common longitudinal accel class

sshane:


sshane:
@adeebshihadeh do you want to test that `RAISE_LONGITUDINAL_LIMITS_TO_ISO_MAX` actually does something, or that it doesn't? It asserts that nothing changes right now

PR #1309, opened 2023-03-30
Honda Bosch Radarless: never set AEB bits

royjr:
Never set the AEB bits in the ACC_CONTROL message


tests/safety/test_honda.py
```diff
@@ -510,6 +512,9 @@ def test_brake_safety_check(self):
         send = self.MIN_ACCEL <= accel <= self.MAX_ACCEL if controls_allowed else accel == 0
         self.assertEqual(send, self._tx(self._send_gas_brake_msg(self.NO_GAS, accel)), (controls_allowed, accel))
 
+  def test_no_aeb(self):
+    self.assertFalse(self._tx(self._send_gas_brake_msg(self.NO_GAS, 0, aeb_prepare=1)))
+    self.assertFalse(self._tx(self._send_gas_brake_msg(self.NO_GAS, 0, aeb_braking=1)))
```
royjr, line 517: Maybe the test should follow the assertEqual in test_brake_safety_check using something similar to this
`self.assertEqual(send, self._tx(self._send_gas_brake_msg(self.NO_GAS, 0, aeb_prepare=1))), (controls_allowed, accel))`

sshane: yes, you probably want to check that you can send a NO_GAS msg previous to this

royjr:
Sorry, don't have the car anymore. We can reopen if someone else wants to work on it.

PR #1312, opened 2023-03-31
IsoTpMessage: check CAN frame length

sshane:
https://en.wikipedia.org/wiki/ISO_15765-2

> ISO-TP prepends one or more metadata bytes to the payload data in the eight byte CAN frame, reducing the payload to seven or fewer bytes per frame. The metadata is called the Protocol Control Information, or PCI. The PCI is one, two or three bytes. The initial field is four bits indicating the frame type, and implicitly describing the PCI length.

### DIFF
--- a/python/uds.py
+++ b/python/uds.py
@@ -463,7 +463,6 @@ class IsoTpMessage():
     # single rx_frame
     if rx_data[0] >> 4 == 0x0:
       self.rx_len = rx_data[0] & 0xFF
-      assert self.rx_len <= 0x7, f"isotp - rx: invalid single frame length: {self.rx_len}"
       self.rx_dat = rx_data[1:1 + self.rx_len]
       self.rx_idx = 0
       self.rx_done = True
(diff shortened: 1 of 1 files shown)


sshane:
It actually looks like the abs ECU on Chrysler does not pad to 8 bytes. More data here: https://github.com/commaai/openpilot/pull/27760

```python
3053760197290 rxaddr=1863, bus=0, 33070.12 ms, 0x0319020d00000000, b'\x03\x19\x02\r\x00\x00\x00\x00', len(can.dat)=8
3053760197290 rxaddr=1863, bus=130, 33070.12 ms, 0x0319020d00000000, b'\x03\x19\x02\r\x00\x00\x00\x00', len(can.dat)=8
3053790202586 rxaddr=1223, bus=0, 33100.13 ms, 0x037f1978, b'\x03\x7f\x19x', len(can.dat)=4
3053790202586 rxaddr=1223, bus=130, 33100.13 ms, 0x037f1978, b'\x03\x7f\x19x', len(can.dat)=4
3053879979103 rxaddr=1223, bus=0, 33189.9 ms, 0x035902cf, b'\x03Y\x02\xcf', len(can.dat)=4
3053879979103 rxaddr=1223, bus=130, 33189.9 ms, 0x035902cf, b'\x03Y\x02\xcf', len(can.dat)=4
```

sshane:
Out of the 2093 dongles checked in the last 90 days, these non-8-byte messages were only found on these two Chrysler platforms and with only this one ECU. The occurences ranged from tester present responses, to responses to other brands' queries. So far, it always responds with 8 byte messages for the Chrysler queries.

Since these platforms aren't using FPv2 and the messages with the violations are from different unused queries, this is safe to merge.

PR #1314, opened 2023-03-31
IsoTpMessage: return if consecutive frame last received

sshane:


python/uds.py
```diff
@@ -438,25 +438,25 @@ def recv(self, timeout=None) -> Tuple[Optional[bytes], bool]:
...
-    updated = False
+    rx_in_progress = False
     try:
       while True:
         for msg in self._can_client.recv():
-          self._isotp_rx_next(msg)
+          frame_type = self._isotp_rx_next(msg)
           start_time = time.monotonic()
-          updated = True
+          rx_in_progress = frame_type == "consecutive"
```
adeebshihadeh, line 447: use an enum for this

PR #1316, opened 2023-03-31
uds: tests

sshane:
Here's an example of the test simulating a tester present query and response:

```python
ISO-TP: REQUEST - 0x750 0x3e
ISO-TP: TX - single frame - 0x750
CAN-TX: 0x750 - 0x0f013e0000000000
CAN-RX: 0x750 - 0x0f013e0000000000
ISO-TP: RX - single frame - 0x750 idx=0 done=True
ISO-TP: RESPONSE - 0x750 0x3e
ISO-TP: REQUEST - 0x758 0x7e
ISO-TP: TX - single frame - 0x758
CAN-TX: 0x758 - 0x0f017e0000000000
CAN-RX: 0x758 - 0x0f017e0000000000
ISO-TP: RX - single frame - 0x758 idx=0 done=True
ISO-TP: RESPONSE - 0x758 0x7e
```

And here's something with multiple frames!

```python
ISO-TP: REQUEST - 0x750 0x000102030405060708
ISO-TP: TX - first frame - 0x750
CAN-TX: 0x750 - 0x0f10090001020304
CAN-RX: 0x750 - 0x0f10090001020304
ISO-TP: RX - first frame - 0x750 idx=0 done=False
ISO-TP: TX - flow control continue - 0x758
CAN-TX: 0x758 - 0x0f30010000000000
ISO-TP: RESPONSE - 0x750 0x0001020304
CAN-RX: 0x758 - 0x0f30010000000000
ISO-TP: RX - flow control continue - 0x750
CAN-TX: 0x750 - 0x0f21050607080000
ISO-TP: TX - consecutive frame - 0x750 idx=1 done=True
CAN-RX: 0x750 - 0x0f21050607080000
ISO-TP: RX - consecutive frame - 0x750 idx=1 done=True
ISO-TP: RESPONSE - 0x750 0x000102030405060708
ISO-TP: REQUEST - 0x758 0x000102030405060708
ISO-TP: TX - first frame - 0x758
CAN-TX: 0x758 - 0x0f10090001020304
CAN-RX: 0x758 - 0x0f10090001020304
ISO-TP: RX - first frame - 0x758 idx=0 done=False
ISO-TP: TX - flow control continue - 0x750
CAN-TX: 0x750 - 0x0f30010000000000
ISO-TP: RESPONSE - 0x758 0x0001020304
CAN-RX: 0x750 - 0x0f30010000000000
ISO-TP: RX - flow control continue - 0x758
CAN-TX: 0x758 - 0x0f21050607080000
[...]

PR #1318, opened 2023-04-01
IsoTpMessage: handle sub addresses when sending consecutive frames

sshane:
Found a bug in the flow control section of IsoTpMessage while writing tests in https://github.com/commaai/panda/pull/1316

If we're sending to a subaddress, then the single frame only contains 5 bytes of data, and we'd start the first consecutive frame at the 7th byte (index 6), skipping the 6th byte (index 5)


sshane:
Already merged in https://github.com/commaai/panda/pull/1320, and now caught by https://github.com/commaai/panda/pull/1329

PR #1329, opened 2023-04-05
UdsServer thread

sshane:
Might want to subclass UdsClient and add the `_uds_response` function? Split out the response behavior into another method? Function? Working, pending cleanup!

### DIFF
--- a/python/uds.py
+++ b/python/uds.py
@@ -141,6 +141,12 @@ class DYNAMIC_DEFINITION_TYPE(IntEnum):
   DEFINE_BY_MEMORY_ADDRESS = 2
   CLEAR_DYNAMICALLY_DEFINED_DATA_IDENTIFIER = 3
 
+class ISOTP_FRAME_TYPE(IntEnum):
+  SINGLE = 0
+  FIRST = 1
+  CONSECUTIVE = 2
+  FLOW = 3
+
 class DynamicSourceDefinition(NamedTuple):
   data_identifier: int
   position: int
@@ -323,8 +329,11 @@ class CanClient():
     return bus == self.bus and addr == self.rx_addr
 
   def _recv_buffer(self, drain: bool = False) -> None:
+    print('top of recv buffer')
     while True:
+      time.sleep(0.1)
       msgs = self.rx()
+      print('msgs', msgs)
       if drain:
         if self.debug:
           print("CAN-RX: drain - {}".format(len(msgs)))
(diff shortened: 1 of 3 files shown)


python/uds.py
```diff
@@ -587,12 +596,13 @@ def _uds_request(self, service_type: SERVICE_TYPE, subfunction: int = None, data
 
     # send request, wait for response
     max_len = 8 if self.sub_addr is None else 7
-    isotp_msg = IsoTpMessage(self._can_client, timeout=self.timeout, debug=self.debug, max_len=max_len)
+    isotp_msg = IsoTpMessage(self._can_client, timeout=self.timeout, debug=self.debug, max_len=max_len, single_frame_mode=True)
```
sshane, line 599: we drain from the can client, so not sure if we can easily test without single frame mode

sshane:
@gregjhogan have any ideas to make the class any simpler or is this good?

tests/python/test_uds.py
```diff
@@ -0,0 +1,119 @@
...
+import unittest
+import struct
+import threading
+
+from panda.python.uds import SERVICE_TYPE, DATA_IDENTIFIER_TYPE, IsoTpMessage, UdsClient, get_rx_addr_for_tx_addr
+
+DEFAULT_VIN = '1H3110W0RLD5'
+
+
+class UdsServer(UdsClient):
```
gregjhogan, line 12: I would probably separate this class from the unit tests, it is useful for other things like building a fake car for playing around with the dealer tools.

tests/python/test_uds.py
```diff
@@ -0,0 +1,119 @@
...
+        resp_data = struct.unpack('!H', resp[1:])[0]
+        print(resp_data)
+        if resp_data == DATA_IDENTIFIER_TYPE.VIN:
+          dat = bytes([resp_sid + 0x40]) + resp[1:] + DEFAULT_VIN.encode('utf-8')
+          print('Car ECU - sending back:', dat)
+          isotp_msg.send(dat)
+        else:
+          dat = bytes([0x7F, resp_sid, *resp[1:]])
+          print('Car ECU - bad data id, sending back:', dat)
+          isotp_msg.send(dat)
```
gregjhogan, line 59: I would have a lookup table something like `service` -> `subfunction (can be empty string)` -> `remaining data` -> `response message` with a function used to add entries to this lookup table.  If the service isn't found it should return the error `service not supported` and if the subfunction isn't found it should return `sub-function not supported`, etc...

adeebshihadeh:
Still useful or let's close?

sshane:
This is still very useful, as it caught a few bugs while writing. It requires a day/few days of work to make it clean and mergable though

PR #1330, opened 2023-04-05
HITL fan test

robbederks:
Test fan controller on embedded pandas

PR #1335, opened 2023-04-06
Dos fan fix

robbederks:
Two changes to pass the new HITL fan tests from #1330:

- Add a cooldown on the dos like on the tres to avoid spurious fan interrupts from a disabled spinning rotor

- Remove the feedforward term from the fan controller to reduce overshoot. The integrating term is more than fast enough from our purposes.

Before:
```
Fan power 10% overshoot: 0.02 Max RPM: 666
Fan power 20% overshoot: 0.26 Max RPM: 1641
Fan power 30% overshoot: 0.27 Max RPM: 2470
Fan power 40% overshoot: 0.27 Max RPM: 3311
Fan power 50% overshoot: 0.27 Max RPM: 4116
Fan power 60% overshoot: 0.25 Max RPM: 4866
Fan power 70% overshoot: 0.23 Max RPM: 5581
Fan power 80% overshoot: 0.20 Max RPM: 6234
Fan power 90% overshoot: 0.13 Max RPM: 6596
Fan power 100% overshoot: 0.02 Max RPM: 6605
```

After:
```
Fan power 10% overshoot: 0.12 Max RPM: 730
Fan power 20% overshoot: 0.06 Max RPM: 1383
Fan power 30% overshoot: 0.07 Max RPM: 2083
Fan power 40% overshoot: 0.06 Max RPM: 2757
Fan power 50% overshoot: 0.05 Max RPM: 3408
Fan power 60% overshoot: 0.03 Max RPM: 4028
Fan power 70% overshoot: 0.02 Max RPM: 4649
Fan power 80% overshoot: 0.01 Max RPM: 5262
Fan power 90% overshoot: 0.00 Max RPM: 5861
Fan power 100% overshoot: -0.00 Max RPM: 6475
```

Neither of these changes should affect the stall detection and recovery logic, but we might want to test this before shipping.


adeebshihadeh:
> Neither of these changes should affect the stall detection and recovery logic, but we might want to test this before shipping.

We definitely need to test this, and ideally put that test in CI. Minor regressions in stall recovery would be pretty bad.

tests/fan_overshoot_test.py
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/env python3
+
+import time
+from panda import Panda
+
+def get_overshoot_rpm(p, power):
```
adeebshihadeh, line 6: can we add this to the hitl tests?

robbederks: already in there!

robbederks:
> > Neither of these changes should affect the stall detection and recovery logic, but we might want to test this before shipping.
> 
> We definitely need to test this, and ideally put that test in CI. Minor regressions in stall recovery would be pretty bad.

I agree, but can't think of a way to easily test this in CI, unless we don't want to do it HITL, should be easy enough to just write unit tests for this with libpanda.

adeebshihadeh:
Removed clipping the integral to 50%, so this shouldn't be much different now.

PR #1338, opened 2023-04-08
README: fix scons error while trying to build

royjr:
Fixes `scons: not found` error when trying to build using the flash/recover scripts on a fresh install of Ubuntu 20.04

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ Setup your dependencies:
 
 ```bash
 # Ubuntu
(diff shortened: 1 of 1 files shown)


adeebshihadeh:
scons should be pip installed https://github.com/commaai/panda/blob/9e03d10857e456a4f0119f09de5a2bb76c9504f6/requirements.txt#L14. perhaps it's not in the PATH for some reason?

PR #1340, opened 2023-04-08
Tesla: move reading AEB state to RX hook

adeebshihadeh:
prep for #1339

### DIFF
--- a/board/safety/safety_tesla.h
+++ b/board/safety/safety_tesla.h
@@ -34,6 +34,7 @@ const CanMsg TESLA_PT_TX_MSGS[] = {
 #define TESLA_PT_TX_LEN (sizeof(TESLA_PT_TX_MSGS) / sizeof(TESLA_PT_TX_MSGS[0]))
 
 AddrCheckStruct tesla_addr_checks[] = {
+  {.msg = {{0x2b9, 0, 8, .expected_timestep = 40000U}, { 0 }, { 0 }}},   // DAS_control (25Hz)
   {.msg = {{0x370, 0, 8, .expected_timestep = 40000U}, { 0 }, { 0 }}},   // EPAS_sysStatus (25Hz)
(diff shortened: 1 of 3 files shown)


board/safety/safety_tesla.h
```diff
@@ -34,6 +34,7 @@ const CanMsg TESLA_PT_TX_MSGS[] = {
 #define TESLA_PT_TX_LEN (sizeof(TESLA_PT_TX_MSGS) / sizeof(TESLA_PT_TX_MSGS[0]))
 
 AddrCheckStruct tesla_addr_checks[] = {
+  {.msg = {{0x2b9, 0, 8, .expected_timestep = 40000U}, { 0 }, { 0 }}},   // DAS_control (25Hz)
```
sshane, line 37: @robbederks the safety replay route does not have this message, can we get a new one?

robbederks: this is `bb50caf5f0945ab1|2021-06-19--17-20-18--7`, same route as where the safety replay is taken from

sshane: looks like the safety looks for it on bus 0 though, which it's missing from

robbederks: oh I see, that is correct. This message comes from my autopilot computer (bus 128), and is forwarded to bus 0.
I assume it needs to be defined for bus 2 then

sshane: If this is a bug, I wonder how the safety tick didn't catch this and set controls allowed to False. Can you look into this? Or send a route on stock panda safety so I can briefly check? Won't have time for a bit

robbederks: don't have any current routes atm, but it shouldn't look any different than the one in replay now. Will take a look tomorrow

sshane: @robbederks did you look into this?

robbederks: That PR we merged the day after was fine, yes. Or did you mean anything else?

PR #1345, opened 2023-04-13
H7: Rx FIFO lost counter

briskspirit:
Previous implementation wasn't working as supposed.
Now it will show how many messages were lost due to get index offset and full FIFO.


sshane:
related to this, receive_error_cnt and transmit_error_cnt are not monotonic, they can jump up when an error occurs but then reset to 0 shortly after. can we fix that as well?

briskspirit:
This counter isn't what you think. It's CAN Core internal counter and it is incremented and then decremented when there is messages without error. It also related to passive and bus-off states. Will add comment to the code. (but it's already called REC and TEC)

PR #1353, opened 2023-04-20
Ford safety: curvature error limit

sshane:
By asserting a curvature command close to the current curvature (from yaw rate and two speed signals), we can limit torque wind up if openpilot were to command a high curvature far away from current


board/safety.h
```diff
@@ -409,7 +409,18 @@ bool max_limit_check(int val, const int MAX_VAL, const int MIN_VAL) {
   return (val > MAX_VAL) || (val < MIN_VAL);
 }
 
-// check that commanded value isn't too far from measured
+// check that commanded angle value isn't too far from measured
+bool angle_dist_to_meas_check(int val, struct sample_t *val_meas, const int MAX_ERROR, const int MAX_VAL) {
+
+  // val must always be near val_meas, limited the maximum value
```
sshane, line 415:
```suggestion
  // val must always be near val_meas, limited to the maximum value
```

board/safety.h
```diff
@@ -409,7 +409,19 @@ bool max_limit_check(int val, const int MAX_VAL, const int MIN_VAL) {
...
 }
 
-// check that commanded value isn't too far from measured
+// check that commanded angle value isn't too far from measured
+bool angle_dist_to_meas_check(int val, struct sample_t *val_meas, const int MAX_ERROR, const int MAX_VAL) {
+
+  // val must always be near val_meas, limited to the maximum value
+  // add 1 to not false trigger the violation
+  int highest_allowed = MIN(MAX(val_meas->max + MAX_ERROR + 1, -MAX_VAL), MAX_VAL);
+  int lowest_allowed = MIN(MAX(val_meas->min - MAX_ERROR - 1, -MAX_VAL), MAX_VAL);
```
adeebshihadeh, line 418: use the new clamp macro https://github.com/commaai/panda/pull/1356

sshane: 🙌

tests/safety/test_ford.py
```diff
@@ -69,6 +80,11 @@ def setUp(self):
     self.safety.set_safety_hooks(Panda.SAFETY_FORD, 0)
     self.safety.init_tests()
 
+  def _curvature_meas_msg_array(self, curvature, speed):
```
adeebshihadeh, line 83: i'd call this something like reset_curvature_measurements

adeebshihadeh (commented):
Let's also double check the speed with another signal

tests/safety/test_ford.py
sshane: just copied the convention for angle

PR #1358, opened 2023-04-26
Ford: check second speed for mismatches

sshane:
For https://github.com/commaai/panda/pull/1353

Check https://github.com/commaai/openpilot/pull/27981 for data about the max delta across the fleet and the reason we set it so high.

- Asserts a max delta between speeds from the ABS and PCM modules (ABS is raw, PCM is a filtered version)
- Sets controls allowed to false for as long as the two are mismatching
- Test to assert this threshold


board/safety/safety_ford.h
```diff
@@ -135,6 +150,14 @@ static int ford_rx_hook(CANPacket_t *to_push) {
       vehicle_speed = ((GET_BYTE(to_push, 0) << 8) | GET_BYTE(to_push, 1)) * 0.01 / 3.6;
     }
 
+    if (addr == MSG_EngVehicleSpThrottle2) {
+      // Set vehicle state mismatch if speeds from ABS and PCM ECUs are far enough apart.
```
sshane, line 154:
```suggestion
      // Set vehicle state mismatch if speeds from ABS and PCM ECUs are too far apart.
```

board/safety/safety_ford.h
```diff
@@ -135,6 +150,14 @@ static int ford_rx_hook(CANPacket_t *to_push) {
       vehicle_speed = ((GET_BYTE(to_push, 0) << 8) | GET_BYTE(to_push, 1)) * 0.01 / 3.6;
     }
 
+    if (addr == MSG_EngVehicleSpThrottle2) {
+      // Set vehicle state mismatch if speeds from ABS and PCM ECUs are too far apart.
+      // Causes controls allowed to fall
+      // Signal: Veh_V_ActlEngz
```
sshane, line 156:
```suggestion
      // Signal: Veh_V_ActlEng
```

sshane:
For this, do we want to set a permanent fault or should the user be able to re-engage? Right now, it allows the user to re-engage if the mismatch has cleared

board/safety/safety_ford.h
```diff
@@ -78,6 +86,11 @@ static uint32_t ford_compute_checksum(CANPacket_t *to_push) {
     chksum += GET_BYTE(to_push, 2) >> 6;                    // VehVActlBrk_D_Qf
     chksum += (GET_BYTE(to_push, 2) >> 2) & 0xFU;           // VehVActlBrk_No_Cnt
     chksum = 0xFFU - chksum;
+  } else if (addr == MSG_EngVehicleSpThrottle2) {
+    chksum += (GET_BYTE(to_push, 2) >> 3) & 0xFU;            // VehVActlEng_No_Cnt
+    chksum += (GET_BYTE(to_push, 4) >> 5) & 0x3U;            // VehVActlEng_D_Qf
+    chksum += GET_BYTE(to_push, 6) + GET_BYTE(to_push, 7);  // Veh_V_ActlEng
```
sshane, line 92:
```suggestion
    chksum += GET_BYTE(to_push, 6) + GET_BYTE(to_push, 7);   // Veh_V_ActlEng
```

sshane:
I'm not convinced we need an explicit `vehicle_state_mismatch` global in this PR, since setting `controls_allowed` directly is the exact same behavior. We can assert the exact same behavior, and less complexity in this PR.

However I opened https://github.com/commaai/panda/pull/1366 which we can merge after as a clean up and remove the `controls_allowed` setting here once more thought is given to this.

adeebshihadeh:
> I'm not convinced we need an explicit `vehicle_state_mismatch` global in this PR, since setting `controls_allowed` directly is the exact same behavior and other safety models also directly interface with controls_allowed.

We want to move away from this; the car-specific safety modes should really just be parsers.

PR #1359, opened 2023-04-26
Ford: add longitudinal safety

sshane:


board/safety/safety_ford.h
```diff
@@ -104,13 +116,25 @@ static bool ford_get_quality_flag_valid(CANPacket_t *to_push) {
   return valid;
 }
 
+const uint16_t FORD_LONG_CONTROL = 1;
```
sshane, line 119:
```suggestion
const uint16_t FORD_PARAM_LONGITUDINAL = 1;
```

board/safety/safety_ford.h
```diff
@@ -249,8 +296,7 @@ static int ford_fwd_hook(int bus_num, int addr) {
 }
 
 static const addr_checks* ford_init(uint16_t param) {
-  UNUSED(param);
-
+  ford_longitudinal = GET_FLAG(param, FLAG_FORD_LONG_CONTROL);
```
sshane, line 299:
```suggestion
#ifdef ALLOW_DEBUG
  ford_longitudinal = GET_FLAG(param, FLAG_FORD_LONG_CONTROL);
#else
  hyundai_longitudinal = false;
#endif
```

we put this behind the debug flag so you can only enable on non-release panda

board/safety/safety_ford.h
```diff
@@ -120,6 +132,24 @@ static bool ford_get_quality_flag_valid(CANPacket_t *to_push) {
...
+  .min_accel = 4230,       // -3.5    m/s^2
+  .inactive_accel = 5128,  // -0.0008 m/s^2
+
+  // gas cmd limits
+  // Signal: AccPrpl_A_Rq
+  .max_gas = 700,          //  2.5 m/s^2
+  .min_gas = 0,
+  .inactive_gas = 0,       // -5.0 m/s^2
+};
+
```
sshane, line 152: it looks like it simplifies things a lot if we make it just gas/brake again

incognitojam: the "brake signal" is in accel though, so pos/neg are wrong way around for brake

PR #1360, opened 2023-04-26
safety: don't check out of bounds array item

sshane:
@adeebshihadeh if there are three messages in the array (which is statically allocated length of 3 `const CanMsgCheck msg[3];`), then the addr of the fourth array item is not guaranteed to be 0 since it's out of bounds, right?

### DIFF
--- a/board/safety.h
+++ b/board/safety.h
@@ -139,7 +139,7 @@ int get_addr_check_index(CANPacket_t *to_push, AddrCheckStruct addr_list[], cons
   for (int i = 0; i < len; i++) {
     // if multiple msgs are allowed, determine which one is present on the bus
     if (!addr_list[i].msg_seen) {
-      for (uint8_t j = 0U; addr_list[i].msg[j].addr != 0; j++) {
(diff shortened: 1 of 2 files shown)


board/safety.h
```diff
@@ -139,7 +139,7 @@ int get_addr_check_index(CANPacket_t *to_push, AddrCheckStruct addr_list[], cons
   for (int i = 0; i < len; i++) {
     // if multiple msgs are allowed, determine which one is present on the bus
     if (!addr_list[i].msg_seen) {
-      for (uint8_t j = 0U; addr_list[i].msg[j].addr != 0; j++) {
+      for (uint8_t j = 0U; addr_list[i].msg[j].addr != 0 && j < 3; j++) {
```
sshane, line 142:
```suggestion
      for (uint8_t j = 0U; j < 3 && addr_list[i].msg[j].addr != 0; j++) {
```

adeebshihadeh:
Yes, but this should be fixed more carefully. Running with UBSan catches this: #1363

PR #1364, opened 2023-04-27
safety replay: safety tick

sshane:
similar to test_models


sshane:
@robbederks this PR causes those blocked message in safety replay

robbederks:
> @robbederks this PR causes those blocked message in safety replay

Which blocked messages?

sshane:
```
TX
total openpilot msgs: 6103
total msgs with controls allowed: 0
blocked msgs: 4978
blocked with controls allowed: 0
blocked addrs: Counter({1160: 4978})
```

before there were messages with controls allowed

adeebshihadeh:
Any reason this isn't merged yet? The tests are passing.

sshane:
The tests aren't great. This would set controls allowed to False if this triggered, and safety replay asserts no messages are blocked while controls allowed... Manually checking the replay results is how I found the Tesla bug, which test_models also would've caught if it weren't for dashcam mode (https://github.com/commaai/panda/pull/1374). I'll ad somethin like test_models's `addr_checks_valid` and should be good to merge

PR #1366, opened 2023-04-27
safety: vehicle state mismatch global

sshane:
- add new `vehicle_state_mismatch`, a global that safety models can set to force controls allowed generically
- common test that asserts `vehicle_state_mismatch` behavior, safety models can then assert that certain conditions make `vehicle_state_mismatch` rise without testing `controls_allowed` directly, since the common test tests that


tests/libpanda/safety_helpers.py
```diff
@@ -46,6 +48,8 @@ def setup_safety_helpers(ffi):
 class PandaSafety:
   def set_controls_allowed(self, c: bool) -> None: ...
   def get_controls_allowed(self) -> bool: ...
+  def set_vehicle_state_mismatch(self, m) -> None: ...
```
sshane, line 51:
```suggestion
  def set_vehicle_state_mismatch(self, m: bool) -> None: ...
```

sshane:
@adeebshihadeh think we might want to add a `vehicleStateMismatch` event in openpilot as well to go along with this, but not sure if there's something more generic you can think of?

adeebshihadeh (approved):

sshane:
Think we can wait until two brands/safety modes use this

PR #1369, opened 2023-04-27
safety: common angle safety function

sshane:
angle and curvature are linearly proportional, so we treat them as the same thing for ford. this makes the safety use one common function

### DIFF
--- a/board/safety.h
+++ b/board/safety.h
@@ -644,14 +644,11 @@ bool steer_angle_cmd_checks(int desired_angle, bool steer_control_enabled, const
     }
   }
 
-  // Angle should be the same as current angle while not steering
-  if (!limits.disable_near_angle_check) {
-    violation |= (!steer_control_enabled &&
+  // Angle should either be 0 or same as current angle while not steering
+  if (!steer_control_enabled) {
+    violation |= (limits.inactive_angle_is_zero ? (desired_angle != 0) :
                     ((desired_angle < (angle_meas.min - 1)) ||
                     (desired_angle > (angle_meas.max + 1))));
-  } else {
-    // If steer control is not enabled, curvature must be 0
-    violation |= (!steer_control_enabled && (desired_angle != 0));
   }
 
   // No angle control allowed when controls are not allowed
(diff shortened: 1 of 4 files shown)


board/safety_declarations.h
```diff
@@ -69,6 +69,11 @@ typedef struct {
   const struct lookup_t angle_rate_up_lookup;
   const struct lookup_t angle_rate_down_lookup;
   const int max_angle_error;  // used to limit error between meas and cmd
+
+  const bool disable_rate_limits;
+  const bool enforce_angle_error;
+  const bool disable_near_angle_check;
```
sshane, line 75: made the defaults the opposite of what ford uses, so no change in other safety modes are needed

board/safety.h
```diff
@@ -643,10 +631,28 @@ bool steer_angle_cmd_checks(int desired_angle, bool steer_control_enabled, const
...
+  }
+
   // Angle should be the same as current angle while not steering
-  violation |= (!controls_allowed &&
-                  ((desired_angle < (angle_meas.min - 1)) ||
-                  (desired_angle > (angle_meas.max + 1))));
+  if (!limits.disable_near_angle_check) {
+    violation |= (!controls_allowed &&
+                    ((desired_angle < (angle_meas.min - 1)) ||
+                    (desired_angle > (angle_meas.max + 1))));
```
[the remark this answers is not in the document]
sshane: done! https://github.com/commaai/panda/pull/1372

board/safety.h
```diff
@@ -643,10 +631,25 @@ bool steer_angle_cmd_checks(int desired_angle, bool steer_control_enabled, const
...
+      violation |= max_limit_check(desired_angle, highest_allowed, lowest_allowed);
+    }
+  }
+
+  // Angle should either be 0 or same as current angle while not steering
+  if (!steer_control_enabled) {
+    violation |= (limits.inactive_angle_is_zero ? (desired_angle != 0) :
+                    ((desired_angle < (angle_meas.min - 1)) ||
+                    (desired_angle > (angle_meas.max + 1))));
+  }
```
sshane, line 652: seems like a cppcheck bug (`(desired_angle != false) :` passes), this should be compliant 😞. not fixed on latest master either

PR #1370, opened 2023-04-27
ford: allow 0 curvature when controls is allowed if not steering

sshane:
the case this allows is similar to torque control limiting:

- new behavior: if EPS is faulted and controls are allowed (or some other condition to make OP not want to steer), safety should allow OP to send 0 curvature with `steer_control_type=False`
- previous behavior: it only checks `controls_allowed`, so angle error limits likely will block msgs

### DIFF
--- a/board/safety/safety_ford.h
+++ b/board/safety/safety_ford.h
@@ -262,6 +262,10 @@ static int ford_tx_hook(CANPacket_t *to_send) {
       }
     }
 
+    if (!steer_control_enabled && (desired_curvature != 0)) {
+      violation = true;
+    }
+
     // No curvature command if controls is not allowed
     if (!controls_allowed && ((desired_curvature != 0) || steer_control_enabled)) {
       violation = true;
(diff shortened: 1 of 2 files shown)


board/safety/safety_ford.h
```diff
@@ -255,13 +255,18 @@ static int ford_tx_hook(CANPacket_t *to_send) {
...
         violation |= angle_dist_to_meas_check(desired_curvature, &angle_meas,
                                               FORD_STEERING_LIMITS.max_angle_error, FORD_STEERING_LIMITS.max_steer);
       }
     }
 
+    // If steer control is not enabled, curvature must be 0
+    if (!steer_control_enabled && (desired_curvature != 0)) {
+      violation = true;
+    }
+
```
sshane, line 269: Would be nice if we can just copy what commonangle safety does (always send it and assert it matches meas), but I don't think that can work here.

1. Curvature is limited to 0.02, which is much lower than what you can achieve,
2. and the system takes a while to start applying torque if desired curvature is non-zero when you set the apply bit

So we'll have to handle this case when we merge with common angle safety in https://github.com/commaai/panda/pull/1369

PR #1372, opened 2023-04-28
angle safety: enforce disabled angle when not steering

sshane:
Previously we were enforcing a disabled angle only when controls were not allowed. When `steer_control_enabled` bit is 0, you could set it to whatever you wanted. Not a huge deal unless an EPS has a bug where it actuates on that, or winds up torque on set of the bit again.

todo:
- [x] test

split from https://github.com/commaai/panda/pull/1369

### DIFF
--- a/board/safety.h
+++ b/board/safety.h
@@ -644,7 +644,7 @@ bool steer_angle_cmd_checks(int desired_angle, bool steer_control_enabled, const
   desired_angle_last = desired_angle;
 
   // Angle should be the same as current angle while not steering
-  violation |= (!controls_allowed &&
+  violation |= (!steer_control_enabled &&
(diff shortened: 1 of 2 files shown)


board/safety.h
```diff
@@ -644,7 +644,7 @@ bool steer_angle_cmd_checks(int desired_angle, bool steer_control_enabled, const
   desired_angle_last = desired_angle;
 
   // Angle should be the same as current angle while not steering
-  violation |= (!controls_allowed &&
+  violation |= (!steer_control_enabled &&
```
sshane, line 647: This is similar to the torque check:

```c
bool steer_req_mismatch = (steer_req == 0) && (desired_torque != 0);
```

PR #1374, opened 2023-04-28
DAS_control is on bus 2

robbederks:
Was broken in #1340. Haven't driven on it to test yet, but this is definitely more correct.


sshane:
This makes the panda address checks pass in test_models, looks good to me!

robbederks:
Let's merge then. Will fix if needed once my car is back up and running

PR #1376, opened 2023-04-28
safety: sample_t is not fully reset on safety mode init

sshane:
The safety mode set just sets the min and max, it doesn't clear out the values array, so on rx of one new measurement, it will use the previously stored values to update the sample_t min and max

this is just a test to catch it, moving the base test to https://github.com/commaai/panda/pull/1250


board/safety.h
```diff
@@ -345,6 +345,12 @@ int set_safety_hooks(uint16_t mode, uint16_t param) {
   valid_steer_req_count = 0;
   invalid_steer_req_count = 0;
 
+  for (int i = 0; i < MAX_SAMPLE_VALS; i++) {
+    torque_meas.values[i] = 0;
+    torque_driver.values[i] = 0;
+    angle_meas.values[i] = 0;
+  }
+
```
sshane, line 353: can also use memset?

tests/safety/common.py
```diff
@@ -634,6 +638,8 @@ def test_reset_angle_measurements(self):
     # reset sample_t by reinitializing the safety mode
     self.setUp()
 
+    # rx one to update sample_t
+    self.assertTrue(self._rx(self._angle_meas_msg(0)))
```
adeebshihadeh, line 642: why is the RX necessary?

tests/safety/common.py
```diff
@@ -634,6 +638,8 @@ def test_reset_angle_measurements(self):
...
     self.assertEqual(self.safety.get_angle_meas_min(), 0)
     self.assertEqual(self.safety.get_angle_meas_max(), 0)
```
adeebshihadeh, line 644: add an assert before the safety mode reset that it's not the reset value

board/safety.h
adeebshihadeh: for loop is more readable

tests/safety/common.py
sshane: because receiving a message calls `update_sample` which sets the min and max from the non-zero sample array

adeebshihadeh: that should also be reset on safety mode init

sshane: That's what this does https://github.com/commaai/panda/pull/1376/files#diff-9e71d0935391e66564e1fe068f0e9c0b7a2523e5a31c96d2556205ee60b6d15aR347-R352

sshane: done a little cleaner

PR #1379, opened 2023-04-29
Subaru: allow Throttle TX on bus 2 for stop and go

martinl:
This PR is prerequisite for adding Subaru SNG support for both global and preglobal models.

Subaru stock ACC stops behind stopped cars but does not resume automatically. We send Throttle:Throttle_Pedal signal from openpilot to ACC when car in front starts moving to resume from ACC Hold state (mimicking stock ACC resume using gas press behaviour).

We also tried to resume using ACC resume button signals but that did not work since ACC buttons are directly wired to Eyesight and ACC does not accept the button press input signals from CAN to initiate resume.

### DIFF
--- a/board/safety/safety_subaru.h
+++ b/board/safety/safety_subaru.h
@@ -26,6 +26,7 @@ const CanMsg SUBARU_TX_MSGS[] = {
   {0x321, 0, 8},
   {0x322, 0, 8},
   {0x323, 0, 8},
+  {0x40,  2, 8},
 };
 #define SUBARU_TX_MSGS_LEN (sizeof(SUBARU_TX_MSGS) / sizeof(SUBARU_TX_MSGS[0]))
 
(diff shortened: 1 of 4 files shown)


sshane:
Is it possible to instead spam the message when needed, not forward it through openpilot? There also may be ramifications by pulling the message through OP, like affecting AEB or FCW if the message is dropped or changed in an undefined way.

For example with GM, we block and send our own driver torque EPS message to the camera to silence an alert, but I made sure that the docs mentioned it didn't take driver torque into account for AEB, as well as tested it still could AEB with the message fully blocked from the camera. Do you know anything about this here?

martinl:
Eyesight is using Throttle message signal as input to disengage but I do not have any documentation that would describe the functions more in-depth than owners manual. Is there a safe way to test AEB without renting a test track and/or buying mock targets (which would probaly not fit into my hobby projects budget)?

martinl:
I'll also try if spamming the message without dropping it in panda safety works

sshane:
I found the description of the conditions and what sensors it uses for AEB in the GM Tech Info manual

martinl:
I have reviewed a few Subaru Techinfo service manuals and it has very little details on Eyesight.

Service manual does have a pre-collision brake (AEB) diagnostics checklist that lists conditions when AEB does not work, including:
- Crash avoidance operation (accelerator pedal, brake pedal, steering wheel, etc.) was performed

Also, if one of the signals (eg Throttle) goes missing or is blocked by panda then Eyesight faults immediately and I'm pretty sure AEB does not work after that

PR #1380, opened 2023-04-30
Subaru: improve readability in panda code

jnewb1:


sshane:
Thanks for the PR. I like what ford and VW do (prefix with `MSG_`, and make it a macro). We should start to be consistent across the safety modes.

board/safety/safety_subaru.h
```diff
@@ -20,40 +20,56 @@ const SteeringLimits SUBARU_GEN2_STEERING_LIMITS = {
   .type = TorqueDriverLimited,
 };
 
+const int Brake_Status =  0x13c;
+const int CruiseControl = 0x240;
+const int Throttle = 0x40;
+const int Steering_Torque = 0x119;
+const int Wheel_Speeds = 0x13a;
+
+const int OPENPILOT_STATUS = 0x100;
```
sshane, line 29: What is this?

jnewb1: mistake, was for gen2 long with second panda

jnewb1: removed

jnewb1:
> Thanks for the PR. I like what ford and VW do (prefix with `MSG_`, and make it a macro). We should start to be consistent across the safety modes like this.

i refactored to match the ford template

sshane:
Looks good, can you do legacy Subaru as well?

jnewb1:
> Looks good, can you do legacy Subaru as well?

added preglobal. we are going to have problem if we continue with the ford-way if there are any names that are the same

ex: MSG_Wheel_Speeds will probably be used in other cars too

maybe something like

MSG_SUBARU_Wheel_Speeds would be better?

jnewb1:
synced with master

PR #1389, opened 2023-05-02
angle safety: add comment about rate limits at 0

sshane:
Related to https://github.com/commaai/panda/pull/1388

Originally at 0 degrees, we'd allow up in one direction and down in another: https://github.com/commaai/panda/blob/11ef24bc1642e5f6787dda2c1ff6a8a3c115c24a/board/safety/safety_toyota_ipas.h#L129-L130

Then in this Tesla safety refactor, it was switched to allow down in both directions: https://github.com/commaai/panda/pull/629

And Nissan copied Tesla's safety and tests.


sshane:
This seems like it actually allows some purposeful tolerance, I forgot about the bug I found in: https://github.com/commaai/openpilot/pull/27428#issuecomment-1442568717 (second issue)

Basically if the float in openpilot is very small and positive, we'd use the down limits to ramp to negative values, above the up rate limit causing a false violation. Fixable by using ints  in openpilot, or rounding in a way that is represented on CAN, but for now I'll add a comment since this seems like it was intentional.

PR #1391, opened 2023-05-03
safety: make vehicle_speed a sample

sshane:


sshane:
@adeebshihadeh did you mean this?

tests/safety/common.py
```diff
@@ -562,6 +563,10 @@ def _reset_angle_measurement(self, angle):
     for _ in range(6):
       self._rx(self._angle_meas_msg(angle))
 
+  def _reset_speed_measurement(self, speed):
+    for _ in range(6):
+      self._rx(self._speed_msg(speed))  # pylint: disable=no-member
```
adeebshihadeh, line 568: why is the pylint exception needed?

tests/safety/common.py
```diff
@@ -637,6 +642,26 @@ def test_reset_angle_measurements(self):
...
+    for speed in np.arange(0, 40, 0.5):
+      for i in range(6):
+        self.assertTrue(self._rx(self._speed_msg(speed + i * 0.1)))  # pylint: disable=no-member
+
+      # assert close by one decimal place
+      self.assertLessEqual(abs(self.safety.get_vehicle_speed_min() - speed * VEHICLE_SPEED_FACTOR), 1)
+      self.assertLessEqual(abs(self.safety.get_vehicle_speed_max() - (speed + 0.5) * VEHICLE_SPEED_FACTOR), 1)
+
+      # reset sample_t by reinitializing the safety mode
+      self.setUp()
```
adeebshihadeh, line 660: do this more explicitly. currently, there happens to be no other setup besides the safety mode init, but that's not a safe assumption

tests/safety/common.py
```diff
@@ -637,6 +642,26 @@ def test_reset_angle_measurements(self):
...
 
+  def test_vehicle_speed_measurements(self):
+    """
+    Tests:
+     - rx hook correctly parses and rounds the vehicle speed
+     - sample is reset on safety mode init
+    """
+    for speed in np.arange(0, 40, 0.5):
+      for i in range(6):
+        self.assertTrue(self._rx(self._speed_msg(speed + i * 0.1)))  # pylint: disable=no-member
```
adeebshihadeh, line 653: same as above

tests/safety/common.py
```diff
@@ -637,6 +642,26 @@ def test_reset_angle_measurements(self):
     self.assertEqual(self.safety.get_angle_meas_min(), 0)
     self.assertEqual(self.safety.get_angle_meas_max(), 0)
 
+  def test_vehicle_speed_measurements(self):
+    """
+    Tests:
+     - rx hook correctly parses and rounds the vehicle speed
+     - sample is reset on safety mode init
+    """
+    for speed in np.arange(0, 40, 0.5):
```
adeebshihadeh, line 651: raise the max?

sshane: raised to 80, that's the max the tesla signal can represent (Nissan is 90 m/s)

tests/safety/common.py
sshane: This test is done in the angle safety as no other safety mode needs to set the vehicle speed,  `_speed_msg` is defined as abstract in the base panda safety test. However, it shouldn't violate anything to also define _speed_msg as abstract in the angle safety class so the dependency is explicit

tests/safety/common.py
sshane: https://github.com/commaai/panda/pull/1398

PR #1397, opened 2023-05-03
safety utils: add ROUND helper

sshane:
for when casting float to int and you want to round around 0.5


board/utils.h
```diff
@@ -22,6 +22,11 @@
   (_a > 0) ? _a : (-_a); \
 })
 
+#define ROUND(a) ({ \
+  __typeof__ (a) _a = (a); \
+  _a + ((_a > 0) ? 0.5 : -0.5); \
```
adeebshihadeh, line 27: this just does the addition and results in a float. no need for a macro here either

sshane, line 27:
```suggestion
  (int)(_a + ((_a > 0) ? 0.5 : -0.5)); \
```

sshane: You mean you don't see a need for the macro, or the typeof operator is not needed?

board/utils.h
```diff
@@ -22,6 +22,11 @@
   (_a > 0) ? _a : (-_a); \
 })
 
+#define ROUND(a) ({ \
```
adeebshihadeh, line 25: this is cleaner as a function, since the input/output types are fixed

PR #1405, opened 2023-05-08
harness: fix register divergent on GEN1

adeebshihadeh:


robbederks:
Hmm, let's maybe just return early on non-harness pandas in this function. Didn't realize it was always called

PR #1409, opened 2023-05-10
macOS: use homebrew gcc using version suffix. update setup instructions

fredyshox:
There're build issues of libtest on macos caused by it using system gcc, which is just a symlink to clang. The previous PRs in that regard (e.g. https://github.com/commaai/panda/pull/1241) proposed installing gcc@12 from brew and updating the links on your own, but IMO changing the system config shouldn't be right thing to do. Since we require installation of gcc@12, we can use it explicitly by invoking `gcc-12`.


tests/SConscript
fredyshox, line 1: @adeebshihadeh can this file be deleted? it does not seem to do anything

adeebshihadeh: no, it's used for building `libpanda`

fredyshox: but it references non-existing files

adeebshihadeh: ah, it is unused now. libpanda's SConscript has moved to tests/libpanda/

fredyshox: ok, gonna delete it then

SConscript
```diff
@@ -1,6 +1,16 @@
...
 
 # test files
 if GetOption('test'):
+  CC = 'gcc'
+  platform = platform.system()
+  if platform == 'Darwin':
+    # gcc installed by homebrew has version suffix (e.g. gcc-12) in order to be  
+    # distinguishable from system one - which acts as a symlink to clang
+    CC += '-12'
+  Export('CC', 'platform')
```
adeebshihadeh, line 14: this should all be done in `tests/libpanda/SConscript`

SConscript
```diff
@@ -1,6 +1,16 @@
...
 SConscript('board/SConscript')
 
 # test files
 if GetOption('test'):
+  CC = 'gcc'
+  platform = platform.system()
+  if platform == 'Darwin':
+    # gcc installed by homebrew has version suffix (e.g. gcc-12) in order to be  
+    # distinguishable from system one - which acts as a symlink to clang
+    CC += '-12'
```
adeebshihadeh, line 13: is there a more robust way to specify the brew version of gcc?

fredyshox: it does not do that (in [formula](https://github.com/Homebrew/homebrew-core/blob/78c6f5303d1b91f0a09d0882835ef62791e661f0/Formula/gcc.rb), every installed gcc binary has a version suffix), so it could be distinguished from gcc/clang supplied by Xcode.

fredyshox: I will have to take a look one more time to confirm, but I believe the suffixes are also there in the formula install directory at `brew --prefix gcc`

PR #1416, opened 2023-05-17
safety: common gas/brake tests

sshane:
- [ ] remove GM's existing tests and combine
- [ ] handle honda


sshane:
going to want on this until Ford is in, since there are multiple different ways to write this, and having ford, gm, and honda all together will help me see how we can do that (for ex. honda tests gas and accel, but together, gm tests gas and brake as separate messages, and ford tests gas and accel together)

sshane:

PR #1419, opened 2023-05-17
safety: rewrite longitudinal accel checks function

sshane:
redistributes dense lines of logic, easier to understand


sshane:
eh about the same

PR #1423, opened 2023-05-18
tests: common setUpClass function

sshane:
One global `setUpClass` that sets cls.safey and skips testing classes ending with `Base`


sshane:
unclear if this is a massive improvement, not yet needed

PR #1453, opened 2023-06-10
CAN health: add IRQ call and core reset counters

briskspirit:
Would be helpful to have interrupt counter in the CAN health packet
Also data should be updated on request (except when error interrupt was fired). With previous method if CAN core stopped to fire interrupts - health packet won't be updated.

### DIFF
--- a/board/drivers/fdcan.h
+++ b/board/drivers/fdcan.h
@@ -73,7 +73,7 @@ void update_can_health_pkt(uint8_t can_number, uint32_t ir_reg) {
       can_health[can_number].total_rx_lost_cnt += 1U;
     }
     // Reset CAN core when TX error couter reaches at least 100 errors
-    if (can_health[can_number].transmit_error_cnt >= 100U) {
+    if (((ir_reg & (FDCAN_IR_PED | FDCAN_IR_PEA)) != 0) && (((ecr_reg & FDCAN_ECR_CEL) >> FDCAN_ECR_CEL_Pos) >= 100U)) {
       can_health[can_number].can_core_reset_cnt += 1U;
       can_health[can_number].total_tx_lost_cnt += (FDCAN_TX_FIFO_EL_CNT - (FDCAN_TXFQS_TFFL & FDCAN_TXFQS_TFFL_Msk)); // TX FIFO msgs will be lost after reset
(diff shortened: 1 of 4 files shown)


python/__init__.py
```diff
@@ -716,6 +719,10 @@ def get_uid(self):
   def get_secret(self):
     return self._handle.controlRead(Panda.REQUEST_IN, 0xd0, 1, 0, 0x10)
 
+  def get_interrupt_load(self, irqnum):
```
adeebshihadeh, line 722:
```suggestion
  def get_interrupt_call_rate(self, irqnum):
```

board/main_comms.h
```diff
@@ -181,6 +182,17 @@ int comms_control_handler(ControlPacket_t *req, uint8_t *resp) {
       (void)memcpy(resp, ((uint8_t *)UID_BASE), 12);
       resp_len = 12;
       break;
+    case 0xc4:
+      // **** 0xc4: get interrupt load rate
```
adeebshihadeh, line 186:
```suggestion
      // **** 0xc4: get interrupt call rate
```

board/health.h
```diff
@@ -55,4 +55,7 @@ typedef struct __attribute__((packed)) {
   uint8_t canfd_enabled;
   uint8_t brs_enabled;
   uint8_t canfd_non_iso;
+  uint32_t it0_call_rate;
```
adeebshihadeh, line 58: irq instead of it?

adeebshihadeh (changes requested):

board/health.h
briskspirit: ok makes sense, that's just how ST named it.

board/drivers/fdcan.h
```diff
@@ -11,6 +11,12 @@ typedef struct {
 
 FDCAN_GlobalTypeDef *cans[] = {FDCAN1, FDCAN2, FDCAN3};
 
+uint8_t can_irq_number[3][3] = {
+  { FDCAN1_IT0_IRQn, FDCAN1_IT1_IRQn, 0 },
```
adeebshihadeh, line 15: why 3 elements here?

PR #1457, opened 2023-06-14
UART overwrite mode

robbederks:
To make sure we don't lose the last logs coming from the SOM


robbederks:
this ring buffer code is copied in so many places, it's in need of a proper cleanup at some point

nidzeta:
Is this issue still active ? @robbederks

PR #1461, opened 2023-06-14
Ford: don't check vehicle roll quality flag

incognitojam:
The roll signal isn't used in openpilot, and isn't populated on the Ford Focus Mk4 (it is always [`6.6086` No data exists](https://github.com/commaai/opendbc/blob/master/ford_lincoln_base_pt.dbc#L12423)). We still have the yaw signal.

Route: `e886087f430e7fe7|2023-06-14--13-19-55`

image


sshane:
We check the roll flag in test_ford, can you briefly see why it didn't fail?

incognitojam:
It didn't fail because we don't check the roll flag independently from the yaw flag - when we set both to "faulty", we expect the tx to fail, and it does.

sshane (approved):
Can you make another PR to add a safety replay route for your Ford?

incognitojam:
I'm going to get a new route tomorrow that I can put in the public bucket

PR #1464, opened 2023-06-17
add Ford Focus Mk4 safety replay segment

incognitojam:
note: the test should fail until https://github.com/commaai/panda/pull/1461 is merged

`e886087f430e7fe7|2023-06-16--22-57-15--6`

EDIT: still seems to fail - I chose a segment with steer overriding so might be tripping on that somehow


sshane:
@incognitojam is it worth it to have the Focus and Bronco, are they different in any way?

incognitojam:
They are pretty similar (same C2 platform). Only difference I can think of is I don't have an automatic transmission. I guess other sensors might be missing like [the vehicle roll](https://github.com/commaai/panda/pull/1461) was.

jnewb1:
safety replay is removed

PR #1468, opened 2023-06-19
More logs

robbederks:


adeebshihadeh:
Still want this? We've already got bootkick and boot time, not sure the rest is necessary.

robbederks:
Not really. Ignition might prove to be useful for some rare cases, but probably not worth it until we run into anything

PR #1475, opened 2023-06-21
query_fw_versions: ignore InvalidSubFunctionError

sshane:
ignores when setting diag mode, negative response is similar to  catch this error after we fix TypeError in: https://github.com/commaai/panda/pull/1474


sshane:
ehh fine to not catch for now

PR #1485, opened 2023-06-27
Disable CAN irqs when transceivers are disabled in power save mode

briskspirit:


tests/hitl/2_health.py
```diff
@@ -94,7 +94,9 @@ def test_hw_type(p):
     assert pp.get_mcu_type() == mcu_type, "Bootstub and app MCU type mismatch"
     assert pp.get_uid() == app_uid
 
+@pytest.mark.expected_logs(1)
 def test_heartbeat(p, panda_jungle):
+  p.reset()
```
adeebshihadeh, line 99: why is this needed? this is already done in conftest.py

PR #1487, opened 2023-06-30
iso-tp: more checks

sshane:
asserts we can't get multiple single or first frames after calling `send` on IsoTpMessage. with this, we can extend timeouts in openpilot after every response, not just consecutive frames.

@gregjhogan do you think it is a good idea to make the user of the class to need to call `send(b'', setup_only=True)` (or similar reset function) every time if they just want to receive?

right now you can continually call `iso_tp_message.recv()` and continually get first or single frames, not sure if that's the intended behavior


gregjhogan:
> @gregjhogan do you think it is a good idea to make the user of the class to need to call `send(b'', setup_only=True)` (or similar reset function) every time if they just want to receive?

I think this is fine, sending the initial messages out of band is definitely not the normal situation and I don't have any better ideas on how to handle it.

> right now you can continually call `iso_tp_message.recv()` and continually get first or single frames, not sure if that's the intended behavior

The asserts are good, something has gone wrong if either get hit.

PR #1489, opened 2023-07-03
SPI: send serial with protocol version

adeebshihadeh:
Allows listing/enumeration even if the protcols don't match


adeebshihadeh:
@robbederks anything else you think we might want here?

robbederks:
Whether we are in bootstub and maybe the panda type?

PR #1491, opened 2023-07-03
CI: Upload hitl test report to codecov

4ydan:
Giving this [issue ](https://github.com/commaai/openpilot/issues/28570) a try.
I am not super aware of the whole codebase, so idk if I can run the HITL tests in github actions. I see that they are executed in Jenkins but if GA works, I probably can manage the coverage report.

Edit:
As I expected, cant run without hardware in actions? I'am also not sure if this is the right approach. This would need coverage to be installed and to pass the token to it. Not as easy as I thought.

Edit:
And I dont think I have the setup to run Jenkins locally. I'm missing various secrets probably.
image

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -63,8 +63,6 @@ jobs:
         run: $RUN "cd tests/usbprotocol && ./test.sh"
       - name: Test logging
         run: $RUN "cd tests/logging && ./test.sh"
-      - name: Test hitl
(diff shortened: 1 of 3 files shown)


adeebshihadeh:
HITL codecov won't really work for these tests. That issue was for the openpilot repo's HITL tests, but that needs to be done by someone internal anyway. 

If you'd like a quick project in this repo, I think it would be nice to enforce full mypy coverage in the python/ directory.

PR #1494, opened 2023-07-05
Increase mypy coverage

4ydan:
As requested in https://github.com/commaai/panda/pull/1491#issuecomment-1620728137, this is aiming to increase the mypy coverage.


4ydan:
I force pushed to rebase my commits
It looks like you already have a mypy check in the python linting action, so I guess this is obsolete or did I understand u wrong.

adeebshihadeh:
I meant enabling [additional mypy flags](https://mypy.readthedocs.io/en/stable/command_line.html#untyped-definitions-and-calls) to enforce that we have 100% coverage.

4ydan:
I was thinking about that, but hard to make a decision on what you want/need.

adeebshihadeh:
I was thinking that we enable some and see what the changes would look like. Want to try some different combinations and post a quick report here on each one? Then we can see which are worth enabling and fixing all the violations for.

4ydan:
I looked at what configs other people(pytorch) use and added some of them

```
warn_unused_configs = True
show_error_codes = True
no_implicit_optional = True
allow_redefinition = True
show_column_numbers = True
check_untyped_defs = True
follow_imports = silent
```
Could also add  `plugins = numpy.typing.mypy_plugin` but I was not sure.

So I tried adding some type hints, but it wasn't very easy to guess stuff. I pushed to share my ideas for now.
Also added 2 folders and a script to generate reports and see the data.

4ydan:
image
Some of them seem unsolvable.

4ydan:
There is some inconsistencies with return types, especiall when having inheritance, it is not usual to have a base class return something else than the derived class or am I wrong about this.
For example BaseHandle.controlWrite returns `int` and PandaSpiHandle.controlWrite returns `bytes`. Increasing the strictness of mypy and following the errors is revealing some problems in the design of the class hierarchy. Also there have been other occurences where some functions return int and some bytes.

4ydan:
After adding these:
```sh
no_implicit_optional = True
check_untyped_defs = True
follow_imports = silent
show_column_numbers = True
show_error_codes = True
```
And fixing lots of type errors, this is the result.
image

These should be done next, maybe in a separate PR.
```sh
disallow_any_generics = True
disallow_subclassing_any = True
disallow_untyped_calls = True
disallow_untyped_defs = True
disallow_incomplete_defs = True
disallow_untyped_decorators = True
implicit_reexport = False
strict_equality = True
```
[mypy_report.zip](https://github.com/commaai/panda/files/11963374/mypy_report.zip) In case you wanna have a more detailed look.

python/__init__.py
```diff
@@ -329,11 +332,11 @@ def connect(self, claim=True, wait=False) -> None:
...
 
   @staticmethod
-  def spi_connect(serial) -> tuple[PandaSpiHandle, Optional[str], bool, None]:
+  def spi_connect(serial) -> tuple[Optional[PandaSpiHandle], Optional[str], bool, None]:
     # get UID to confirm slave is present and up
-    handle: PandaSpiHandle = None
-    spi_serial = None
-    bootstub: bool = False
+    handle: Optional[PandaSpiHandle]
+    spi_serial: str
```
4ydan, line 338: Blank space missing

python/__init__.py
```diff
@@ -283,21 +283,24 @@ def close(self) -> None:
...
   def connect(self, claim=True, wait=False) -> None:
-    serial: str
+    handle: Optional[Union[PandaSpiHandle, PandaUsbHandle]]
+    serial: Optional[str]
     bcd: bytearray | None
     self.close()
 
-    self._handle = None
-    while self._handle is None:
+    handle = None
```
4ydan, line 291: This part is really awkward.

python/__init__.py
```diff
@@ -841,7 +843,7 @@ def can_recv(self):
     dat = bytearray()
     while True:
       try:
-        dat = self._handle.bulkRead(1, 16384) # Max receive batch size + 2 extra reserve frames
+        dat = bytearray(self._handle.bulkRead(1, 16384)) # Max receive batch size + 2 extra reserve frames
```
4ydan, line 846: Unsure about this.

python/isotp.py
```diff
@@ -1,148 +1,152 @@
...
+    if DEBUG:
+        print("S:", binascii.hexlify(x))
+    if len(x) <= 7:
+        ret = bytes([len(x)]) + x
+    else:
+        assert False
+    return ret.ljust(8, b"\x00")
+
+
+kmsgs: List[Tuple[int, int, int, int]] = []
```
4ydan, line 18: Unsure about this.

4ydan:
image

4ydan:
Need to ignore now the tests folder with pylint I guess?

adeebshihadeh:
Yes, we only want these stricter options in the `python/` directory. Looks like we're already finding some inconsistencies. It's generally safe to assume these inconsistencies are unnecessary and should be fixed, rather than them being done this way intentionally :)

Found this blog post about mypy that seems to have some good tips: https://justincaustin.com/blog/mypy-tips-and-tricks/.

4ydan:
I think this issue/PR should be split up in small PR's increasing the mypy coverage per file(s).

adeebshihadeh:
Feel free to open more minimal PRs we can merge right away. Closing as this has gone stale.

PR #1497, opened 2023-07-06
SPI kernel driver

adeebshihadeh:
Extends the [existing spidev driver](https://www.kernel.org/doc/Documentation/spi/spidev) we use with a custom command for the panda.


adeebshihadeh:
Merging behind a flag if we want to revisit later.

PR #1499, opened 2023-07-11
Subaru: Global gen1 longitudinal safety

jnewb1:
- OpenPilot/DBC refs needed updating because of AEB_Status being added
- Cleaned up subaru tests to use a base test that is extended by stock and op long tests
- Made macros in safety_subaru for common messages between gen1/gen2
- Added a common test for Gas/Brake cars (cars that don't have direct acceleration requests)
- subaru_longitudinal is behind the allow_debug flag


jnewb1:
moving this to comma repo for tests

PR #1500, opened 2023-07-11
 Subaru: Global gen1 longitudinal safety

jnewb1:
- OpenPilot/DBC refs needed updating because of AEB_Status being added
- Cleaned up subaru tests to use a base test that is extended by stock and op long tests
- Made macros in safety_subaru for common messages between gen1/gen2
- Added a common test for Gas/Brake cars (cars that don't have direct acceleration requests)
- subaru_longitudinal is behind the allow_debug flag


board/safety/safety_subaru.h
```diff
@@ -20,6 +20,14 @@ const SteeringLimits SUBARU_GEN2_STEERING_LIMITS = {
   .type = TorqueDriverLimited,
 };
 
+const LongitudinalLimits SUBARU_LONG_LIMITS = {
+  .min_gas = 0,
+  .max_gas = 3400,
+  .min_rpm = 0,
+  .max_rpm = 3200,
```
sshane, line 27: is RPM safety related? can setting it high cause more acceleration (gas) than we intend? If not, then we don't need to enforce it

board/safety_declarations.h
```diff
@@ -220,6 +223,9 @@ uint32_t ts_angle_last = 0;
 int desired_angle_last = 0;
 struct sample_t angle_meas;         // last 6 steer angles/curvatures
 
+// for safety modes that need to passthru aeb
+bool stock_aeb = false;
+
```
sshane, line 228: I think we can do AEB safety and passthrough in a separate PR. We make no guarantees for any alpha longitudinal behind the debug flag

tests/libpanda/safety_helpers.h
```diff
@@ -168,6 +168,14 @@ void set_honda_fwd_brake(bool c){
...
 }
 
+void set_stock_aeb(bool c){
+  stock_aeb = c;
+}
+
+bool get_stock_aeb(){
+  return stock_aeb;
+}
+
```
sshane, line 178: We might be able to combine honda's fwd_brake with a generic variable

board/safety/safety_subaru.h
jnewb1: Stock seems to scale throttle and rpm together, so I think its reasonable to limit it to the bounds of the stock system

board/safety_declarations.h
jnewb1: I'll remove aeb stuff for now and move it to a new PR

tests/libpanda/safety_helpers.h
jnewb1: Will make new PR for AEB stuff

board/safety/safety_subaru.h
sshane: In that case, we can probably use one max value for both signals. Are they relatively close and move together?

jnewb1: rpm is typically a lot less, and will slowly ramp down even when throttle goes to zero

board/safety/safety_subaru.h
```diff
@@ -20,6 +20,14 @@ const SteeringLimits SUBARU_GEN2_STEERING_LIMITS = {
   .type = TorqueDriverLimited,
 };
 
+const LongitudinalLimits SUBARU_LONG_LIMITS = {
+  .min_gas = 0,
+  .max_gas = 3400,
```
sshane, line 25: Make sure to label how much acceleration in m/s^2 this applies

board/safety/safety_subaru.h
```diff
@@ -20,6 +20,14 @@ const SteeringLimits SUBARU_GEN2_STEERING_LIMITS = {
...
+  .min_rpm = 0,
+  .max_rpm = 3200,
+  .max_brake = 400,
```
sshane, line 28: Same, is this -4 m/s^2?

board/safety/safety_subaru.h
sshane: Do you have any stock routes showing how the EyeSight system uses it? I would understand how it uses it and how it affects acceleration. If setting both to max values results in the desired max accel, we can limit both to max_gas

board/safety/safety_subaru.h
jnewb1: actual brake pressure. I will do some plots to see how brake and throttle relates to acceleration

jnewb1: brake is pretty linear. 400 is about -2.5m/s^2 according to c3 accelerometer 
image

board/safety/safety_subaru.h
jnewb1: image

RPM and Throttle are roughly related. I think just scaling them together will be sufficient for the initial PR, and I'll limit RPM to the same as throttle to make it easier.

board/safety/safety_subaru.h
sshane: looks like it changes with speed, you can do something similar to GM like in https://github.com/commaai/openpilot/pull/25453 when you start to tune. Doesn't look like it can engine brake which is interesting. Maybe RPM can do that.

sshane: We usually allow 2 m/s in the safety, you can bump it

board/safety/safety_subaru.h
```diff
@@ -20,6 +20,14 @@ const SteeringLimits SUBARU_GEN2_STEERING_LIMITS = {
...
 };
 
+const LongitudinalLimits SUBARU_LONG_LIMITS = {
+  .max_brake = 400,  // approx -2.5 m/s^2
+
+  .min_gas = 0,
+  .max_gas = 3400,   // approx  1.2 m/s^2 at low speeds, less at higher speeds
+  .inactive_gas = 808,
+};
+
```
sshane, line 30: We usually allow 2 m/s^2 and -3.5 m/s^2 in safety

sshane, line 30:
```suggestion
const LongitudinalLimits SUBARU_LONG_LIMITS = {
  .max_gas = 3400,   // approx  1.2 m/s^2 at low speeds, less at higher speeds
  .inactive_gas = 808,
  .max_brake = 400,  // approx -2.5 m/s^2
};

```

corrected order. min_gas is only used if the range is not 0 to +x

jnewb1: I'll increase brake limit, but I'm going to leave gas limit as is until I can test max RPM and max gas in our subaru

tests/safety/common.py
```diff
@@ -175,6 +174,53 @@ def test_accel_actuation_limits(self, stock_longitudinal=False):
           self.assertEqual(should_tx, self._tx(self._accel_msg(accel)))
 
 
+class LongitudinalGasBrakeSafetyTest(PandaSafetyTestBase, abc.ABC):
```
sshane, line 177: Nice, I wanted to eventually do this, but I forgot if there was a snag making it generic. Feel free to do this in another PR before or after this one

tests/safety/common.py
```diff
@@ -175,6 +174,53 @@ def test_accel_actuation_limits(self, stock_longitudinal=False):
...
+  MIN_BRAKE: int = 0
+  MAX_BRAKE: Optional[int] = None
+
+  MIN_THROTTLE: int = 1818
```
sshane, line 181: let's use gas for now, it's what the other brands use

tests/safety/common.py
jnewb1: once i get this merged I'll look at cleaning up the safety tests for the other brands too

tests/safety/common.py
jnewb1: done

jnewb1:
Got a safety replay from martin, this should be good to merge now @sshane

sshane:
Great, I'll look over everything soon!

board/safety/safety_subaru.h
```diff
@@ -109,18 +130,18 @@ static int subaru_rx_hook(CANPacket_t *to_push) {
...
     // update vehicle moving with any non-zero wheel speed
-    if ((addr == MSG_SUBARU_Wheel_Speeds) && (bus == alt_bus)) {
+    if ((addr == MSG_SUBARU_Wheel_Speeds) && (bus == alt_main_bus)) {
       vehicle_moving = ((GET_BYTES(to_push, 0, 4) >> 12) != 0U) || (GET_BYTES(to_push, 4, 4) != 0U);
     }
 
-    if ((addr == MSG_SUBARU_Brake_Status) && (bus == alt_bus)) {
-      brake_pressed = ((GET_BYTE(to_push, 7) >> 6) & 1U);
+    if ((addr == MSG_SUBARU_Brake_Status) && (bus == alt_main_bus)) {
+      brake_pressed = GET_BIT(to_push, 62U) != 0U;
```
sshane, line 144: you can do the variable rename and these GET_BIT refactors in another PR if you want

board/safety/safety_subaru.h
```diff
@@ -161,7 +203,11 @@ static int subaru_fwd_hook(int bus_num, int addr) {
   int bus_fwd = -1;
 
   if (bus_num == SUBARU_MAIN_BUS) {
-    bus_fwd = SUBARU_CAM_BUS;  // forward to camera
+    // Global Platform
+    bool block_msg = subaru_longitudinal && ((addr == MSG_SUBARU_Brake_Status) || (addr == MSG_SUBARU_CruiseControl));
+    if (!block_msg) {
+      bus_fwd = SUBARU_CAM_BUS;  // to the eyesight camera
+    }
```
sshane, line 210: What does this do? Why block these messages to the camera? Comment explaining this would be a good idea

sshane (changes requested):

board/safety/safety_subaru.h
```diff
@@ -14,6 +14,12 @@ const SteeringLimits SUBARU_STEERING_LIMITS = SUBARU_STEERING_LIMITS_GENERATOR(2
 const SteeringLimits SUBARU_GEN2_STEERING_LIMITS = SUBARU_STEERING_LIMITS_GENERATOR(1000, 40, 40);
 
 
+const LongitudinalLimits SUBARU_LONG_LIMITS = {
+  .max_gas = 3400,      // approx  1.2 m/s^2 at low speeds, less at higher speeds
+  .inactive_gas = 1818, // this is zero acceleration
```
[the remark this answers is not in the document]
jnewb1: 1818 is zero acceleration, 3400 is full, and 808 appears to be engine braking (set to 808 when braking value is high)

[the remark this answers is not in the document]
sshane: then make sure to set min_gas and clip to 808 in openpilot. check GM, it's very similar (there's a zero accel and regen gas value)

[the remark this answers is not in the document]
sshane: since 3400 is only 1.2 m/s^2, does the stock system ever go above? maybe the RPM adds on to the acceleration somewhat

board/safety/safety_subaru.h
```diff
@@ -14,6 +14,12 @@ const SteeringLimits SUBARU_STEERING_LIMITS = SUBARU_STEERING_LIMITS_GENERATOR(2
 const SteeringLimits SUBARU_GEN2_STEERING_LIMITS = SUBARU_STEERING_LIMITS_GENERATOR(1000, 40, 40);
 
 
+const LongitudinalLimits SUBARU_LONG_LIMITS = {
+  .max_gas = 3400,      // approx  1.2 m/s^2 at low speeds, less at higher speeds
```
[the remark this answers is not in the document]
jnewb1: it goes above 4000 when on the highway and accelerating

board/safety/safety_subaru.h
jnewb1: Brake_Status and CruiseControl are intercepted and rewritten in openpilot.

Brake_Status is to fake the ES_Brake feedback signal to eyesight, which was for preserving AEB.
CruiseControl is to tell eyesight that it is not enabled (we set Cruise_Activated to false), so it only will try and perform AEB

board/safety/safety_subaru.h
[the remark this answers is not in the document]
jnewb1: I'll do some testing with RPM and Throttle on our outback. Stock goes up to 4000 when accelerating on the highway

board/safety/safety_subaru.h
```diff
@@ -132,13 +153,17 @@ static int subaru_rx_hook(CANPacket_t *to_push) {
   return valid;
 }
 
+
 static int subaru_tx_hook(CANPacket_t *to_send) {
 
   int tx = 1;
   int addr = GET_ADDR(to_send);
+  bool violation = false;
```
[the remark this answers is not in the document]
jnewb1: refactored to use "violation" everywhere

board/safety/safety_subaru.h
sshane: Why can't you let the EyeSight camera see the car's signals? Will it thinking it's in control of ACC somehow change the AEB behavior?

@adeebshihadeh told me that when working on the first iteration of an experimental longitudinal support PR, we don't need or want to take the time to verify that AEB still works, since we don't guarantee that, and when we do ship this to release (not master), we will need to re-do AEB testing anyway.

If there are no visible faults, I think we should remove this blocking/forwarding (complexity) and just show the AEB disabled light for now.

tests/safety/common.py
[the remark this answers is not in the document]
jnewb1: New PR for making this test. The only other brand I saw that was using gas/brake was GM, do you know of any others?

https://github.com/commaai/panda/pull/1537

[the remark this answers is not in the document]
sshane: Think honda might? Also Ford. Though it's a bit different, they implement a gas check and accel check (for braking only). Might want classes to individually test each, or a class for these two cases as it seems common.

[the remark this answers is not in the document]
sshane: Though since we don't have a nice way to know the max ranges of signals, you can keep the gas/brake tests locally to Subaru for now, and we can clean up in a future PR.

board/safety/safety_subaru.h
```diff
@@ -170,7 +216,13 @@ static int subaru_fwd_hook(int bus_num, int addr) {
                        (addr == MSG_SUBARU_ES_DashStatus) ||
                        (addr == MSG_SUBARU_ES_LKAS_State) ||
                        (addr == MSG_SUBARU_ES_Infotainment));
-    if (!block_lkas) {
+
+    bool block_long = ((addr == MSG_SUBARU_ES_Brake) ||
+                       (addr == MSG_SUBARU_ES_Distance) ||
+                       (addr == MSG_SUBARU_ES_Status));
```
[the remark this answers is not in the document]
jnewb1: c3d1ccb52f5f9d65|2023-08-01--15-02-58 -> sending max cruise_rpm but very low cruise_throttle (just above inactive), car very slowly accelerates, and the engine eventually gets to cruise_rpm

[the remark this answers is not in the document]
sshane: so does the acceleration correspond to cruise_throttle, or is it slightly higher? and does sending both high cause more acceleration than just cruise_throttle or what you were doing in openpilot (can the RPM allow more acceleration than what you've defined in safety)?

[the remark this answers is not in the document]
jnewb1: c3d1ccb52f5f9d65|2023-08-01--16-16-02 -> maxing cruise_throttle (3400) and cruise_rpm (3200) gives around 3 m/s^2 from a stop
c3d1ccb52f5f9d65|2023-08-01--16-26-58 -> maxing cruise_throttle (3400) and a low cruise_rpm (1200) gives more like 2 m/s^2 from stop

So it seems like we should also limit cruise_rpm as it does have some effect on acceleration

sshane:
It looks like ES_Status->Cruise_RPM is just a request of the transmission RPM, can we just set it to some median value at all times(since it seems related to the speed), and use the throttle, or does that reduce some performance?

image

board/safety/safety_subaru.h
jnewb1: Moved these small changes to new PR: https://github.com/commaai/panda/pull/1543

board/safety/safety_subaru.h
```diff
@@ -14,6 +14,13 @@ const SteeringLimits SUBARU_STEERING_LIMITS = SUBARU_STEERING_LIMITS_GENERATOR(2
 const SteeringLimits SUBARU_GEN2_STEERING_LIMITS = SUBARU_STEERING_LIMITS_GENERATOR(1000, 40, 40);
 
 
+const LongitudinalLimits SUBARU_LONG_LIMITS = {
+  .min_gas = 808,
+  .max_gas = 3400,      // approx  2 m/s^2 when maxing cruise_rpm and cruise_throttle
+  .inactive_gas = 1818, // this is zero acceleration, and 808 appears to be engine braking
```
adeebshihadeh, line 20:
```suggestion
  .inactive_gas = 1818, // this is zero acceleration
```

board/safety/safety_subaru.h
```diff
@@ -195,6 +243,10 @@ static int subaru_fwd_hook(int bus_num, int addr) {
 static const addr_checks* subaru_init(uint16_t param) {
   subaru_gen2 = GET_FLAG(param, SUBARU_PARAM_GEN2);
 
+#ifdef ALLOW_DEBUG
+  subaru_longitudinal = GET_FLAG(param, SUBARU_PARAM_LONGITUDINAL);
```
adeebshihadeh, line 247: also gate on not gen2?

board/safety/safety_subaru.h
```diff
@@ -165,7 +187,27 @@ static int subaru_tx_hook(CANPacket_t *to_send) {
     violation |= steer_torque_cmd_checks(desired_torque, -1, limits);
   }
 
-  if (violation){
+  if (subaru_longitudinal) {
```
adeebshihadeh, line 190: this shouldn't be needed since this only further blacklists messages. if we're not doing long, tx should already be false from the `msg_allowed` check

board/safety/safety_subaru.h
```diff
@@ -165,7 +187,27 @@ static int subaru_tx_hook(CANPacket_t *to_send) {
...
+
+    // check es_distance cruise_throttle limits
+    if (addr == MSG_SUBARU_ES_Distance) {
+      int cruise_throttle = ((GET_BYTES(to_send, 0, 4) >> 16) & 0xFFFU);
+      violation |= longitudinal_gas_checks(cruise_throttle, SUBARU_LONG_LIMITS);
+    }
+
+    // check es_status cruise_rpm limits
+    if (addr == MSG_SUBARU_ES_Status) {
+      int cruise_rpm = ((GET_BYTES(to_send, 0, 4) >> 16) & 0xFFFU);
```
adeebshihadeh, line 205: simplify this, it's getting two extra bytes then immediately shifting them out

board/safety/safety_subaru.h
```diff
@@ -165,7 +187,27 @@ static int subaru_tx_hook(CANPacket_t *to_send) {
     violation |= steer_torque_cmd_checks(desired_torque, -1, limits);
   }
 
-  if (violation){
+  if (subaru_longitudinal) {
+    // check es_brake brake_pressure limits
+    if (addr == MSG_SUBARU_ES_Brake) {
+      int es_brake_pressure = ((GET_BYTES(to_send, 0, 4) >> 16) & 0xFFFFU);
```
adeebshihadeh, line 193: same here

board/safety/safety_subaru.h
```diff
@@ -165,7 +187,27 @@ static int subaru_tx_hook(CANPacket_t *to_send) {
...
+      violation |= longitudinal_brake_checks(es_brake_pressure, SUBARU_LONG_LIMITS);
+    }
+
+    // check es_distance cruise_throttle limits
+    if (addr == MSG_SUBARU_ES_Distance) {
+      int cruise_throttle = ((GET_BYTES(to_send, 0, 4) >> 16) & 0xFFFU);
```
adeebshihadeh, line 199: and here

adeebshihadeh:
Also rebase on master to get the HITL test fix

board/safety/safety_subaru.h
jnewb1: done

board/safety/safety_subaru.h
jnewb1: done

board/safety/safety_subaru.h
jnewb1: done

board/safety/safety_subaru.h
jnewb1: done

board/safety/safety_subaru.h
jnewb1: done

board/safety/safety_subaru.h
jnewb1: moved it up by the min_gas field

tests/safety/test_subaru.py
```diff
@@ -101,6 +104,52 @@ class TestSubaruGen2SafetyBase(TestSubaruSafetyBase):
...
+
+  MIN_BRAKE = 0
+  MAX_BRAKE = 600
+  MAX_POSSIBLE_BRAKE = 2**16
+
+  FWD_BLACKLISTED_ADDRS = {2: [MSG_SUBARU_ES_LKAS, MSG_SUBARU_ES_Brake, MSG_SUBARU_ES_Distance,
+                               MSG_SUBARU_ES_Status, MSG_SUBARU_ES_DashStatus,
+                               MSG_SUBARU_ES_LKAS_State, MSG_SUBARU_ES_Infotainment]}
+
+  def _generic_limit_safety_check(self, msg_function, min_value, max_value, max_possible_value, inactive_value=0):
```
adeebshihadeh, line 122: doesn't seem subaru specific. @sshane do we have something like this somewhere else?

jnewb1: I started one here: https://github.com/commaai/panda/pull/1537

RPM would be specific to subaru, and some cars use an acceleration request for brake, and throttle-like message for gas. But I can look at cleaning it up into a common test

adeebshihadeh: yeah let's do that first

board/safety/safety_subaru.h
jnewb1: I think we do want this check actually, because we use ES_Distance for a cancel request even when stock long is enabled, and it might get blocked here if eyesight happens to be doing the max accel (I've seen it go to 4000)

adeebshihadeh: that shouldn't be allowed in stock long. can you make a pair of PRs to openpilot and panda to fix that? no actuation should be allowed while cancelling

tests/safety/test_subaru.py
jnewb1: done https://github.com/commaai/panda/pull/1537

board/safety/safety_subaru.h
jnewb1: Done

tests/safety/test_subaru.py
```diff
@@ -100,6 +103,40 @@ def test_cancel_message(self):
...
+  INACTIVE_GAS = 1818
+  MAX_POSSIBLE_GAS = 2**12
+
+  MIN_BRAKE = 0
+  MAX_BRAKE = 600
+  MAX_POSSIBLE_BRAKE = 2**16
+
+  MIN_RPM = 0
+  MAX_RPM = 2400
+  MAX_POSSIBLE_RPM = 2**12
```
adeebshihadeh, line 118: ideally we get this from the DBC

board/safety/safety_subaru.h
```diff
@@ -166,12 +191,32 @@ static int subaru_tx_hook(CANPacket_t *to_send) {
...
+  if (addr == MSG_SUBARU_ES_Distance) {
     int cruise_throttle = (GET_BYTES(to_send, 2, 2) & 0xFFFU);
     bool cruise_cancel = GET_BIT(to_send, 56U) != 0U;
-    violation |= (cruise_throttle != 1818);
-    violation |= (!cruise_cancel);
+    
+    if (subaru_longitudinal) {
+      violation |= longitudinal_gas_checks(cruise_throttle, SUBARU_LONG_LIMITS);
+    }
+    else {
```
adeebshihadeh, line 208: same line

board/safety/safety_subaru.h
```diff
@@ -14,6 +14,16 @@ const SteeringLimits SUBARU_STEERING_LIMITS = SUBARU_STEERING_LIMITS_GENERATOR(2
 const SteeringLimits SUBARU_GEN2_STEERING_LIMITS = SUBARU_STEERING_LIMITS_GENERATOR(1000, 40, 40);
 
 
+const LongitudinalLimits SUBARU_LONG_LIMITS = {
+  .min_gas = 808,       // appears to be engine braking
+  .max_gas = 3400,      // approx  2 m/s^2 when maxing cruise_rpm and cruise_throttle
+  .inactive_gas = 1818, // this is zero acceleration
+  .max_brake = 600,     // approx -3.5 m/s^2
+
+  .min_alt = 0,         // Transmission RPM limits
```
adeebshihadeh, line 23: this alt name is confusing

jnewb1: the goal was to make it generic in case any other brands have a similar "alternative" gas or brake signal that needs limiting. "rpm" is also weird, because it's actually the transmission rpm not the engine rpm

adeebshihadeh: it's **incredibly important** that this particular code is as simple and readable as possible. in this case, the variable should probably be called `min_transmission_rpm`

board/safety/safety_subaru.h
jnewb1: done

board/safety/safety_subaru.h
jnewb1: renamed to transmission_rpm

tests/safety/test_subaru.py
sshane: yeah we should add to CANDefine (right now it just gives you the value map, but we can expand to give general signal info)

PR #1502, opened 2023-07-12
H7: CAN core reset on no ACK and high transmit error counter

briskspirit:


adeebshihadeh:
No issues so far driving around in the EV6, but do we have a reliable way to reproduce this? If not, we should be able to repro by running fingerprinting in a loop and confirm it never fails now.

sshane:
Out of 100 fingerprint attempts with this branch, none failed (had dropped frames), while only 14 of 100 passed on panda master.

adeebshihadeh:
How exactly does this fix the fingerprinting issue? Just recover faster when multiplexing quickly?

briskspirit:
Recover faster and not to overload H7 with interrupts while trying to send frame on non-responsive bus (as we have auto retry on failed TX enabled)

PR #1505, opened 2023-07-12
Subaru: Use macros for gen1/gen2

jnewb1:
- Pulled the macro stuff from my long PR into this one, so I can use it for the forester PR too

### DIFF
--- a/tests/safety/test_subaru.py
+++ b/tests/safety/test_subaru.py
@@ -6,39 +6,13 @@ import panda.tests.safety.common as common
 from panda.tests.safety.common import CANPackerPanda
 
 
-MSG_SUBARU_Brake_Status     = 0x13c
-MSG_SUBARU_CruiseControl    = 0x240
-MSG_SUBARU_Throttle         = 0x40
-MSG_SUBARU_Steering_Torque  = 0x119
-MSG_SUBARU_Wheel_Speeds     = 0x13a
-MSG_SUBARU_ES_LKAS          = 0x122
-MSG_SUBARU_ES_Brake         = 0x220
-MSG_SUBARU_ES_Distance      = 0x221
-MSG_SUBARU_ES_Status        = 0x222
-MSG_SUBARU_ES_DashStatus    = 0x321
-MSG_SUBARU_ES_LKAS_State    = 0x322
-MSG_SUBARU_ES_Infotainment  = 0x323
-
-SUBARU_MAIN_BUS = 0
-SUBARU_ALT_BUS  = 1
-SUBARU_CAM_BUS  = 2
-
-
-def lkas_tx_msgs(alt_bus):
-  return [[MSG_SUBARU_ES_LKAS,          SUBARU_MAIN_BUS], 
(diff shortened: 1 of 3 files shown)


jnewb1:
@sshane would like to use this to reduce copy/pasting in the forester PR

tests/safety/test_subaru.py
```diff
@@ -6,13 +6,39 @@
...
+MSG_SUBARU_Throttle         = 0x40
+MSG_SUBARU_Steering_Torque  = 0x119
+MSG_SUBARU_Wheel_Speeds     = 0x13a
+MSG_SUBARU_ES_LKAS          = 0x122
+MSG_SUBARU_ES_Brake         = 0x220
+MSG_SUBARU_ES_Distance      = 0x221
+MSG_SUBARU_ES_Status        = 0x222
+MSG_SUBARU_ES_DashStatus    = 0x321
+MSG_SUBARU_ES_LKAS_State    = 0x322
+MSG_SUBARU_ES_Infotainment  = 0x323
```
sshane, line 20: Now that I'm seeing this in Python, maybe it makes sense to use an Enum for this, but the `.value` will be cumbersome, so maybe just a class? And a struct/enum in C. What do you think?

tests/safety/test_subaru.py
```diff
@@ -24,12 +50,23 @@ class TestSubaruSafety(common.PandaSafetyTest, common.DriverTorqueSteeringSafety
...
+  MAX_RATE_UP = 50
+  MAX_RATE_DOWN = 70
+  MAX_TORQUE = 2047
+
+  @classmethod
+  def setUpClass(cls):
+    if cls.__name__.endswith("Base"):
+      cls.packer = None
+      cls.safety = None
+      raise unittest.SkipTest
```
sshane, line 64: PandaSafetyTest already does this for you, but hmm `TorqueSteeringSafetyTestBase` overrides it. Want to clean all that interaction up in another PR?

https://github.com/commaai/panda/blob/6b9b448259d1790ed8e18abde09ee72446996b3a/tests/safety/common.py#L689-L693

tests/safety/test_subaru.py
```diff
@@ -24,12 +50,23 @@ class TestSubaruSafety(common.PandaSafetyTest, common.DriverTorqueSteeringSafety
   DRIVER_TORQUE_ALLOWANCE = 60
   DRIVER_TORQUE_FACTOR = 50
 
-  ALT_BUS = 0
+  ALT_BUS = SUBARU_MAIN_BUS
 
+  MAX_RATE_UP = 50
+  MAX_RATE_DOWN = 70
+  MAX_TORQUE = 2047
```
sshane, line 57: redefinition

sshane:
the safety_subaru.h looks like a strict clean up, can merge that part now if you want

jnewb1:
> the safety_subaru.h looks like a strict clean up, can merge that part separately now if you want
will merge this part and make a new PR for the tests

PR #1508, opened 2023-07-14
Subaru: cleanup tests

jnewb1:


jnewb1:
On using enums: this is how ford port does it. Enum would be fine but we probably want to be consistent. What do you prefer?

https://github.com/commaai/panda/blob/6d2b70a1475922983f59c57e68e68a6f61099268/tests/safety/test_ford.py#L11-L21

tests/safety/test_hyundai_canfd.py
```diff
@@ -95,7 +96,8 @@ class TestHyundaiCanfdHDA1Base(TestHyundaiCanfdBase):
 
   @classmethod
   def setUpClass(cls):
-    if cls.__name__ in ("TestHyundaiCanfdHDA1", "TestHyundaiCanfdHDA1AltButtons") or cls.__name__.endswith('Base'):
+    super().setUpClass()
```
jnewb1, line 99: Called base setup functions here, which removes need for the "Base" check

jnewb1:
> On using enums: this is how ford port does it. Enum would be fine but we probably want to be consistent. What do you prefer?
> 
> https://github.com/commaai/panda/blob/6d2b70a1475922983f59c57e68e68a6f61099268/tests/safety/test_ford.py#L11-L21

Maybe it would be good to refactor one port fully to exactly how we want future ports to be, and then try and use that as a standard?

tests/safety/common.py
```diff
@@ -190,6 +190,7 @@ class TorqueSteeringSafetyTestBase(PandaSafetyTestBase, abc.ABC):
 
   @classmethod
   def setUpClass(cls):
+    super().setUpClass()
```
sshane, line 193: don't think this does anything. the endswith check is in `PandaSafetyTest`, not the base class. forgot why I didn't just put endswith base in the `PandaSafetyTestBase` class, can you see if there's any issues? can be separate from this subaru-specific PR

tests/safety/common.py
```diff
@@ -335,6 +335,7 @@ class DriverTorqueSteeringSafetyTest(TorqueSteeringSafetyTestBase, abc.ABC):
...
```
sshane, line 338: this doesn't seem to do anything at the moment

PR #1512, opened 2023-07-15
H7: increase CAN IRQ rate limit from 16000 to 21000

briskspirit:
Add extra margin as the same IRQ is used for RX and errors.

### DIFF
--- a/board/stm32h7/stm32h7_config.h
+++ b/board/stm32h7/stm32h7_config.h
@@ -25,8 +25,10 @@ separate IRQs for RX and TX.
 - all currently supported cars are 0.5 Mbps / 2 Mbps
 
 1 / ((29 bits / 0.5Mbps) + (23 bits / 2Mbps)) = 14388Hz
+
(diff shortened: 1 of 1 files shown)


board/stm32h7/stm32h7_config.h
```diff
@@ -25,8 +25,10 @@ separate IRQs for RX and TX.
 - all currently supported cars are 0.5 Mbps / 2 Mbps
 
 1 / ((29 bits / 0.5Mbps) + (23 bits / 2Mbps)) = 14388Hz
+
+Add extra margin as the same IRQ is used for RX and errors.
 */
-#define CAN_INTERRUPT_RATE 16000U
+#define CAN_INTERRUPT_RATE 21000U
```
adeebshihadeh, line 31: do we need to increase this much if we disable the auto TX retry until we see activity on the bus? this was pretty helpful for finding that bug

briskspirit: hm, then no reason to increase it

PR #1514, opened 2023-07-15
CAN: turn off auto retransmission while no bus activity seen

briskspirit:
Hacky implementation, proof of concept, WIP

### DIFF
--- a/board/drivers/bxcan.h
+++ b/board/drivers/bxcan.h
@@ -183,7 +183,7 @@ void can_rx(uint8_t can_number) {
 
   // Enable automatic retransmission of messages after some activity on the bus
   if ((can_health[can_number].total_rx_cnt != 0U) && ((CAN->MCR & CAN_MCR_NART) != 0U)) {
-    register_clear_bits(&(CAN_obj->MCR), CAN_MCR_NART);
+    register_clear_bits(&(CAN->MCR), CAN_MCR_NART);
   }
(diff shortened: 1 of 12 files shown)


board/drivers/bxcan.h
```diff
@@ -181,6 +182,11 @@ void can_rx(uint8_t can_number) {
   CAN_TypeDef *CAN = CANIF_FROM_CAN_NUM(can_number);
   uint8_t bus_number = BUS_NUM_FROM_CAN_NUM(can_number);
 
+  // Enable automatic retransmission of messages after some activity on the bus
+  if (can_health[can_number].total_rx_cnt != 0U && (CAN->MCR & CAN_MCR_NART) != 0U) {
+    CAN->MCR &= ~(CAN_MCR_NART);
```
adeebshihadeh, line 187: let's just do this for the H7 for now

briskspirit: As discussed - auto retry was always enabled on F4

gregjhogan:
I feel like this PR should document why we are doing this.   It seems counterintuitive we would need this complexity, and possibly something that someone would try to remove in the future to simplify things.

I am also hoping it doesn't cause any weird problems on a bus that is normally silent (e.g. diagnostic port).  Dropping messages in general is always annoying when it comes to diagnostic traffic.

PR #1516, opened 2023-07-15
Ford: namespaced constants

incognitojam:
not sure if we want to add prefixes to the CAN MSG constants yet?


adeebshihadeh:
Yes, everything should be namespaced. Perhaps replace MSG_ with FORD_

PR #1267, opened 2023-03-02
HKG: Car Port for Hyundai Palisade 2023 and Kia Telluride 2023 (non-HDA2)

sunnyhaibin:
**Requires**
- https://github.com/commaai/panda/pull/1638
- https://github.com/commaai/panda/pull/1634

**Prerequisite of**
- https://github.com/commaai/opendbc/pull/1273

PR #1243, opened 2023-02-19
HKG: Car Port for Hyundai Palisade 2023 and Kia Telluride 2023 (HDA2)

sunnyhaibin:
**Checklist**
- [x] Unit Test

**Requires**
- https://github.com/commaai/panda/pull/1638
- https://github.com/commaai/panda/pull/1634

**Prerequisite of**
- https://github.com/commaai/opendbc/pull/1268

PR #1518, opened 2023-07-17
Hyundai Safety: Common bus assignment

sunnyhaibin:
Use common bus assignment method for CAN and CAN-FD safety.

**Related**
- https://github.com/commaai/panda/pull/1243
- https://github.com/commaai/panda/pull/1267


sshane:

PR #1520, opened 2023-07-18
CI: Bump refs

jnewb1:


jnewb1:
runtime error from https://github.com/commaai/opendbc/pull/891/files appears to be causing issues here

sshane:
That will be reverted soon

jnewb1:
> [commaai/opendbc#891](https://github.com/commaai/opendbc/pull/891) will be reverted soon

okay, need updated panda CI to move this forester PR forward so are you okay with me bumping the refs here?

PR #1524, opened 2023-07-18
IsoTpMessage: support ISO-TP w/ CAN FD

sshane:
very much a draft just to make it work, wanted to have a CAN FD user test.


python/uds.py
```diff
@@ -473,16 +473,23 @@ def _isotp_rx_next(self, rx_data: bytes) -> ISOTP_FRAME_TYPE:
...
 
     if rx_data[0] >> 4 == ISOTP_FRAME_TYPE.SINGLE:
-      self.rx_len = rx_data[0] & 0xFF
-      assert self.rx_len < self.max_len, f"isotp - rx: invalid single frame length: {self.rx_len}"
-      self.rx_dat = rx_data[1:1 + self.rx_len]
+      self.rx_len = rx_data[0] & 0x0F
+      # "if the first byte of SF=0, then second byte specifies the size of the data."
+      # - https://en.wikipedia.org/wiki/CAN_FD
+      offset = 1
+      if self.rx_len == 0x0 and len(rx_data) > 8:
```
sshane, line 480:
```suggestion
      if self.rx_len == 0x0 and len(rx_data) >= rx_data[1]:
```

sheaduncan:
Can we get this PR merged? I've tested it on my own branch, and it picked up the additional signals on bus 4. From there, it was able to get a Fuzzy Fingerprint.

coffee-cake-isaac:
Working on my F150 Lightning fork as well.

python/uds.py
coffee-cake-isaac: Worth noting we did have some issues with this. @sheaduncan can probably explain more.

sshane: Which version?

coffee-cake-isaac: Instead of using a hardcoded 8, using the `rx_data[1]` caused it to fail to fingerprint properly for him. I dont' have much more info other than that. I've used the `> 8` code since the beginning and that's seemingly working with no issues.

rwxr-xr-x:
I have been running with this change in this PR, as is, for a few weeks - it fingerprints correctly every time - 2023 F-150 ICE

sheaduncan:
@adeebshihadeh we have a lot of users on the Ford CAN-FD test branch that @coffee-cake-isaac created and continue to see successful fingerprints using this fix.

sheaduncan:
@sshane, this is working for C3X users. However, we do have one Mach-E user who missed a FP and was returning Mazda ECUs.

This UDS fix works to get all the responses, but there may need to be some work on the FW version query on the OP side.

image
These VEHICLE_MANUFACTURER_ECU_SOFTWARE_NUMBER seemed to be shared between Ford and Mazda

coffee-cake-isaac:
Also seem to have a bit of an issue with some HKG users no longer fingerprinting properly, either.

sshane:
What is the issue related to HKG?

coffee-cake-isaac:
> @coffee-cake-isaac what is the issue related to HKG?

I don't actually know if this was related or not; Sunny's users were having issues when introducing this with Sunnypilot, but I think it was unrelated as we haven't heard anything since.

coffee-cake-isaac:
> > @sshane, this is working for C3X users. However, we do have one Mach-E user who missed a FP and was returning Mazda ECUs.
> > This UDS fix works to get all the responses, but there may need to be some work on the FW version query on the OP side.
> > image
> > These VEHICLE_MANUFACTURER_ECU_SOFTWARE_NUMBER seemed to be shared between Ford and Mazda
> 
> Do you have a dongle ID for that user?

Actually, since I've had some ECU upgrades done for a recall mine is doing it now, as well. Since most of my routes recently have been on a fork without this panda change, I have a route that shows this as well.

`112e4d6e0cad05e1|2024-01-03--12-11-40`

sheaduncan:
> > @sshane, this is working for C3X users. However, we do have one Mach-E user who missed a FP and was returning Mazda ECUs.
> > This UDS fix works to get all the responses, but there may need to be some work on the FW version query on the OP side.
> > image
> > These VEHICLE_MANUFACTURER_ECU_SOFTWARE_NUMBER seemed to be shared between Ford and Mazda
> 
> Do you have a dongle ID for that user?

83a4e056c7072678
IMG_0277

sshane:
> 112e4d6e0cad05e1|2024-01-03--12-11-40

That route looks fine as long as we get everything back from the Ford queries, if ECUs happen to respond to Mazda queries, that's okay

sheaduncan:
> > 112e4d6e0cad05e1|2024-01-03--12-11-40
> 
> That route looks fine as long as we get everything back from the Ford queries, if ECUs happen to respond to Mazda queries, that's okay
> 
> Edit: same with 83a4e056c7072678, EPS responds to both ford and mazda so during fingerprinting it will only look at Mazda

Strange, not sure why they're failing to fingerprint then. Those FW versions are in Ford's listed FW versions.

coffee-cake-isaac:
Yeah, they all come back but it says "Car Unrecognized".

sshane:
Which route? All recent mock routes (not many) for 83a4e056c7072678 fingerprint on master.

python/uds.py
```diff
@@ -473,15 +473,22 @@ def _isotp_rx_next(self, rx_data: bytes) -> ISOTP_FRAME_TYPE:
...
+      # assert self.rx_len < self.max_len, f"isotp - rx: invalid single frame length: {self.rx_len}"
+      self.rx_dat = rx_data[offset:offset + self.rx_len]
       self.rx_idx = 0
       self.rx_done = True
       if self.debug:
         print(f"ISO-TP: RX - single frame - {hex(self._can_client.rx_addr)} idx={self.rx_idx} done={self.rx_done}")
       return ISOTP_FRAME_TYPE.SINGLE
 
     elif rx_data[0] >> 4 == ISOTP_FRAME_TYPE.FIRST:
+      # should work with CAN FD. maybe a check for max 62 bytes from the spec?
```
sshane, line 491: @sheaduncan can you get a route where you cause a query longer than the single frame can hold, so we can see the first and consecutive frame action?

sheaduncan: This is with a fresh copy of master and pointing to this Panda Branch:
09136c309ba9461d/2024-02-07--08-51-50


This is with a fresh copy of master with current Panda master:
09136c309ba9461d/2024-02-07--14-42-56

sheaduncan: 2 more routes:

Red Panda in Aux port of C3 and USB Power into Main
09136c309ba9461d/2024-02-07--17-31-57

Red Panda in Aux port and OBD2 Power in Main
09136c309ba9461d/2024-02-07--17-36-38

sheaduncan:
What's needed to help get this PR merged?

python/uds.py
sshane: @sheaduncan can you try this branch and upload the rlog? It won't fingerprint. https://github.com/commaai/openpilot/tree/ford-test

sheaduncan: @sshane None of these fingerprinted for that branch. Getting Hyundai responses and getting new FW extended versions.

09136c309ba9461d/00000000--7340e1444b
09136c309ba9461d/00000001--578b507ed2
09136c309ba9461d/00000002--b138639015
09136c309ba9461d/00000003--7e95bc66d8

sshane: Thanks, that's what I was looking for. I'll check it out when I have time.

sshane:
When we officially support CAN FD Ford, we will put the time into verifying and merging this!

blue-genie:
Just bringing this issue up that it will need to be merged when Ford CANFDs are upstreamed - other wise the fingerprinting won't work. I do not want this work to get lost.

uds.py was moved to opendbc.

PR #1525, opened 2023-07-19
Add logging to command safety checks

rmcc22:


adeebshihadeh:
These logs write to flash and are not for casual debugging like this.

rmcc22:
@adeebshihadeh  Oh yea I see that now. I should have looked at that first. Would be nice to have something like this that is buffered in ram and sent over SPI to show in the logs. Do we have such a thing?

PR #1526, opened 2023-07-19
print lat/long command safety violations

rmcc22:


adeebshihadeh:
This should be debugged offline. We have good tests and tools for this.

rmcc22:
> This should be debugged offline. We have good tests and tools for this.

The tests I have seen, namely panda/tests/safety/* , panda/tests/safety_replay/replay_drive.py, seldrive/car/tests/test_lateral_limits.py,  have all passed and yet it does not work onroad. In their current form these tests only verify that the safety limits are in place, they do not help debug the on-road violation cause. 

With the code I have added, when running panda/tests/safety_replay/replay_drive.py, it tells me what the violation was.

But maybe I missed something. is there another test I should be using that will pin point the issue?

PR #1528, opened 2023-07-20
Subaru: Add measured steering angle and vehicle speed

jnewb1:
Forester uses an angle lkas command. Need measured steering angle and vehicle speed.

https://github.com/commaai/panda/pull/1187


board/safety/safety_subaru.h
```diff
@@ -68,6 +68,8 @@ AddrCheckStruct subaru_gen2_addr_checks[] = {
 #define SUBARU_GEN2_ADDR_CHECK_LEN (sizeof(subaru_gen2_addr_checks) / sizeof(subaru_gen2_addr_checks[0]))
 addr_checks subaru_gen2_rx_checks = {subaru_gen2_addr_checks, SUBARU_GEN2_ADDR_CHECK_LEN};
 
+#define STEER_ANGLE_MEAS_FACTOR -0.0217
```
sshane, line 71: can you put in `SteeringLimits.angle_deg_to_can`?

board/safety/safety_subaru.h
```diff
@@ -68,6 +68,8 @@ AddrCheckStruct subaru_gen2_addr_checks[] = {
...
+#define WHEEL_SPEED_FACTOR 0.057
```
sshane, line 72: we can move this inline

board/safety/safety_subaru.h
```diff
@@ -106,6 +108,10 @@ static int subaru_rx_hook(CANPacket_t *to_push) {
       torque_driver_new = ((GET_BYTES(to_push, 0, 4) >> 16) & 0x7FFU);
       torque_driver_new = -1 * to_signed(torque_driver_new, 11);
       update_sample(&torque_driver, torque_driver_new);
+
+      int angle_meas_new = (GET_BYTES(to_push, 4, 2) & 0xFFFFU);
+      angle_meas_new = ROUND(to_signed(angle_meas_new, 16) * STEER_ANGLE_MEAS_FACTOR);
+      update_sample(&angle_meas, angle_meas_new);
```
sshane, line 114: think it might be a good idea to think about adding some kind of test that ensures safety modes parse these values correctly (sending messages from python, testing the sample or global variable)

sshane: are you sure it starts at byte 5?

sshane: ~are you sure it starts at byte 4?~ my bad, was looking at the command message

jnewb1: I think vehicle speed is tested for angle based cars (https://github.com/commaai/panda/blob/933a19af848104d2a6e5b50e3aca441960fe7a80/tests/safety/common.py#L654), doesn't seem like angle is tested though? I can add that

jnewb1: added measured angle test

board/safety/safety_subaru.h
jnewb1: moved inline

board/safety/safety_subaru.h
jnewb1: used SUBARU_ prefix. This PR doesn't have SteeringLimits, that will come with the forester PR.

sshane:
I noticed we have two more duplicate tests like this, can we not make them use this?

PR #1537, opened 2023-08-01
Safety Tests: Create common gas brake safety test

jnewb1:
Similar to LongitudinalAccelSafetyTest, but for cars that use a gas/brake interface rather than an acceleration one, like GM and Subaru.


tests/safety/common.py
```diff
@@ -175,6 +175,54 @@ def test_accel_actuation_limits(self, stock_longitudinal=False):
...
+          self.assertFalse(self._tx(msg_function(v)), (v, min_value, max_value))
+        else:
+          self.assertTrue(self._tx(msg_function(v)), (v, min_value, max_value))
+
+  def test_brake_safety_check(self):
+    self._generic_limit_safety_check(self._brake_msg, self.MIN_BRAKE, self.MAX_BRAKE, 1.5)
+  
+  def test_gas_safety_check(self):
+    self._generic_limit_safety_check(self._gas_msg, self.MIN_GAS, self.MAX_GAS, 1.2, self.INACTIVE_GAS)
+
```
sshane, line 224: What are these factors for?

tests/safety/test_gm.py
```diff
@@ -138,24 +133,6 @@ def _button_msg(self, buttons):
...
-
-  def test_gas_safety_check(self):
-    # Block if enabled and out of actuation range, disabled and not inactive regen, or if stock longitudinal
-    for enabled in [0, 1]:
-      for gas_regen in range(0, 2 ** 12 - 1):
-        self.safety.set_controls_allowed(enabled)
-        should_tx = ((enabled and self.MAX_REGEN <= gas_regen <= self.MAX_GAS) or
-                     gas_regen == self.INACTIVE_REGEN)
-        self.assertEqual(should_tx, self._tx(self._send_gas_msg(gas_regen)), (enabled, gas_regen))
-
```
sshane, line 158: there's more you can delete in GM. I would refactor GM in a separate PR

jnewb1: will move that to another pr

tests/safety/common.py
jnewb1: It was to give a range to test with (min_gas -> max_gas * 1.5), etc. But the acceleration test only does min_gas-1 -> max_gas+1, which is easier

sshane: I don't like that we would lose coverage of testing the entire signal range, I think that's why we didn't make this common yet, we'd need the DBC parser to get the max signal value. You can wait on this until after the Subaru PR, that can be specific to Subaru to have more coverage

tests/safety/common.py
```diff
@@ -175,6 +175,54 @@ def test_accel_actuation_limits(self, stock_longitudinal=False):
...
+    self.assertGreater(self.MAX_GAS, self.MIN_GAS)
+  
+  def _generic_limit_safety_check(self, msg_function, min_allowed_value: int, max_allowed_value: int, inactive_value=0):
+    for enabled in [0, 1]:
+      for v in np.concatenate((np.arange(min_allowed_value - 1, max_allowed_value + 1, 1), [0, inactive_value])):
+        self.safety.set_controls_allowed(enabled)
+        if (not enabled and v != inactive_value) or v > max_allowed_value or v < min_allowed_value:
+          self.assertFalse(self._tx(msg_function(v)), (v, min_allowed_value, max_allowed_value))
+        else:
+          self.assertTrue(self._tx(msg_function(v)), (v, min_allowed_value, max_allowed_value))
```
sshane, line 217: Think it makes sense to make this a helper in the base safety test, that way it can be even more generic. Any issues with that?

tests/safety/common.py
jnewb1: for now we can just hardcode the max value, since that's what we did before anyway (at least for GM)

tests/safety/common.py
jnewb1: moved it into base class and switched over GM

jnewb1:
@sshane  mind reviewing this again? subaru long is currently blocked on this. GM works, I'll do the rest of the brands that use the gas/brake after this is merged

tests/safety/common.py
```diff
@@ -69,6 +71,28 @@ def _rx(self, msg):
...
+
+    for controls_allowed in [False, True]:
+      # enforce we don't skip over 0 or inactive
+      for v in np.concatenate((np.arange(min_possible_value, max_possible_value, test_delta), np.array([0, inactive_value]))):
+        v = round(v, 2)  # floats might not hit exact boundary conditions without rounding
+        self.safety.set_controls_allowed(controls_allowed)
+        should_tx = controls_allowed and min_allowed_value <= v <= max_allowed_value
+        should_tx = (should_tx or v == inactive_value) and msg_allowed
+        self.assertEqual(self._tx(msg_function(v)), should_tx, (controls_allowed, should_tx, v))
+
```
sshane, line 95: I would make this into a new PR, trying to make it as generic as possible and replace duplicate code

jnewb1: https://github.com/commaai/panda/pull/1575

jnewb1:
I'll merge the test only (not GM) so subaru long isn't blocked by it. We can switch over GM and other brands in another PR

tests/safety/common.py
```diff
@@ -204,6 +204,39 @@ def test_accel_actuation_limits(self, stock_longitudinal=False):
...
+  def test_gas_brake_limits_correct(self):
+    self.assertIsNotNone(self.MAX_POSSIBLE_BRAKE)
+    self.assertIsNotNone(self.MAX_POSSIBLE_GAS)
+
+    self.assertGreater(self.MAX_BRAKE, self.MIN_BRAKE)
+    self.assertGreater(self.MAX_GAS, self.MIN_GAS)
+
+  @abc.abstractmethod
+  def _send_gas_msg(self, gas: int):
```
sshane, line 226: _generic_limit_safety_check takes a callable with a float, not an int

PR #1542, opened 2023-08-02
H7: fix GMLAN bitbang timer

briskspirit:


tests/hitl/3_usb_to_can.py
```diff
@@ -119,6 +119,14 @@ def test_gmlan_bad_toggle(p):
     assert comp_kbps_normal > (0.6 * SPEED_NORMAL)
     assert comp_kbps_normal < (1.0 * SPEED_NORMAL)
 
+@pytest.mark.panda_expect_can_error
+@pytest.mark.skip_panda_types(PandaGroup.GMLAN)
+def test_gmlan_bitbang(p):
```
adeebshihadeh, line 124: does this test catch the bug? it was pretty subtle

tests/hitl/3_usb_to_can.py
```diff
@@ -119,6 +119,14 @@ def test_gmlan_bad_toggle(p):
...
+  p.set_safety_mode(Panda.SAFETY_ALLOUTPUT)
+  for _ in range(10):
+    p.can_send(0x10, b"data", 3)
+    time.sleep(0.1)
+  assert p.health()['gmlan_send_errs'] == 0
```
adeebshihadeh, line 129: also check the messages were sent on the loopback bus? do they get added there?

tests/hitl/3_usb_to_can.py
adeebshihadeh, line 124: ideally, this test runs together with a white panda so we can read the message back

briskspirit: bug was caused by wrong timer settings, so yes, was caught on RP everytime

tests/hitl/3_usb_to_can.py
briskspirit: don't think we ever forwarded those back, only rejected messages by safety mode.

tests/hitl/3_usb_to_can.py
briskspirit: discussed

adeebshihadeh:
Can we finish this?

PR #1555, opened 2023-08-05
jungle: disable transceivers while doing CAN multiplexing

briskspirit:
Imo should be also done to all pandas, multiplexing disturbs bus.


briskspirit:
Actually we should keep not connected bus transceiver off

PR #1561, opened 2023-08-06
STM32H7 DFU bug workaround

robbederks:
The issue with the H7 not coming up when DFU-ing straight after a power cycle comes down to a bug in their dfu routine, which leaves PWR->CR3 unwritable until the next power-on reset. This means that the bootstub / application code gets stuck waiting in the while loop for the LDO to turn off and the switch from VOC3 to VOC1. While technically outside of the operating range of some of the PLLs, everything seems to work fine if we just skip the check and stay at VOC3.

### DIFF
--- a/board/stm32fx/clock.h
+++ b/board/stm32fx/clock.h
@@ -1,4 +1,8 @@
-void clock_init(void) {
+bool failed_clock_init = false;
+
+void clock_init(bool allow_vos3) {
+  UNUSED(allow_vos3);
+
   // enable external oscillator
   register_set_bits(&(RCC->CR), RCC_CR_HSEON);
   while ((RCC->CR & RCC_CR_HSERDY) == 0);
--- a/board/stm32h7/clock.h
+++ b/board/stm32h7/clock.h
@@ -17,13 +17,35 @@ APB4 per: 60MHz
 PCLK1: 60MHz (for USART2,3,4,5,7,8)
 */
 
-void clock_init(void) {
+bool failed_clock_init = false;
+
+void clock_init(bool allow_vos3) {
   //Set power mode to direct SMPS power supply(depends on the board layout)
   register_set(&(PWR->CR3), PWR_CR3_SMPSEN, 0xFU); // powered only by SMPS
-  //Set VOS level (VOS3 to 170Mhz, VOS2 to 300Mhz, VOS1 to 400Mhz, VOS0 to 550Mhz)
+
+  // Set VOS level (VOS3 to 170Mhz, VOS2 to 300Mhz, VOS1 to 400Mhz, VOS0 to 550Mhz)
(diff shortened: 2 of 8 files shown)


board/stm32fx/clock.h
```diff
@@ -1,4 +1,8 @@
-void clock_init(void) {
+bool failed_clock_init = false;
```
adeebshihadeh, line 1: should only be defined once. could also use a better name that makes it clear we're in a "limited" mode - current name is a bit misleading when reading code that references it in other places

board/stm32h7/clock.h
```diff
@@ -17,13 +17,35 @@ APB4 per: 60MHz
...
   //Set power mode to direct SMPS power supply(depends on the board layout)
   register_set(&(PWR->CR3), PWR_CR3_SMPSEN, 0xFU); // powered only by SMPS
-  //Set VOS level (VOS3 to 170Mhz, VOS2 to 300Mhz, VOS1 to 400Mhz, VOS0 to 550Mhz)
+
+  // Set VOS level (VOS3 to 170Mhz, VOS2 to 300Mhz, VOS1 to 400Mhz, VOS0 to 550Mhz)
   register_set(&(PWR->D3CR), PWR_D3CR_VOS_1 | PWR_D3CR_VOS_0, 0xC000U); //VOS1, needed for 80Mhz CAN FD
-  while ((PWR->CSR1 & PWR_CSR1_ACTVOSRDY) == 0);
-  while ((PWR->CSR1 & PWR_CSR1_ACTVOS) != (PWR->D3CR & PWR_D3CR_VOS)); // check that VOS level was actually set
+
+  // Wait until VOS level is reached. If flashed through DFU right after power-on, the CR3 register is locked, so this won't work.
```
adeebshihadeh, line 29: can you add something about how this only affects some chips with an older bootloader? and link to a reference from ST if it exists

adeebshihadeh:
Perhaps we build a temporary bootstub with this for IP, then revisit if we can't make the power cycle through fake SOM fast enough?

robbederks:
> Perhaps we build a temporary bootstub with this for IP, then revisit if we can't make the power cycle through fake SOM fast enough?

That would work. Power cycling should be fast with the new fake SOM, but we could just use this in the meantime (also for the app code I guess to make it easier to use) until that arrives.

Don't see a way a device would ever hit this in the field, since the panda power supply always comes up before the SOM boots

adeebshihadeh:
Yeah, I'd just rather not merge this if we don't have to. If we just use in IP, we can enable the app section too.

robbederks:
OK, sounds like a plan.

adeebshihadeh:
Oh, let's keep this open until we're not using it so we can rebuild if necessary.

PR #1572, opened 2023-08-11
panda: add CAN-FD throughput test

briskspirit:


tests/canfd/test_canfd.py
```diff
@@ -40,13 +39,42 @@ def panda_init(serial, enable_canfd=False, enable_non_iso=False):
...
+    # set bus 0 data speed to speed
+    p.set_can_data_speed_kbps(0, tests[0][i])
+    if p_recv is not None:
+      p_recv.set_can_data_speed_kbps(0, tests[0][i])
+    time.sleep(0.05)
+
+    comp_kbps = time_many_sends(p, 0, p_recv=p_recv, msg_count=400, two_pandas=two_pandas, msg_len=64)
+
+    # bit count from https://en.wikipedia.org/wiki/CAN_bus
+    saturation_pct = (comp_kbps / tests[0][i]) * 100.0
```
adeebshihadeh, line 74: should probably be more specific about the overall throughput, not just the data speed

briskspirit: makes sense, will fix. needed mostly to have thresholds before I will lower TX FIFO size (similar to F4) for more control and less losses.

PR #1579, opened 2023-08-14
Ruff: propgate config from OP

jnewb1:


tests/rtc_test.py
```diff
@@ -4,7 +4,7 @@
 import datetime
 
 sys.path.append(os.path.join(os.path.dirname(os.path.realpath(__file__)), ".."))
```
adeebshihadeh, line 6: this just looks like an old hacks, you can remove all of these too

adeebshihadeh (commented):
Nice, these all look like good improvements!

tests/rtc_test.py
jnewb1: done

PR #1584, opened 2023-08-16
Hyundai CAN FD: fix PCM cruise check

sshane:
The parsing of the CANFD `ACC_STATUS->ACCMode` signal only considers the first two bits (0x3 instead of 0x7), so it would consider any value from 1 to 3 to be enaged, and 0 and 4 to be disengaged. Which is mostly correct, since the EV6 route I found has it at 4 when disabled.

When the car enters a fault status (3), only then is there a mismatch between openpilot and panda. Will fix this by:
- correctly parsing the whole 3 bit signal,
- checking the same specific values openpilot does, 
- and adding a test to test_models to check cruise state, not only controlsAllowed, which shows a mismatch in most HKG routes (doesn't need the fault status to come up) - https://github.com/commaai/openpilot/pull/29423


sshane:
TODO: check if CAN FD uses 4 or 0 for disabled mainly, with test_models

PR #1587, opened 2023-08-17
Subaru: gen2 long safety

jnewb1:
there are 2 static signals, and a high beam assist message that we need to emulate to keep the car happy when eyesight is disabled

### DIFF
--- a/board/safety/safety_subaru.h
+++ b/board/safety/safety_subaru.h
@@ -48,18 +48,18 @@ const LongitudinalLimits SUBARU_LONG_LIMITS = {
 #define SUBARU_ALT_BUS  1
 #define SUBARU_CAM_BUS  2
 
-#define SUBARU_COMMON_TX_MSGS(alt_main_bus)         \
+#define SUBARU_COMMON_TX_MSGS(alt_bus)              \
   {MSG_SUBARU_ES_LKAS,         SUBARU_MAIN_BUS, 8}, \
-  {MSG_SUBARU_ES_Distance,     alt_main_bus,    8}, \
+  {MSG_SUBARU_ES_Distance,     alt_bus,         8}, \
   {MSG_SUBARU_ES_DashStatus,   SUBARU_MAIN_BUS, 8}, \
   {MSG_SUBARU_ES_LKAS_State,   SUBARU_MAIN_BUS, 8}, \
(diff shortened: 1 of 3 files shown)


board/safety/safety_subaru.h
```diff
@@ -218,6 +243,15 @@ static int subaru_tx_hook(CANPacket_t *to_send) {
     violation |= longitudinal_transmission_rpm_checks(transmission_rpm, SUBARU_LONG_LIMITS);
   }
 
+  if (addr == MSG_SUBARU_ES_UDS_Request){
+    int sid = GET_BYTES(to_send, 1, 1);
+
+    // Allow "tester present", "read data by identifier", and "communication control"
+    bool allowed_sid = (sid == 0x3e) || (sid == 0x22) || (sid == 0x28);
+
```
sshane, line 251: why do you need communication control? this can allow programming session, perhaps we should block that

board/safety/safety_subaru.h
```diff
@@ -175,7 +195,12 @@ static int subaru_tx_hook(CANPacket_t *to_send) {
   bool violation = false;
 
   if (subaru_gen2) {
-    tx = msg_allowed(to_send, SUBARU_GEN2_TX_MSGS, SUBARU_GEN2_TX_MSGS_LEN);
+    if(subaru_longitudinal){
+      tx = msg_allowed(to_send, SUBARU_GEN2_LONG_TX_MSGS, SUBARU_GEN2_LONG_TX_MSGS_LEN);
+    }
+    else{
+      tx = msg_allowed(to_send, SUBARU_GEN2_TX_MSGS, SUBARU_GEN2_TX_MSGS_LEN);
+    }
```
sshane, line 203: formatting is not consistent

sshane:
can you make sure to request a review anything panda safety related? if adeeb already reviewed then nvm, but there's a couple mistakes here like allowing the diagnostic sid when we don't need it

board/safety/safety_subaru.h
jnewb1: Yup. Communication control is required to disable eyesight TX, unless I'm missing something?

sshane: Ah I suppose we don't check data in ELM mode, so no safety issue here, but init is run before you enter the safety mode, so 0x28 isn't required. (as long as you only need to send communication control once)

adeebshihadeh:
Reverted

PR #1589, opened 2023-08-17
Safety: unify `controls_allowed` with boolean

sunnyhaibin:
Unified the assignment of `controls_allowed` with boolean values. This change improves consistency by using boolean values throughout.

### DIFF
--- a/board/main.c
+++ b/board/main.c
@@ -237,7 +237,7 @@ void tick_handler(void) {
       if (controls_allowed && !heartbeat_engaged) {
         heartbeat_engaged_mismatches += 1U;
         if (heartbeat_engaged_mismatches >= 3U) {
-          controls_allowed = 0U;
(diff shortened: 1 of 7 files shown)


adeebshihadeh:
can you do a quick check to see if there's a cppcheck flag we can set to catch this?

sunnyhaibin:
> can you do a quick check to see if there's a cppcheck flag we can set to catch this?

@adeebshihadeh It does not seem to be something that is available with cppcheck, unfortunately.

sshane:
I only see this: `- Assigning bool value to pointer or float`. This alludes to cppcheck allowing implicit 0 or 1 for bool: `- Returning an integer other than 0 or 1 from a function with boolean return value`

https://sourceforge.net/p/cppcheck/wiki/ListOfChecks/

PR #1590, opened 2023-08-17
Subaru: infrastructure to support the new angle based controllers

jnewb1:


tests/safety/test_subaru.py
```diff
@@ -162,6 +163,30 @@ def _torque_cmd_msg(self, torque, steer_req=1):
...
+    return self.packer.make_can_msg_panda("ES_LKAS_ANGLE", 0, values)
+
+  def _angle_meas_msg(self, angle):
+    values = {"Steering_Angle": angle}
+    return self.packer.make_can_msg_panda("Steering_Torque", 0, values)
+
+  def _pcm_status_msg(self, enable):
+    values = {"Cruise_Activated": enable}
+    return self.packer.make_can_msg_panda("ES_Status", self.ALT_CAM_BUS, values)
+
```
sshane, line 188: So we've figured out that ES_Status->Cruise_Activated has none of the issues I *thought* I found?

board/safety/safety_subaru.h
```diff
@@ -216,6 +230,15 @@ static int subaru_tx_hook(CANPacket_t *to_send) {
     violation |= steer_torque_cmd_checks(desired_torque, -1, limits);
   }
 
+  if (addr == MSG_SUBARU_ES_LKAS_ANGLE) {
+    int desired_angle = (GET_BYTES(to_send, 5, 3) & 0x1FFFFU);
+    desired_angle = to_signed(desired_angle, 17);
+    bool lkas_request = GET_BIT(to_send, 12U);
+
+    const SteeringLimits limits = SUBARU_ANGLE_STEERING_LIMITS;
+    violation |= steer_angle_cmd_checks(desired_angle, lkas_request, limits);
```
sshane, line 239: does the command angle from the stock system go to 0 when disabled, or matches the meas angle, and is it limited to a certain degree? On Toyota LTA for example, it matched meas, stopping at 95 degrees. Anything higher and it faulted the EPS

jnewb1: IIRC it follows the measured steering angle, at least to 300 deg. I'll double check that though

tests/safety/test_subaru.py
jnewb1: It doesn't appear to have the same issue:

https://github.com/commaai/panda/pull/1187#issuecomment-1641283653

sshane: The EyeSight system must listen to at least a fault signal from the PCM, we should find that and log it all the time. Then log the fault bits from the camera only when not openpilotLongitudinalControl (as long as we can control with them set)

martinl: when looking at gen2 CAN network diagram https://github.com/martinl/openpilot/wiki/CAN-Network-diagrams#2020my-outbacklegacy-can-diagram the message that contains the PCM fault signal (previously CruiseControl) should be sent by VDC CM on CAN1. I hope it helps narrowing it down

board/safety/safety_subaru.h
jnewb1: It tracks the angle at least between -433 and 395, I will try and get a route with lock to lock from someone

f3b34c0d2632aa83|2023-07-23--18-57-25
image

jnewb1:
Going to go ahead and merge the infrastructure since it will be required anyway.

board/safety/safety_subaru.h
```diff
@@ -13,6 +13,17 @@
...
+  .angle_deg_to_can = 100,
+  .angle_rate_up_lookup = {
+    {0., 5., 15.},
+    {5., .8, .15}
+  },
+  .angle_rate_down_lookup = {
+    {0., 5., 15.},
+    {5., 3.5, .4}
+  },
+};
```
sshane, line 26: We are trying to reduce complexity by only using two breakpoints, and improve safety by limiting the low speed limit to roughly 3 m/s^3 max at high speed (~75 mph)

board/safety/safety_subaru.h
```diff
@@ -255,7 +278,8 @@ static const addr_checks* subaru_init(uint16_t param) {
   subaru_gen2 = GET_FLAG(param, SUBARU_PARAM_GEN2);
 
 #ifdef ALLOW_DEBUG
-  subaru_longitudinal = GET_FLAG(param, SUBARU_PARAM_LONGITUDINAL) && !subaru_gen2;
+  subaru_longitudinal = GET_FLAG(param, SUBARU_PARAM_LONGITUDINAL);
+  subaru_lkas_angle = GET_FLAG(param, SUBARU_PARAM_LKAS_ANGLE);
```
sshane, line 282: why do we allow longitudinal now?

tests/safety/test_subaru.py
```diff
@@ -151,6 +152,30 @@ def _torque_cmd_msg(self, torque, steer_req=1):
...
+    values = {"LKAS_Output": angle, "LKAS_Request": enabled}
+    return self.packer.make_can_msg_panda("ES_LKAS_ANGLE", 0, values)
+
+  def _angle_meas_msg(self, angle):
+    values = {"Steering_Angle": angle}
+    return self.packer.make_can_msg_panda("Steering_Torque", 0, values)
+
+  def _pcm_status_msg(self, enable):
+    values = {"Cruise_Activated": enable}
+    return self.packer.make_can_msg_panda("ES_Status", self.ALT_CAM_BUS, values)
```
sshane, line 176: I would have split out the code to support the new PCM status message from this angle PR. This changes a lot.

PR #1591, opened 2023-08-17
body: set vehicle_moving

sshane:
this has no effect on safety, this only matches what openpilot sets.


board/safety/safety_body.h
```diff
@@ -11,6 +11,9 @@ static int body_rx_hook(CANPacket_t *to_push) {
 
   bool valid = addr_safety_check(to_push, &body_rx_checks, NULL, NULL, NULL, NULL);
 
+  // body always assumed to be moving
```
sshane, line 14:
```suggestion
  // body is never at standstill
```

PR #1595, opened 2023-08-21
SCons: add option to load git revision from file

jnewb1:
Sometimes when building OP, we only copy the files of the submodules, not the full .git history (ex: https://github.com/commaai/openpilot/pull/29436)

In these cases, git rev-parse fails and we get a cache miss. I instead propose we support loading .git_version from a file, which will allow us to maintain the cache by creating and copying this file manually before build

Loading the full .git history into the docker container will bloat the size of the image


adeebshihadeh:
CI shouldn’t be a special case like this.

PR #1597, opened 2023-08-21
H7: clean up peripherals.h

briskspirit:
We call detect_board() before current_board->init() so no need for that code (and already being initialized as analog, so should save some power too)
@robbederks agree?


robbederks:
lgtm!

robbederks (approved):

PR #1609, opened 2023-08-26
remove rx check for legacy safety mode car

multikyd:
2017 Optima Hybrid fails for checking the 1057 address.


multikyd:
Optima Hybrid 2017 has failed to check the address 1057.

sshane:
Unfortunately your checksum for SCC12/0x421 is slightly different than the other cars. You'll have to first figure out the new algorithm (it looks very similar, there's only an offset between other cars and your car), and then understand why it's different.

One thing I noticed is your car sets bits in byte 6 of SCC12, while our Optima does not (related to AEB status).

visualscopepro:
don't know if it's the right place, i got a KIA_OPTIMA_H car (optima PHEV 2017 from europe), i see that i will use it in dashcam mode for problems in radar checksum, any way i can help for full support? thanks!

PR #1613, opened 2023-08-28
Subaru: add steer req bit test

jnewb1:


sshane:
Like adeeb mentioned in the other PR, it's worth more than a PR to just do one. A test asserting it can't be unset would prevent confusions where we think there's safety applied to https://github.com/commaai/openpilot/pull/29116 but there's not.

jyoung8607:
> Can you also see if it makes sense to set it on all other torque safety modes and test it explicitly, so we never forget to set this properly?

If you move forward with this, VW will need a frame or two of tolerance due to our lockout mitigations.

sshane:
I'll have to check how what VW does, but it looks semi similar to the other ports' workarounds.

Issue #1621, opened 2023-08-30
safety tests: check steer request bit for all cars


sshane:
Related: https://github.com/commaai/panda/pull/1613

- [x] ~~Mazda https://github.com/commaai/panda/pull/1627~~
- [x] Chrysler https://github.com/commaai/panda/pull/1628
- [x] GM https://github.com/commaai/panda/pull/1629
- [x] Subaru - https://github.com/commaai/panda/pull/1613
- [x] Subaru Pre-global - https://github.com/commaai/panda/pull/1626
- [x] Volkswagen MQB https://github.com/commaai/panda/pull/1631
- [x] Volkswagen PQ https://github.com/commaai/panda/pull/1630

Then:
- [ ] Add test to ensure it's never unset - https://github.com/commaai/panda/pull/1625

PR #1625, opened 2023-08-31
Safety: assert that we use the steering req bit

jnewb1:
TODO:

- [x] verify all of these and split into PR for each
- [ ] ~~VW needs to control this bit for fault avoidance, any others?~~

https://github.com/commaai/panda/issues/1621


jnewb1:
just missing vw now

sshane:
example failure catching a safety mode not setting steer_req

```python
======================================================================
FAIL: test_steer_req_bit (__main__.TestGmCameraSafety.test_steer_req_bit)
Asserts all torque safety modes check the steering request bit
----------------------------------------------------------------------
  File "/home/batman/openpilot/panda/tests/safety/common.py", line 300, in test_steer_req_bit
    self.assertFalse(self._tx(self._torque_cmd_msg(self.MAX_TORQUE, 0)))
AssertionError: 1 is not false
```

sshane:
Changed the test to not modify the safety code, think it's better not to add a variable that's not used anywhere in the safety -- as it's more likely prone to bugs (was missing from the init/reset function).

tests/safety/common.py
```diff
@@ -283,6 +285,22 @@ def test_non_realtime_limit_up(self):
...
 
+  def test_steer_req_bit(self):
+    """Asserts all torque safety modes check the steering request bit"""
+    if self.NO_STEER_REQ_BIT:
+      raise unittest.SkipTest("No steering request bit")
+
+    self.safety.set_controls_allowed(True)
+    self._set_prev_torque(self.MAX_TORQUE)
+
+    # Send torque successfully, then only drop the request bit and ensure it's blocked
```
sshane, line 296:
```suggestion
    # Send torque successfully, then only drop the request bit and ensure it stays blocked
```

PR #1627, opened 2023-08-31
Mazda: check steer req bit

jnewb1:


jnewb1:
I don't think mazda has this bit? Not seeing anything change when ACC is activated

board/safety/safety_mazda.h
```diff
@@ -89,7 +89,9 @@ static int mazda_tx_hook(CANPacket_t *to_send) {
     if (addr == MAZDA_LKAS) {
       int desired_torque = (((GET_BYTE(to_send, 0) & 0x0FU) << 8) | GET_BYTE(to_send, 1)) - 2048U;
 
-      if (steer_torque_cmd_checks(desired_torque, -1, MAZDA_STEERING_LIMITS)) {
+      bool steer_req = (desired_torque != 0);
```
adeebshihadeh, line 92: if the bit doesn't exist, we shouldn't pretend like it does

PR #1630, opened 2023-08-31
VW PQ: Check steer req bit

jnewb1:
as far as I can tell, VW doesn't need any special logic around the steer req frames. It only sets HCA_ENABLED to false when apply_steer is zero, which is fine in the current panda logic:

https://github.com/commaai/openpilot/blob/master/selfdrive/car/volkswagen/carcontroller.py#L56
https://github.com/commaai/panda/blob/master/board/safety.h#L573


jyoung8607:
Sorry for the false alarm, I forgot I cleaned that up a few months ago. commaai/openpilot#27624

PR #1631, opened 2023-08-31
VW MQB: Check steer req bit

jnewb1:


adeebshihadeh (approved):

adeebshihadeh:
I believe VW needs some tolerance on this since it drops the bit at 0 due to a lockout that happens after ~6 minutes.

jnewb1:
> I believe VW needs some tolerance on this since it drops the bit at 0 due to a lockout that happens after ~6 minutes.

I made a comment about that in the other PR: https://github.com/commaai/panda/pull/1630
It seems like it isn't required because it will also set apply_steer to zero, which will not trigger the tolerance check: https://github.com/commaai/panda/blob/62bb0c2e339101c041990162c03174017a4b2fe6/board/safety.h#L573

PR #1632, opened 2023-09-01
Subaru: add safety around cutting steer req bit

jnewb1:
for https://github.com/commaai/openpilot/pull/29116

@sshane did I do the math for the rt_interval correct? message is 50hz, and we want to cut the torque once every 8 frames, so 

((8 frames) / (50 frames / second)) * 0.9 (for 10% tolerance) = .144 seconds


jnewb1:
should it also be behind another flag? it's only required for 2020+ models

sshane:
It should be fine to always allow it (no flag).

jnewb1:
@sshane anything else on this?

board/safety/safety_subaru.h
```diff
@@ -1,18 +1,32 @@
...
+    .type = TorqueDriverLimited,                                                    \
+
+
+// the EPS will temporary fault if the steering rate is too high, so we cut   
+// the  steering torque every 7 frames for 1 frame if the steering rate is high
+#define SUBARU_STEERING_RATE_LIMIT_GENERATOR()                                      \
+    .min_valid_request_frames = 7,                                                  \
+    .max_invalid_request_frames = 1,                                                \
+    .min_valid_request_rt_interval = 144000,                                        \
+    .has_steer_req_tolerance = true,                                                \
```
sshane, line 18: can we add to the above? macro name is confusing

tests/safety/test_subaru.py
```diff
@@ -61,6 +61,11 @@ class TestSubaruSafetyBase(common.PandaSafetyTest, MeasurementSafetyTest):
 
   INACTIVE_GAS = 1818
 
+  # Safety around steering req bit
+  MIN_VALID_STEERING_FRAMES = 7
+  MAX_INVALID_STEERING_FRAMES = 1
+  MIN_VALID_STEERING_RT_INTERVAL = 144000
+
```
sshane, line 68: might be a good idea to make these couple of tests a new test class, so it's more explicit when adding this safety.

board/safety/safety_subaru.h
```diff
@@ -1,18 +1,32 @@
...
+    .driver_torque_allowance = 60,                                                  \
+    .type = TorqueDriverLimited,                                                    \
+
+
+// the EPS will temporary fault if the steering rate is too high, so we cut   
+// the  steering torque every 7 frames for 1 frame if the steering rate is high
+#define SUBARU_STEERING_RATE_LIMIT_GENERATOR()                                      \
+    .min_valid_request_frames = 7,                                                  \
+    .max_invalid_request_frames = 1,                                                \
+    .min_valid_request_rt_interval = 144000,                                        \
```
sshane, line 17: looks good, I used `1000/(50/8)*0.9` which is also 144ms

board/safety/safety_subaru.h
jnewb1: done

tests/safety/test_subaru.py
jnewb1: made a common test

board/safety/safety_subaru.h
```diff
@@ -1,14 +1,21 @@
...
+    .max_steer = (steer_max),                                                         \
+    .max_rt_delta = 940,                                                              \
+    .max_rt_interval = 250000,                                                        \
+    .max_rate_up = (rate_up),                                                         \
+    .max_rate_down = (rate_down),                                                     \
+    .driver_torque_factor = 50,                                                       \
+    .driver_torque_allowance = 60,                                                    \
+    .type = TorqueDriverLimited,                                                      \
+    /* the EPS will temporary fault if the steering rate is too high, so we cut the   \
+      the steering torque every 7 frames for 1 frame if the steering rate is high */  \
```
sshane, line 12:
```suggestion
    // the EPS will temporary fault if the steering rate is too high, so we cut the   \
    // the steering torque every 7 frames for 1 frame if the steering rate is high   \
```

tests/safety/common.py
```diff
@@ -288,6 +283,18 @@ def test_non_realtime_limit_up(self):
...
+
+  # Safety around steering req bit
+  MIN_VALID_STEERING_FRAMES: Optional[int] = None
+  MAX_INVALID_STEERING_FRAMES: int = 1
+  MIN_VALID_STEERING_RT_INTERVAL: Optional[int] = None
+
+  def test_steer_req_parameters(self):
+    # assert the parameters are properly set
+    self.assertIsNotNone(self.MIN_VALID_STEERING_FRAMES)
+    self.assertIsNotNone(self.MIN_VALID_STEERING_RT_INTERVAL)
```
sshane, line 296: I'd remove the checks below so things will fail if they are 0. Can you also make this a new PR?

tests/safety/common.py
```diff
@@ -288,6 +283,18 @@ def test_non_realtime_limit_up(self):
     self._set_prev_torque(0)
     self.assertFalse(self._tx(self._torque_cmd_msg(-self.MAX_RATE_UP - 1)))
 
+class SteerRequestCutTestBase(unittest.TestCase, abc.ABC):
+
+  # Safety around steering req bit
+  MIN_VALID_STEERING_FRAMES: Optional[int] = None
+  MAX_INVALID_STEERING_FRAMES: int = 1
+  MIN_VALID_STEERING_RT_INTERVAL: Optional[int] = None
```
sshane, line 291:
```suggestion
  # Safety around steering req bit
  MIN_VALID_STEERING_FRAMES: int
  MAX_INVALID_STEERING_FRAMES: int = 1
  MIN_VALID_STEERING_RT_INTERVAL: int
```

can we do this?

board/safety/safety_subaru.h
jnewb1: this is a quirk of the macros, everything is actually on one line when it's expanded, so if we use //, then everything after that will be commented out

sshane, line 12:
```suggestion
       the steering torque every 7 frames for 1 frame if the steering rate is high */ \
```

jnewb1: I haven't found a way to have proper multiline macros

tests/safety/common.py
jnewb1: made a separate PR

tests/safety/common.py
jnewb1: done in new PR

sshane (approved):

board/safety/safety_subaru.h
```diff
@@ -1,14 +1,21 @@
...
+    .max_rate_down = (rate_down),                                                     \
+    .driver_torque_factor = 50,                                                       \
+    .driver_torque_allowance = 60,                                                    \
+    .type = TorqueDriverLimited,                                                      \
+    /* the EPS will temporary fault if the steering rate is too high, so we cut the   \
+       the steering torque every 7 frames for 1 frame if the steering rate is high */ \
+    .min_valid_request_frames = 7,                                                    \
+    .max_invalid_request_frames = 1,                                                  \
+    .min_valid_request_rt_interval = 144000,                                          \
+    .has_steer_req_tolerance = true,                                                  \
```
sshane, line 16:
```suggestion
    .max_invalid_request_frames = 1,                                                  \
    .min_valid_request_rt_interval = 144000,  /* 10% tolerance */          \
    .has_steer_req_tolerance = true,                                                  \
```

PR #1637, opened 2023-09-04
Hyundai: Car Port for Nexo 2021

sunnyhaibin:
**Prerequisite of**
- https://github.com/commaai/opendbc/pull/933


sshane:
- [ ] Needs tests

jyoung8607 (changes requested):
Looks reasonable at a glance, but as Shane mentioned, will need a test added. That's why CI isn't passing, because the FCEV code path isn't exercised at all.

sunnyhaibin:
> Looks reasonable at a glance, but as Shane mentioned, will need a test added. That's why CI isn't passing, because the FCEV code path isn't exercised at all.

Updated with unit test for FCEV.

sunnyhaibin:
> Looks reasonable at a glance, but as Shane mentioned, will need a test added. That's why CI isn't passing, because the FCEV code path isn't exercised at all.

Unite test updated and all passed.

jyoung8607 (approved):
LGTM

sunnyhaibin:
Awaiting for test route from user `simon402_61863 (Discord)`.

jyoung8607:
The actual code here is GTM, but the original public test route has gone away, and we shouldn't merge this component without an immediate path to merging the opendbc port.

sunnyhaibin:
Safety is now migrated to `opendbc`!
- https://github.com/commaai/opendbc/pull/933

PR #1647, opened 2023-09-05
enable red and green leds until USB/SPI init

briskspirit:
Will help support to better understand if panda froze before USB/SPI init


briskspirit:
think should change it to red+green, as red itself can indicate freeze during panda heartbeat.

PR #1648, opened 2023-09-05
Subaru Pre-Global: Reverse driver torque signal for certain platforms

sunnyhaibin:
This PR resolves permanent steering faults when the user overrides steering and then releases the steering wheel. Thanks to @martinl for this fix from `subaru-community`.

Route ID: `b81cc3ea44acd372|2023-09-06--15-36-44`

**Affected Platforms**
- Subaru Forester 2017-18
- Subaru Outback ~~2015-17~~ 2018-2019

**Related**
- https://github.com/commaai/openpilot/pull/29782


sshane:
We should always add a test for new safety params. This wasn't caught because it's behind dashcamOnly, correct?

sunnyhaibin:
> 

Yep, correct.

sunnyhaibin:
> We should always add a test for new safety params. This wasn't caught because it's behind dashcamOnly, correct?

Yep, correct. With the latest changes, it seems that Pre-Global Subaru platforms should be ready to be taken out of dashcamOnly with this being properly implemented.

tests/safety/test_subaru_preglobal.py
```diff
@@ -61,5 +61,12 @@ def _pcm_status_msg(self, enable):
     return self.packer.make_can_msg_panda("CruiseControl", 0, values)
 
 
+class TestSubaruPreglobalReversedDriverTorqueSafety(TestSubaruPreglobalSafety):
+  FLAGS = Panda.FLAG_SUBARU_PREGLOBAL_REVERSED_DRIVER_TORQUE
+
+  def _torque_driver_msg(self, torque):
+    return super()._torque_cmd_msg(-torque)
```
adeebshihadeh, line 68: @jnewb1 the tests don't fail if you remove the sign from here

jnewb1: fixed by using correct DBC

PR #1655, opened 2023-09-07
Hyundai CAN: utilize macros for common address checks

sshane:
No change in behavior.


sshane:
@sunnyhaibin any changes to make this more compatible with the Palisade port?

sunnyhaibin:
> @sunnyhaibin any changes to make this more compatible with the [Palisade port](https://github.com/commaai/panda/pull/1243)?

This looks good to me!

PR #1657, opened 2023-09-07
safety: enforce expected messages in rx hook

sshane:
Valid should be false if message is not in address checks. Catches a lot of missing checks/bugs in test_models
- [x] race condition with alt brake message (both exist on sienna)
- [ ] interceptor not in address checks

There are some impacts to think about with the interceptor safety param:

- If openpilot were not to set it when there's an interceptor, panda would think openpilot gas is user gas and disengage. That's fine.
- If openpilot were to set it when there is no interceptor, panda would not detect user gas, but address checks should prevent enabling. That's fine.

Meanwhile:

- Currently if we're using interceptor and its message dropped out and openpilot were to stay engaged, panda would not notice anything wrong and stay engaged, not detecting any user gas. This is not fine.

So the interceptor param addition should strictly be safe.


board/safety/safety_hyundai_canfd.h
```diff
@@ -111,6 +111,8 @@ AddrCheckStruct hyundai_canfd_ice_addr_checks[] = {
   {.msg = {{0xea, 0, 24, .check_checksum = true, .max_counter = 0xffU, .expected_timestep = 10000U}, { 0 }, { 0 }}},
   {.msg = {{0x175, 0, 24, .check_checksum = true, .max_counter = 0xffU, .expected_timestep = 20000U}, { 0 }, { 0 }}},
   {.msg = {{0x1aa, 0, 16, .check_checksum = false, .max_counter = 0xffU, .expected_timestep = 20000U}, { 0 }, { 0 }}},
+  {.msg = {{0x1a0, 0, 32, .check_checksum = true, .max_counter = 0xffU, .expected_timestep = 20000U},
+           {0x1a0, 2, 32, .check_checksum = true, .max_counter = 0xffU, .expected_timestep = 20000U}, { 0 }}},
```
sshane, line 115: can't do this, race condition. need new rx check list, pending a PR like https://github.com/commaai/panda/pull/1655 for CAN FD

sshane:
Pre-req for Hyundai cleanups, found bugs there too and want to start with one brand. Also requires a safety param for the interceptor which isn't in the address checks, but dynamically detected.

sshane:
not sure how to handle the interceptor param yet, don't want to double the addr checks (?)

PR #1659, opened 2023-09-08
CI: run each job only once

adeebshihadeh:


adeebshihadeh:
@jnewb1 can you update the rest of the submodules with this?

PR #1661, opened 2023-09-13
erase log flash sectors in DFU recovery

adeebshihadeh:


robbederks:
Looks like this will also erase the last sector on the H7s, which we use as "OTP" to store provisioning data, we have to be very careful to not do that!

adeebshihadeh:
> Looks like this will also erase the last sector on the H7s, which we use as "OTP" to store provisioning data, we have to be very careful to not do that!

The python library intentionally doesn't "know" about that sector. https://github.com/commaai/panda/blob/90eb24bf1922e9c33ae4885029e3595228d8fc95/python/constants.py#L40

robbederks:
lgtm then

robbederks (approved):

PR #1669, opened 2023-09-25
jungle v2: control power individually per port

briskspirit:


board/jungle/main_comms.h
```diff
@@ -58,6 +58,10 @@ int comms_control_handler(ControlPacket_t *req, uint8_t *resp) {
     case 0xa2:
       current_board->set_ignition((req->param1 == 1U));
       break;
+    // **** 0xa0: Set panda power per channel by bitmask.
+    case 0xa3:
+      current_board->set_panda_individual_power(req->param1);
```
adeebshihadeh, line 63: bitmask isn't a nice api. how about first param is port and second is whether it's powered

PR #1673, opened 2023-09-29
Safety: Only allow tester present TX with openpilot longitudinal

sunnyhaibin:
We only send tester present on ECU disable when openpilot longitudinal control is enabled. This PR enforces that tester present TX is only allowed when the openpilot longitudinal control flag is set.
- Honda Bosch
- Hyundai CAN
- Hyundai CAN-FD
- Subaru (Global)

### DIFF
--- a/board/safety/safety_subaru.h
+++ b/board/safety/safety_subaru.h
@@ -232,7 +232,7 @@ static int subaru_tx_hook(CANPacket_t *to_send) {
   if (addr == MSG_SUBARU_ES_Distance) {
     int cruise_throttle = (GET_BYTES(to_send, 2, 2) & 0xFFFU);
     bool cruise_cancel = GET_BIT(to_send, 56U) != 0U;
-    
+
     if (subaru_longitudinal) {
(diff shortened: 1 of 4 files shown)


board/safety/safety_subaru.h
```diff
@@ -256,7 +256,7 @@ static int subaru_tx_hook(CANPacket_t *to_send) {
     // reading ES button data by identifier (b'\x03\x22\x11\x30\x00\x00\x00\x00') is also allowed (DID 0x1130)
     bool is_button_rdbi = (GET_BYTES(to_send, 0, 4) == 0x30112203U) && (GET_BYTES(to_send, 4, 4) == 0x0U);
 
-    violation |= !(is_tester_present || is_button_rdbi);
+    violation |= !(is_tester_present || is_button_rdbi) || !subaru_longitudinal;
```
jnewb1, line 259: for subaru, this is already checked by having a different set of allowed messages when using long

adeebshihadeh:
Like @jnewb1 said, this is already handled by the TX msgs whitelist for each safety mode.

PR #1675, opened 2023-09-30
fix siren false positive while offroad

adeebshihadeh:
Fixes the siren playing when going offroad quickly after being engaged and unplugging the device.


adeebshihadeh:
Needs some more state since it's possible to get a heartbeat quickly followed by a cable unplug, then ignition would be low.

PR #1678, opened 2023-10-03
Safety: test driver torque message for all brands

jnewb1:
we never test the driver torque message for most of the brands

TODO: figure out why hyundai has a scale factor of 0.79, since that's not in the dbc


jnewb1:
> If this is to catch https://github.com/commaai/panda/pull/1648/files#r1343478755, did you check why the existing driver torque safety tests didn't catch it?

It sets the torque directly with the helpers in those tests, rather than using the message

sshane:
If we don't use the safety helper for other safety modes, then let's just use the CAN message generation

PR #1679, opened 2023-10-03
safety tests: fix measurement test

sshane:
- doesn't actually check within 1 decimal place like the comment says
- if no abs, and on negative values, it doesn't actually check errors

fix for https://github.com/commaai/panda/pull/1528

### DIFF
--- a/tests/safety/common.py
+++ b/tests/safety/common.py
@@ -636,11 +636,11 @@ class MeasurementSafetyTest(PandaSafetyTestBase):
 
   def common_measurement_test(self, msg_func, min_value, max_value, factor, get_min_func, get_max_func):
     for val in np.arange(min_value, max_value, 0.5):
-      # if val > -10:
-      #   continue
+      if val > 10:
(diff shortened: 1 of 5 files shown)


tests/safety/common.py
```diff
@@ -640,8 +640,8 @@ def common_measurement_test(self, msg_func, min_value, max_value, factor, get_mi
         self.assertTrue(self._rx(msg_func(val + i * 0.1)))
 
       # assert close by one decimal place
-      self.assertLessEqual(abs(get_min_func() - val * factor), 1 * abs(factor))
-      self.assertLessEqual(abs(get_max_func() - (val + 0.5) * factor), 1 * abs(factor))
+      self.assertLessEqual(abs(get_min_func() / factor - val), 1)
```
sshane, line 643: TODO: abs also doesn't seem to matter

sshane:
@jnewb1 make sure when refactoring tests that behavior doesn't change

PR #1684, opened 2023-10-04
Add volvo c1 safety model

danielzmod:
Hi!
This is a cleaned up version and improved safety model. This safety model is for Volvo C1 vehicles.

The safety values comes from testing in different speed and forcing max steering request. The values is the absolut highest possibly allowed values to stay within your internal requirement to not deviate out of lane within 1 second.

Best regards
Daniel

### DIFF
--- a/board/safety.h
+++ b/board/safety.h
@@ -18,6 +18,7 @@
 #include "safety/safety_volkswagen_pq.h"
 #include "safety/safety_elm327.h"
 #include "safety/safety_body.h"
+#include "safety/safety_volvo.h"
 
 // CAN-FD only safety modes
 #ifdef CANFD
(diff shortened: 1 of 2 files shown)


adeebshihadeh:
Is this platform still in production? How many new cars does this cover?

danielzmod:
> Is this platform still in production? How many new cars does this cover?

No it’s not. It will add one car Volvo V40. 

Based on the old pull request (https://github.com/commaai/openpilot/pull/1970). I’m stripping out everything extra to get an absolute minimum.

adeebshihadeh:
Nice to see that the diff to support this has gotten much smaller over the last couple years!  Since this only adds support for one car, it doesn't make sense for us to merge and support it. However, we'd accept generic PRs that cleanup/abstract upstream panda and openpilot such that supporting this car in your fork is mostly a matter of dropping in a few files, e.g. a PR that reserves a safety mode for cases like this.

PR #1189, opened 2022-12-08
Toyota: LTA steering control safety

sshane:

PR #1533, opened 2023-07-27
Subaru: LKAS Angle based car safety

jnewb1:
We have so far verified that the 2022 forester and 23 outback use an angle-request based LKAS message.

22 forester is similar to gen1, where everything we care about is on bus0/2
23 outback is similar to gen2, where some messages are on bus1

PR #1694, opened 2023-10-23
safety tests: clean up measurement tests

sshane:
Split from LTA PR https://github.com/commaai/panda/pull/1189, added in: https://github.com/commaai/panda/pull/1528

This measurement safety test that's specifically for angle should not have been split out into a generic test that can't be modified from the angle test. For example, we need to parameterize the max angle for Ford and Toyota LTA.

Subaru should just have used the angle test class

### DIFF
--- a/tests/safety/common.py
+++ b/tests/safety/common.py
@@ -99,16 +99,21 @@ class PandaSafetyTestBase(unittest.TestCase):
         should_tx = (should_tx or v == inactive_value) and msg_allowed
         self.assertEqual(self._tx(msg_function(v)), should_tx, (controls_allowed, should_tx, v))
 
-  def _common_measurement_test(self, msg_func: Callable, min_value: float, max_value: float, factor: int,
-                               meas_min_func: Callable[[], int], meas_max_func: Callable[[], int]):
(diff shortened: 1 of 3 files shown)


tests/safety/common.py
```diff
@@ -99,6 +99,22 @@ def _generic_limit_safety_check(self, msg_function: MessageFunction, min_allowed
         should_tx = (should_tx or v == inactive_value) and msg_allowed
         self.assertEqual(self._tx(msg_function(v)), should_tx, (controls_allowed, should_tx, v))
 
+  def _common_measurement_test(self, msg_func: Callable, min_value: float, max_value: float, factor: int,
+                               meas_min_func: Callable[[], int], meas_max_func: Callable[[], int]):
+    """Tests accurate measurement parsing, and that it's reset on safety mode init"""
```
sshane, line 104:
```suggestion
    """Tests accurate measurement parsing, and that the struct is reset on safety mode init"""
```

sshane:
@jnewb1 can you clean up and merge https://github.com/commaai/panda/pull/1533 behind debug flag?

PR #1699, opened 2023-10-26
safety: test coverage

sshane:
Follow up to https://github.com/commaai/panda/pull/1696

PR #1705, opened 2023-11-02
safety tests: safety_defaults

sshane:
Split from https://github.com/commaai/panda/pull/1699, adds tests for ALLOUTPUT and NOOUTPUT

Split the massive `PandaSafetyTest` class into two: one with car-specific tests like for gas, brake, speed, PCM response, engaging/disengaging, etc, and one for more general tests like ELM327, ALLOUTPUT, and NOOUTPUT.


tests/safety/test_defaults.py
```diff
@@ -0,0 +1,70 @@
+#!/usr/bin/env python3
```
adeebshihadeh, line 1: "defaults" is not an accurate name for these safety modes

sshane: Copied the convention we use for all other safety modes/tests: `safety_defaults.h`: `test_defaults.py` so it seems clear to me. It also tests the default rx and fwd hook.

tests/safety/test_defaults.py
```diff
@@ -0,0 +1,81 @@
...
+from panda import Panda
+from panda.tests.libpanda import libpanda_py
+
+
+class TestDefaultRxHookBase(common.PandaSafetyTest):
+  def test_rx_hook(self):
+    # default rx hook allows all msgs
+    for bus in range(4):
+      for addr in self.SCANNED_ADDRS:
+        self.assertTrue(self._rx(common.make_msg(bus, addr, 8)), f"not allowed RX {addr=}")
```
sshane, line 15:
```suggestion
        self.assertTrue(self._rx(common.make_msg(bus, addr, 8)), f"failed RX {addr=}")
```

PR #1690, opened 2023-10-12
refactor bootkick

adeebshihadeh:
prep for reset

PR #1712, opened 2023-11-09
SOM boot recovery

adeebshihadeh:
- [x] logging
- [x] HITL test
- [x] check GPIO + serial

### DIFF
--- a/board/drivers/bootkick.h
+++ b/board/drivers/bootkick.h
@@ -2,7 +2,12 @@ bool bootkick_ign_prev = false;
 BootState boot_state = BOOT_BOOTKICK;
 uint8_t bootkick_harness_status_prev = HARNESS_STATUS_NC;
 
+uint8_t boot_reset_countdown = 0;
+uint8_t waiting_to_boot_countdown = 0;
+bool bootkick_reset_triggered = false;
+
 void bootkick_tick(bool ignition, bool recent_heartbeat) {
+  BootState boot_state_prev = boot_state;
   const bool harness_inserted = (harness.status != bootkick_harness_status_prev) && (harness.status != HARNESS_STATUS_NC);
 
   if ((ignition && !bootkick_ign_prev) || harness_inserted) {
@@ -11,6 +16,33 @@ void bootkick_tick(bool ignition, bool recent_heartbeat) {
   } else if (recent_heartbeat) {
     // disable bootkick once openpilot is up
     boot_state = BOOT_STANDBY;
+  } else {
+
+  }
+
+  // ensure SOM boots
+  if ((boot_state == BOOT_BOOTKICK) && (boot_state_prev == BOOT_STANDBY)) {
+    waiting_to_boot_countdown = 45U;
+  }
+  if (waiting_to_boot_countdown > 0U) {
+    if (current_board->read_som_gpio() || (boot_state != BOOT_BOOTKICK)) {
+      waiting_to_boot_countdown = 0U;
+    } else {
+      waiting_to_boot_countdown -= 1U;
+
+      // try a reset
+      if (waiting_to_boot_countdown == 0U) {
+        boot_reset_countdown = 5U;
+      }
+    }
+  }
+
+  if (boot_reset_countdown > 0U) {
+    boot_reset_countdown--;
+    boot_state = BOOT_RESET;
+    bootkick_reset_triggered = true;
+  } else if (boot_state == BOOT_RESET) {
+    boot_state = BOOT_BOOTKICK;
   } else {
 
   }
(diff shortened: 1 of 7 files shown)


board/drivers/bootkick.h
```diff
@@ -0,0 +1,54 @@
...
+
+  if ((ignition && !bootkick_ign_prev) || harness_inserted) {
+    // bootkick on rising edge of ignition or harness insertion
+    boot_state = BOOT_BOOTKICK;
+  } else if (recent_heartbeat) {
+    // disable bootkick once openpilot is up
+    boot_state = BOOT_STANDBY;
+  } else {
+
+  }
```
sshane, line 21: This is missing `usb_enumerated` from the original `dos` logic, does that matter? This code that does a similar check says to fall back to USB enumeration where it's supported (not sure of the differences): https://github.com/commaai/panda/blob/cd0f0bd034dde6814b4f5aa12da96a6f5631d141/board/main.c#L275C16-L275C22

board/drivers/bootkick.h
```diff
@@ -0,0 +1,54 @@
...
+
+  // ensure SOM boots
+  if ((boot_state == BOOT_BOOTKICK) && (boot_state_prev == BOOT_STANDBY)) {
+    waiting_to_boot_countdown = 45U;
+  }
+  if (waiting_to_boot_countdown > 0U) {
+    if (current_board->read_som_gpio() || (boot_state != BOOT_BOOTKICK)) {
```
sshane, line 28: same here, close if not an issue

board/drivers/bootkick.h
```diff
@@ -0,0 +1,54 @@
...
+      }
+    }
+  }
+
+  if (boot_reset_countdown > 0U) {
+    boot_reset_countdown--;
+    boot_state = BOOT_RESET;
+    bootkick_reset_triggered = true;
+  } else if (boot_state == BOOT_RESET) {
+    boot_state = BOOT_BOOTKICK;
```
sshane, line 45: This won't cause `waiting_to_boot_countdown` to be set again, so it only tries reset once. Desired?

board/drivers/bootkick.h
```diff
@@ -0,0 +1,54 @@
+bool bootkick_ign_prev = false;
+BootState boot_state = BOOT_BOOTKICK;
+uint8_t bootkick_harness_status_prev = HARNESS_STATUS_NC;
+
+uint8_t boot_reset_countdown = 0;
+uint8_t waiting_to_boot_countdown = 0;
+bool bootkick_reset_triggered = false;
+
```
sshane, line 8: Might want to make a reset function for the various state variables. For example if the device sends a heartbeat or receives another ignition before `boot_reset_countdown` is 0, it will leave the reset state only to enter it again again on the next loop.

board/drivers/bootkick.h
adeebshihadeh: yes, this is desired

board/drivers/bootkick.h
adeebshihadeh: https://github.com/commaai/panda/pull/1690#discussion_r1387449917

robbederks:
As discussed, we shouldn't only rely on the GPIO being set, but also look for serial messages to detect a boot. This would be compatible with all XBLs that we have shipped with 3X.

adeebshihadeh:
@robbederks can you do a final review? This should only trigger in a very narrow case, so any potential FPs shouldn't be too bad.

board/drivers/bootkick.h
```diff
@@ -15,8 +21,47 @@ void bootkick_tick(bool ignition, bool recent_heartbeat) {
 
   }
 
+  /*
+    Ensure SOM boots in case it goes into QDL mode. Reset behavior:
+    * shouldn't trigger on the first boot after power-on
+    * only try reset once per bootkick, i.e. don't keep trying until booted
```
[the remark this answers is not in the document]
adeebshihadeh: maybe this is worded in a confusing way, but the reset should only happen once per panda boot, and consequently once per bootkick

board/drivers/bootkick.h
```diff
@@ -15,8 +21,47 @@ void bootkick_tick(bool ignition, bool recent_heartbeat) {
...
+    * only try once per panda boot, since openpilot will reset panda on startup
+    * once BOOT_RESET is triggered, it stays until countdown is finished
+  */
+  if (!bootkick_reset_triggered && (boot_state == BOOT_BOOTKICK) && (boot_state_prev == BOOT_STANDBY)) {
+    waiting_to_boot_countdown = 45U;
+  }
+  if (waiting_to_boot_countdown > 0U) {
```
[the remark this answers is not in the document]
adeebshihadeh: `(boot_state != BOOT_BOOTKICK)` will trip and the "waiting to boot" state will clear

PR #1716, opened 2023-11-10
safety: add body safety tests

sshane:
Split from https://github.com/commaai/panda/pull/1699

- Added address check for CAN flasher - referenced from https://github.com/commaai/body/commit/1fc01d70715aa943dc6c95366f993a706a5d2a2e#diff-bcec6c3a8fbf0fa79219486d3b495204d8f58664feafeecdc27f0efa2592a4bb
- Made controls allowed logic more explicit, valid never goes false after first valid message.
- Check address so not any message causes controls_allowed to rise

Will test on body to make sure these bugs weren't being relied on to work

### DIFF
--- a/board/safety/safety_body.h
+++ b/board/safety/safety_body.h
@@ -14,7 +14,9 @@ static int body_rx_hook(CANPacket_t *to_push) {
   // body is never at standstill
   vehicle_moving = true;
 
-  controls_allowed = valid;
+  if (valid && (addr == 0x201)) {
+    controls_allowed = true;
+  }
 
   return valid;
 }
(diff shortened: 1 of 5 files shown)


board/safety/safety_body.h
```diff
@@ -33,8 +36,10 @@ static int body_tx_hook(CANPacket_t *to_send) {
     tx = 1;
   }
 
-  // Allow going into CAN flashing mode even if controls are not allowed
-  if (!controls_allowed && (GET_BYTES(to_send, 0, 4) == 0xdeadfaceU) && (GET_BYTES(to_send, 4, 4) == 0x0ab00b1eU)) {
+  // Allow going into CAN flashing mode for base & knee even if controls are not allowed
+  bool flash_msg = ((addr == 0x250) || (addr == 0x350)) && (len == 8);
+  if (!controls_allowed && (GET_BYTES(to_send, 0, 4) == 0xdeadfaceU) && (GET_BYTES(to_send, 4, 4) == 0x0ab00b1eU) &&
+      flash_msg) {
```
sshane, line 42:
```suggestion
  if (!controls_allowed && (GET_BYTES(to_send, 0, 4) == 0xdeadfaceU) && (GET_BYTES(to_send, 4, 4) == 0x0ab00b1eU) && flash_msg) {
```

PR #1717, opened 2023-11-11
Volkswagen PQ sfety: check counter for ABS wheel speed message

sshane:
This causes a mismatch for https://github.com/commaai/openpilot/pull/30443. @jyoung8607 any reason not to check counter? It was switched from a different message without a counter here, and this passes `test_models.py`: https://github.com/commaai/panda/pull/730


jyoung8607:
Sorry, just saw this. It's okay to check the counter so far as I know. Not sure why this is closed, LMK if you want a PR for it.

sshane:
@jyoung8607 fyi, this just worked in the fuzzy carstate test, so should be fine to open and run test_models on a few segments if you have any

PR #1719, opened 2023-11-12
chrysler longitudinal

gregjhogan:
TODO:
- [x] limit for `DAS_3.ENGINE_TORQUE_REQUEST`
- [x] limit for `DAS_3.ACC_DECEL`
- [x] tests


tests/safety/common.py
```diff
@@ -842,6 +842,12 @@ def test_tx_hook_on_wrong_safety_mode(self):
               continue
             if {attr, current_test}.issubset({'TestVolkswagenMqbSafety', 'TestVolkswagenMqbStockSafety', 'TestVolkswagenMqbLongSafety'}):
               continue
+            if {attr, current_test}.issubset({'TestChryslerRamDTSafety', 'TestChryslerRamDTLongitudinalSafety'}):
+              continue
+            # TODO: no sure this is OK
+            if {attr, current_test}.issubset({'TestChryslerSafety', 'TestChryslerLongitudinalSafety',
+                                              'TestChryslerRamHDSafety', 'TestChryslerRamHDLongitudinalSafety'}):
+              continue
```
gregjhogan, line 850: some addresses overlap between these platforms, is this the correct way to handle that?

tests/safety/test_chrysler.py
```diff
@@ -121,5 +123,125 @@ def _speed_msg(self, speed):
...
+                                     self.MIN_ACCEL, self.MAX_ACCEL,
+                                     self.MIN_POSSIBLE_ACCEL, self.MAX_POSSIBLE_ACCEL,
+                                     test_delta=0.1, inactive_value=None)
+    self._generic_limit_safety_check(self._send_torque_msg,
+                                     self.MIN_ENGINE_TORQUE, self.MAX_ENGINE_TORQUE,
+                                     self.MIN_POSSIBLE_ENGINE_TORQUE, self.MAX_POSSIBLE_ENGINE_TORQUE,
+                                     test_delta=10, inactive_value=None)
+
+  def test_buttons(self):
+    button_combinations = set(itertools.permutations([0,0,0,0,1,1,1,1], 4))
```
gregjhogan, line 195: this is an inefficient way to generate all permutations of possible button combinations, is there something better?

sshane: itertools.chain?

tests/safety/common.py
gregjhogan: seems like this is correct since the overlapping addrs have the same definition

tests/safety/test_chrysler.py
```diff
@@ -96,8 +101,9 @@ def _speed_msg(self, speed):
     values = {"Vehicle_Speed": speed}
     return self.packer.make_can_msg_panda("ESP_8", 0, values)
 
+
 class TestChryslerRamHDSafety(TestChryslerSafety):
-  TX_MSGS = [[0x275, 0], [0x276, 0], [0x23A, 2]]
+  TX_MSGS = [[0x23A, 2], [0x275, 0], [0x276, 0]]
```
gregjhogan, line 106: this change makes the order consistent with the other platforms (button message first)

board/safety/safety_chrysler.h
```diff
@@ -249,35 +372,43 @@ static int chrysler_fwd_hook(int bus_num, int addr) {
   int bus_fwd = -1;
 
   // forward to camera
-  if (bus_num == 0) {
+  const bool is_buttons = (addr == chrysler_addrs->CRUISE_BUTTONS);
+  if ((bus_num == 0) && (!chrysler_longitudinal || !is_buttons)) {
```
sshane, line 376: Why are we blocking the button message to the camera/DASM? Won't this fault it if we don't send in openpilot? Usually we let this message through in other ports.

gregjhogan: I don't think I tried without blocking the buttons, so it may work fine.  openpilot doesn't send any button messages for this safety mode, see openpilot/chrysler-long2 branch for how it works on that side.

sshane: Looks like there's no need to block the button message then. We usually only do that if openpilot replaces the message we're blocking

sshane:
I don't have enough time to fully review, but this PR looks quite large. Is there anything that can be split out, such as the address macro refactor? That would make things quicker to review, hard to know what is required and what is part of a clean up.

gregjhogan:
There are probably several PRs you could split this up into (RX checks and essentially 3 sub-platforms).  This has been open so long now that I am also guessing there are other general improvements I am unaware of that need to be incorporated, too.  If this isn't mergeable and you find time to work on this maybe you just want to create new PRs and use this one as a reference.  I probably won't have time to in the near future, but the tests should be quite good.

sshane:

PR #1722, opened 2023-11-17
safety: common Honda addr checks

sshane:


sshane:
better to fix at the same time: https://github.com/commaai/panda/pull/1723

PR #1732, opened 2023-11-20
rx hooks: check address whitelisted

sshane:
try 2 of https://github.com/commaai/panda/pull/1657, enabled by https://github.com/commaai/panda/commit/3d2f99dd40ed391b277edf6788d6beec1fc32390.

Now for safety_rx_hook to see msg, it needs to match address, bus, and length of an address in `RxCheck` struct. Should catch Honda Bosch bug and others like it, where actually-used addresses have their rx checks ignored: https://github.com/commaai/panda/pull/1723

todo:
- [ ] Decide which is cleaner: new hook just to check relay malfunctions, or new struct to specify relay malfunction addresses (for latter, would need multiple structs as addresses, buses are different across safety params)


sshane:

PR #1734, opened 2023-11-21
safety: common Honda rx checks

sshane:
Cleaning up first before https://github.com/commaai/panda/pull/1723 since it makes the change easier to read.


sshane:
Nvm, this adds this bug to alt Nidec param

PR #1735, opened 2023-11-21
safety: add safety param for gas interceptor

sshane:
- Tests that Honda Nidec switches from a PCM-enable car to button-enable car when it's using an interceptor
- Tests that interceptor can't be used when Toyota is in stock long mode

todo:

- [x] Honda
- [x] Separate test class
- [x] Clean up detection (remove)
- [x] ~Make param common~ would take away useful slots from safety models which would never use it (for the time being)

The two possible scenarios where the param is set incorrectly should be fine:
- set if no interceptor - addr checks prevent enabling
- not set and there's an interceptor - panda will see openpilot gas as user gas and cancel long


board/safety/safety_toyota.h
```diff
@@ -304,13 +334,22 @@ static safety_config toyota_init(uint16_t param) {
...
   toyota_dbc_eps_torque_factor = param & TOYOTA_EPS_FACTOR;
 
+  // Gas interceptor should not be used if openpilot is not controlling longitudinal
+  if (toyota_stock_longitudinal) {
+    enable_gas_interceptor = false;
+  }
+
   safety_config ret;
   if (toyota_lta) {
     ret = BUILD_SAFETY_CFG(toyota_lta_rx_checks, TOYOTA_TX_MSGS);
```
sshane, line 347:
```suggestion
```

board/safety/safety_toyota.h
```diff
@@ -304,13 +334,21 @@ static safety_config toyota_init(uint16_t param) {
...
+  // Gas interceptor should not be used if openpilot is not controlling longitudinal
+  if (toyota_stock_longitudinal) {
+    enable_gas_interceptor = false;
+  }
+
   safety_config ret;
   if (toyota_lta) {
-    ret = BUILD_SAFETY_CFG(toyota_lta_rx_checks, TOYOTA_TX_MSGS);
+    ret = enable_gas_interceptor ? BUILD_SAFETY_CFG(toyota_lta_interceptor_rx_checks, TOYOTA_INTERCEPTOR_TX_MSGS) : \
+                                   BUILD_SAFETY_CFG(toyota_lta_rx_checks, TOYOTA_TX_MSGS);
```
sshane, line 348: we can not allow LTA, I just kept it because openpilot allows it in any configuration

adeebshihadeh (approved):

PR #1749, opened 2023-11-28
remove GMLAN

adeebshihadeh:


ServerDummy:
lowspeed should not be removed, especially now that we know that the bolts and rear radars are available on it and most likely any other GM vehicle with BSM. Just dash lights are a nice to have, but BSM is an important safety feature. There could be much more on low speed such as blinker control and possibly auto resume support. Some of this stuff should be at least be looked into before killing it off.

https://github.com/commaai/opendbc/pull/962

PR #1750, opened 2023-11-29
safety: full coverage

sshane:
Would be nice to actually test the full tick logic, but need a custom safety mode where we know all the messages needed to send, and at the correct rates, to test the lag limits.

---

Impossible for dx to be <= 0. This never fails:

```python
for _ in range(1000000):
  xl = np.random.rand(3)
  x = np.random.rand(1)[0]
  if x <= xl[0]:
    pass
  else:
    for i in range(2):
      if x < xl[i + 1]:
        dx = xl[i+1] - xl[i]
        assert dx > 0
        break
```


tests/safety/common.py
```diff
@@ -805,6 +805,12 @@ def test_manually_enable_controls_allowed(self):
     self.safety.set_controls_allowed(0)
     self.assertFalse(self.safety.get_controls_allowed())
 
+  def test_safety_tick(self):
+    self.safety.set_timer(int(2e6))
+    self.safety.set_controls_allowed(True)
+    self.safety.safety_tick_current_safety_config()
+    self.assertFalse(self.safety.get_controls_allowed())
```
adeebshihadeh, line 812: add asset on rx checks invalid?

board/safety.h
```diff
@@ -503,10 +503,6 @@ float interpolate(struct lookup_t xy, float x) {
         float y0 = xy.y[i];
         float dx = xy.x[i+1] - x0;
         float dy = xy.y[i+1] - y0;
-        // dx should not be zero as xy.x is supposed to be monotonic
-        if (dx <= 0.) {
-          dx = 0.0001;
-        }
```
adeebshihadeh, line 509: replace with MIN?

PR #1758, opened 2023-12-05
One file coverage

sshane:


sshane:
superseded by https://github.com/commaai/panda/pull/1865

PR #1767, opened 2023-12-09
safety: fully reset dynamic rx check fields

sshane:
Having to type extra `.status` is annoying, but think it's the cleanest way to only reset those fields without touching `CanMsgCheck`. We can also use `memset` instead, but that is probably much more error prone


sshane:
This is also needed for https://github.com/commaai/openpilot/pull/30443, since the last counter and number of invalid counters weren't getting reset in the call to `init_tests`

adeebshihadeh (approved):

PR #1776, opened 2023-12-15
Ford: check predicted gas

sshane:
Unknown what effect this signal has if set while other signal is inactive, so it's best to assert safety on this. We do this for other safety modes too: https://github.com/commaai/panda/blob/bb75afc84e4d33af1554ada81851547db648e33e/board/safety/safety_hyundai.h#L242-L243 and https://github.com/commaai/panda/blob/bb75afc84e4d33af1554ada81851547db648e33e/board/safety/safety_hyundai_canfd.h#L272-L273

It just happened that a byte value of 0 is the actual inactive value (-5 m/s^2), and that it actuates gas.

### DIFF
--- a/board/safety/safety_ford.h
+++ b/board/safety/safety_ford.h
@@ -157,7 +157,7 @@ const LongitudinalLimits FORD_LONG_LIMITS = {
   .inactive_accel = 5128,  // -0.0008 m/s^2
 
   // gas cmd limits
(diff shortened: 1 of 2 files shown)


sheaduncan:
I disagree with this logic. Predicted Gas <> Required Gas

image

As you can see Predicted Gas aligns more with accel in this stock route. So, limiting it to -0.5 doesn't add any value.

PR #1778, opened 2023-12-15
Subaru: fix units for vehicle speed

jnewb1:
caught by https://github.com/commaai/openpilot/pull/30750


sshane:
adding the angle test also should catch this, can you check that it does?

jnewb1:
the panda check requires that _speed_msg has the correct units

PR #1780, opened 2023-12-16
Nissan: fix velocity mismatch with openpilot

jnewb1:
openpilot uses all 4 wheels to calculate speed, whereas panda only uses the rear wheels. this can cause a slight difference between the two.

need a better way of handling when a value we care about in panda is split across two separate messages


board/safety/safety_nissan.h
```diff
@@ -56,12 +60,18 @@ static void nissan_rx_hook(CANPacket_t *to_push) {
       update_sample(&angle_meas, angle_meas_new);
     }
 
+    if ((addr == 0x284) || (addr == 0x29a)) {
```
sshane, line 63: Make sure you add to rx checks

Issue #1770, opened 2023-12-09
add more MISRA mutation test cases


adeebshihadeh:
See https://github.com/commaai/panda/blob/master/tests/misra/test_mutation.py

adeebshihadeh:
$100 for testing all board/safety/* files 
$250 for adding coverage to 10+ new MISRA rules  (just need to add patterns that violate each rule)

With the new cases, the mutation stage can't significantly slow down and must finish in 2mins.

jimbrend:
where do I get all all board/safety/* files?

PR #1792, opened 2024-01-13
Misra safety Check

bongbui321:
For first part of [#1770](https://github.com/commaai/panda/issues/1770), 
add misra testing for safety files, enable `misra-c2012-10.4`

### DIFF
--- a/board/drivers/fdcan.h
+++ b/board/drivers/fdcan.h
@@ -112,7 +112,7 @@ void process_can(uint8_t can_number) {
           fifo = (canfd_fifo *)(TxFIFOSA + (tx_index * FDCAN_TX_FIFO_EL_SIZE));
 
           fifo->header[0] = (to_send.extended << 30) | ((to_send.extended != 0U) ? (to_send.addr) : (to_send.addr << 18));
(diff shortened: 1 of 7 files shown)


adeebshihadeh:
Is this still a draft? If not, I'm not sure what it's supposed to do. The test already checks the safety files. Let me know if you need clarification on the issue.

bongbui321:
Yeah, so right now one of the test in `./test_mutation.py` is failing - it doesn't show on GHA(not sure why), but right now I'm going to enable `misrac-10.4` so that that test would pass and then the first part (board/safety/* files) will be done

bongbui321:
@adeebshihadeh Okay it is ready for review now, enable misrac 10.4 and all the tests `./test_mutation.py` pass now

adeebshihadeh:
@bongbui321 looks like there's a few different things going on in this PR that should be their own PR
* editing test_misra.sh
* adding new mutation test cases
* enabling 10.4

I also formalized the bounty and wrote more instructions on enabling the rest of the MISRA rules: https://github.com/commaai/panda/issues/1794. There's probably a bunch of low hanging fruit in there, along with some tricky ones.

PR #1795, opened 2024-01-13
Misra check for safety files

bongbui321:
For first part of [#1770](https://github.com/commaai/panda/issues/1770)


adeebshihadeh:
I don't understand either of these changes:
* the sanity check exists so that we don't silently disable all or any of the coverage
* the board/safety files are already tested in F4 and H7 tests. F4 and H7 refer to the two different panda MCUs - we test the full source for each of them

bongbui321:
yeah, didn't read the main.c code that has safety.h in it before hand, will close now

Issue #1794, opened 2024-01-13
Finish MISRA coverage


adeebshihadeh:
In https://github.com/commaai/panda/pull/926, we updated cppcheck to the latest version which completed cppcheck's MISRA coverage. These new checks were suppressed in that PR in order to get the cppcheck version updated, and now we want to enable those checks and fix the violations.

Read about MISRA C:2012 [here](https://en.wikipedia.org/wiki/MISRA_C); the guidelines and information about them can be found searching online (ChatGPT also knows about it).

$50 for each rule addressed, either fixed all violations or suppressed for good reason. There's 28 new rules to enable, and progress can be checked [here](https://github.com/commaai/panda/blob/master/tests/misra/suppressions.txt).
* each rule should get its own PR
* PR title should be something like "enable misra-c2012-2.2"

Example for `misra-c2012-2.2`: f64ba24685fb3c71e37c720734d24d6b9d995617.

adeebshihadeh:
@0x41head @samrum @bongbui321 I missed these in my initial count, but the ones that don't have "misra" in the name also count:
```
unusedFunction
constParameterPointer
constParameterCallback
knownConditionTrueFalse
```

0x41head:
An update on `unusedFunction` :
```
Currently unusedFunction needs to see the whole code to avoid false positives making it unfeasible to enable that check on single files only. 
```
In the case of pandas, since we are only checking a couple of files, the cppchecker doesn't get the entire context of the codebase.
I confirmed this by running cpp checker over the entire codebase, where it gave no issues with respect to unusedFunction

[Reference](https://trac.cppcheck.net/ticket/10727)

adeebshihadeh:
> In the case of pandas, since we are only checking a couple of files, the cppchecker doesn't get the entire context of the codebase. I confirmed this by running cpp checker over the entire codebase, where it gave no issues with respect to unusedFunction
> 
> [Reference](https://trac.cppcheck.net/ticket/10727)

Want to open a PR to run on the whole repo? Any reason not to do this?

adeebshihadeh:
10.3 seems to have false positives for bools. Putting up an extra $150 (for $200 total on that rule) to fix it in cppcheck and make an upstream PR to cppcheck. We can potentially ship the patch here first if they don't merge it quickly.

https://sourceforge.net/p/cppcheck/discussion/development/thread/c96d6d44f1/

0x41head:
[...]

PR #1796, opened 2024-01-13
enable misra-c2012-10.4

bongbui321:
For [#1794](https://github.com/commaai/panda/issues/1794)

### DIFF
--- a/board/drivers/can_common.h
+++ b/board/drivers/can_common.h
@@ -11,8 +11,8 @@ typedef struct {
   int8_t forwarding_bus;
   uint32_t can_speed;
   uint32_t can_data_speed;
-  uint32_t canfd_enabled;
-  uint32_t brs_enabled;
+  bool canfd_enabled;
+  bool brs_enabled;
   bool canfd_non_iso;
 } bus_config_t;
 
--- a/board/drivers/can_common.h
+++ b/board/drivers/can_common.h
@@ -11,8 +11,8 @@ typedef struct {
   int8_t forwarding_bus;
   uint32_t can_speed;
   uint32_t can_data_speed;
-  bool canfd_enabled;
-  bool brs_enabled;
+  uint32_t canfd_enabled;
+  uint32_t brs_enabled;
   bool canfd_non_iso;
 } bus_config_t;
 
--- a/board/drivers/fdcan.h
+++ b/board/drivers/fdcan.h
@@ -112,7 +112,9 @@ void process_can(uint8_t can_number) {
           fifo = (canfd_fifo *)(TxFIFOSA + (tx_index * FDCAN_TX_FIFO_EL_SIZE));
 
           fifo->header[0] = (to_send.extended << 30) | ((to_send.extended != 0U) ? (to_send.addr) : (to_send.addr << 18));
-          fifo->header[1] = (to_send.data_len_code << 16) | (bus_config[can_number].canfd_enabled << 21) | (bus_config[can_number].brs_enabled << 20);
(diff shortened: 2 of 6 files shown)


board/drivers/harness.h
```diff
@@ -39,7 +39,7 @@ void set_intercept_relay(bool intercept, bool ignition_relay) {
     }
 
     // wait until we're not reading the analog voltages anymore
-    while (harness.sbu_adc_lock == true) {}
+    while ((bool)harness.sbu_adc_lock == true) {}
```
adeebshihadeh, line 42: `harness.sbu_adc_lock` is already a bool, I suspect it's complaining about the "== true" that you can just remove. it's really important to do these fixes in a principled way and to think very carefully about them

board/safety/safety_nissan.h
```diff
@@ -104,7 +104,7 @@ static bool nissan_tx_hook(CANPacket_t *to_send) {
     bool lka_active = (GET_BYTE(to_send, 6) >> 4) & 1U;
 
     // Factor is -0.01, offset is 1310. Flip to correct sign, but keep units in CAN scale
-    desired_angle = -desired_angle + (1310 * NISSAN_STEERING_LIMITS.angle_deg_to_can);
+    desired_angle = -desired_angle + ((float)1310 * NISSAN_STEERING_LIMITS.angle_deg_to_can);
```
adeebshihadeh, line 107:
```suggestion
    desired_angle = -desired_angle + (1310.0f * NISSAN_STEERING_LIMITS.angle_deg_to_can);
```

board/drivers/fdcan.h
```diff
@@ -112,7 +112,7 @@ void process_can(uint8_t can_number) {
           fifo = (canfd_fifo *)(TxFIFOSA + (tx_index * FDCAN_TX_FIFO_EL_SIZE));
 
           fifo->header[0] = (to_send.extended << 30) | ((to_send.extended != 0U) ? (to_send.addr) : (to_send.addr << 18));
-          fifo->header[1] = (to_send.data_len_code << 16) | (bus_config[can_number].canfd_enabled << 21) | (bus_config[can_number].brs_enabled << 20);
+          fifo->header[1] = (typeof(fifo->header[1]))(to_send.data_len_code << 16) | (bus_config[can_number].canfd_enabled << 21) | (bus_config[can_number].brs_enabled << 20);
```
adeebshihadeh, line 115: why does this line need a typeof and the others like it don't? I suspect there's a cleaner fix here

adeebshihadeh (changes requested):

board/drivers/fdcan.h
bongbui321: hmm... I changed the canfd_enabled to uint32_t since it is shifted and updated to be of int type anyway. I don't really like changing it to another type like this though, may be do inline suppress for this?

bongbui321: I think the way cppcheck catch this is looking at the type of canfd_enabled and data_len_code, and it doesn't take into account that bool type get shifted turn into an int.

board/drivers/can_common.h
```diff
@@ -11,8 +11,8 @@ typedef struct {
   int8_t forwarding_bus;
   uint32_t can_speed;
   uint32_t can_data_speed;
-  bool canfd_enabled;
-  bool brs_enabled;
+  uint32_t canfd_enabled;
+  uint32_t brs_enabled;
```
adeebshihadeh, line 15: I think it would be cleaner to just break the cases cppcheck is flagging into if statements and just or them, e.g.
```
if canfd_enabled:
  data |= (1 << 21)
```

PR #1797, opened 2024-01-14
Add 10 mutation tests 

bongbui321:
For [#1770](https://github.com/commaai/panda/issues/1770).  I removed the (None, None, False) case since it was checked in the step before misrac check already, and it slows down CI time

Just to make sure I understand the bounty, the added patterns have to replace the part of the code while maintaining the same functionality?

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -120,7 +120,7 @@ jobs:
         timeout-minutes: 1
         run: ${{ env.RUN }} "cd tests/misra && ./test_misra.sh"
       - name: MISRA mutation tests
-        timeout-minutes: 2
+        timeout-minutes: 3
         run: ${{ env.RUN }} "cd tests/misra && ./test_mutation.py"
 
   python_linter:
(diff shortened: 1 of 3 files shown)


adeebshihadeh:
> For [#1770](https://github.com/commaai/panda/issues/1770). I removed the (None, None, False) case since it was checked in the step before misrac check already, and it slows down CI time

We want to keep this. It's a better test this way, otherwise we can easily write a bug where the test always fails but for reasons other than the violations we setup. Checking that it still passes with no patches catches that bug.

> Just to make sure I understand the bounty, the added patterns have to replace the part of the code while maintaining the same functionality?

Not necessarily. Ideally what we want here are generic patterns (not specific to any file) that we can then mix and match with files.

bongbui321:
@adeebshihadeh was wondering if I can use caching of scons and misra mutation test for the 2 minute limit or do you want it  compile and cppcheck under 2 minutes sratch. It seems kind of impossible to me since the (none, none, false) case already take like 1m on the github runners

PR #1798, opened 2024-01-14
Enable caching in test_misra

bongbui321:
Before it was not caching into `--cppcheck-build-dir` because each folder can only used for one file and we have 3 files to check. After caching if there is no change in compilation then it would take ~1s, and only recache specifically changed files. I disabled the sanity check to only check if the folder doesn't already exists because checkers are not called if the files are cached.
```
** PANDA F4 CODE **
Checking board/main.c ...
nofile:0:0: information: Active checkers: 177/792 (use --checkers-report=<filename> to see details) [checkersReport]

** PANDA H7 CODE **
Checking board/main.c ...
nofile:0:0: information: Active checkers: 177/792 (use --checkers-report=<filename> to see details) [checkersReport]

** PEDAL CODE **
Checking board/pedal/main.c ...
nofile:0:0: information: Active checkers: 177/792 (use --checkers-report=<filename> to see details) [checkersReport]

Success! took 1 seconds
```
If this PR is merge, we put this cache into GHA similar to scons_cache

### DIFF
--- a/tests/misra/test_misra.sh
+++ b/tests/misra/test_misra.sh
@@ -27,7 +27,8 @@ cd $PANDA_DIR
 scons -j8
 
 cppcheck() {
-  build_dir=/tmp/cppcheck_build_$7
+  hashed_args=$(echo -n "$@" | sha256sum | awk '{print $1}')
+  build_dir=/tmp/cppcheck_build/$hashed_args
   if [ -d $build_dir ]; then
     build_dir_exists=true
   else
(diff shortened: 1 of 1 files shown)


tests/misra/test_misra.sh
```diff
@@ -39,26 +44,28 @@ cppcheck() {
...
-  yes="$(grep '^Yes' $report | wc -l)"
-  echo "$yes checks enabled, $no disabled"
-  if [[ $yes -lt 250 ]]; then
-    echo "Count of enabled checks seems too low."
-    exit 1
-  fi
-  if [[ $no -ne 99 ]]; then
-    echo "Disabled check count threshold doesn't match, update to $no"
-    exit 1
+  if [ $build_dir_exists == false ]; then
```
adeebshihadeh, line 47: is there some other way to do this sanity check? if not, i'd consider just removing it. you always want your test to run the same every time

tests/misra/test_misra.sh
```diff
@@ -27,8 +27,13 @@ cd $PANDA_DIR
 scons -j8
 
 cppcheck() {
-  build_dir=/tmp/cppcheck_build
-  mkdir -p $build_dir
+  build_dir=/tmp/cppcheck_build_$7
```
adeebshihadeh, line 30: let's make a common root for them
```suggestion
  build_dir=/tmp/cppcheck_build/$7
```

tests/misra/test_misra.sh
```diff
@@ -39,26 +44,28 @@ cppcheck() {
...
+    if [[ $no -ne 99 ]]; then
+      echo "Disabled check count threshold doesn't match, update to $no"
+      exit 1
+    fi
   fi
 }
 
 printf "\n${GREEN}** PANDA F4 CODE **${NC}\n"
-cppcheck -DCAN3 -DPANDA -DSTM32F4 -UPEDAL -DUID_BASE board/main.c
+cppcheck -DCAN3 -DPANDA -DSTM32F4 -UPEDAL -DUID_BASE board/main.c f4
```
adeebshihadeh, line 63: can we just hash the args instead of passing in a name?

adeebshihadeh (changes requested):

adeebshihadeh:
Good catch! This should be a good speedup.

tests/misra/test_misra.sh
bongbui321: Hmm in my latest commit I removed it and the report - can restore if you see it necessary. As I see it we already have `--enable=all` and `--force` enabled which forces every necessary checkers to wake up already, so it should be safe. As I see it, checkers are not used for cached files unless the content change, so I can't think of any native way to ensure the same number of checkers everytime with cache enabled

adeebshihadeh:
I see there's no speedup in the mutation tests. Perhaps we want to also copy the cache for each test case?

bongbui321:
@adeebshihadeh  Yeah, that would speed it up. However, I think the reason it is not caching for test_muation is because we are using tempfile which is different everytime which makes the path different everytime.

PR #1800, opened 2024-01-15
enable misra-c2012-5.8

0x41head:
for #1794 
lmk if I missed anything


adeebshihadeh:
Nope, didn't miss anything. This is exactly what we're looking for for these MISRA fixes.

PR #1801, opened 2024-01-15
enable misra-c2012-2.3 & misra-c2012-2.4

bongbui321:
For [#1794](https://github.com/commaai/panda/issues/1794), I included both the 2.3 and 2.4 since they are the same fix. 

1. Inline suppress the CANPacket_t type because it is used in other files and it is a false positives of cppcheck to check only if it is used in that specific file - `board/can_definitions.h`. 
2. Commented out `USB_OTG_HostPortTypeDef` type because they are commented out in `board/drivers/usb.h` and not used everywhere else


board/stm32fx/llusb.h
```diff
@@ -1,13 +1,13 @@
-typedef struct
-{
-  __IO uint32_t HPRT;
-}
-USB_OTG_HostPortTypeDef;
+//typedef struct
+//{
+//  __IO uint32_t HPRT;
+//}
+//USB_OTG_HostPortTypeDef;
```
adeebshihadeh, line 5: just delete the commented out lines

board/can_definitions.h
```diff
@@ -15,7 +15,8 @@ const uint8_t PANDA_BUS_CNT = 4U;
   #define CANPACKET_DATA_SIZE_MAX 8U
 #endif
 
-typedef struct {
+// cppcheck-suppress [misra-c2012-2.3, misra-c2012-2.4]
```
adeebshihadeh, line 18: hmm, why do we have to do this here but not for health.h?

bongbui321: yeah it was a bit tricky to understand for me in the beginning too. 

if you try to run my code, cppcheck is flagging misra-c2012 2.3 at the `board/pedal/main.c` file only not the other two test of `board/main.c`. I'm not entirely sure of the implementation of cppcheck, but I read through the code base and see that they would read through all of the include header files. For instance to check for misra_2.3 it would look through all of the `typedef struct `definition of the main file as well as the included files. In the `board/pedal/main.c`, if you follow the trail then it would include some header files that include the `can_definitions.h` file, while the pedal/main.c file doesn't use any header file that has `health.h` file in it. 

I also did one experiment and put `#include health.h` in the `board/pedal/main.c` file and get this. 
```
/home/openpilot/panda/board/health.h:35:1: style: misra violation (use --rule-texts=<file> to get proper output) [misra-c2012-2.3]
typedef struct __attribute__((packed)) {
^
/home/openpilot/panda/board/health.h:4:41: style: misra violation (use --rule-texts=<file> to get proper output) [misra-c2012-2.4]
struct __attribute__((packed)) health_t {
```

So I think it is a false positive of cppcheck.

adeebshihadeh: it's not necessary, I just reverted it. please be more careful with changes like this in the future

PR #1802, opened 2024-01-16
enable misra-c2012-5.6

samrum:


tests/misra/test_misra.sh
```diff
@@ -40,12 +40,12 @@ cppcheck() {
 }
 
 printf "\n${GREEN}** PANDA F4 CODE **${NC}\n"
-cppcheck -DCAN3 -DPANDA -DSTM32F4 -UPEDAL -DUID_BASE board/main.c
+cppcheck -DCAN3 -DPANDA -DSTM32F4 -UPEDAL -DUID_BASE $PANDA_DIR/board/main.c
```
adeebshihadeh, line 43: why does this fix it?

samrum: 5.6 requires typedefs to be unique and it seems like since the cppcheck command sets `-I $PANDA_DIR/board/` and this uses a relative path for `board/main.c` cppcheck considers it a separate file that duplicates typedefs:
```bash
/home/samrum/dev/ commaai/panda/board/can_definitions.h:28:39: style: misra violation (use --rule-texts=‹file› to get proper output)
[misra-c2012-5.6]
_attribute__((packed, aligned(4))) CANPacket_t;

board/can definitions.h:28:39: style: misra violation (use --rule-texts=‹file› to get proper output) [misra-c2012-5.6]
} _attribute_
_((packed, aligned(4))) CANPacket_t;
```

So this actually fixes similar rules like c2012-8.6: An identifier with external linkage shall have exactly one external definition.
Dunno if you'd like additional PRs to remove those suppressions or just to remove them here.

adeebshihadeh: you can just include them here

PR #1811, opened 2024-01-16
enable misra-c2012-10.6

0x41head:
for #1794


board/safety/safety_chrysler.h
```diff
@@ -183,7 +183,7 @@ static void chrysler_rx_hook(CANPacket_t *to_push) {
   }
 
   // enter controls on rising edge of ACC, exit controls on ACC off
-  const int das_3_bus = (chrysler_platform == CHRYSLER_PACIFICA) ? 0 : 2;
+  const int das_3_bus = (chrysler_platform == CHRYSLER_PACIFICA) ? (int) 0 : (int) 2;
```
adeebshihadeh, line 186: what's with all the casting to int? we shouldn't have to do this

PR #1813, opened 2024-01-16
enable misra-c2012-2.7 

0x41head:
for #1794

### DIFF
--- a/board/stm32h7/lli2c.h
+++ b/board/stm32h7/lli2c.h
@@ -119,32 +119,29 @@ end:
   return ret;
 }
 
-// cppcheck-suppress misra-c2012-2.7; not sure why it triggers here?
-bool i2c_set_reg_bits(I2C_TypeDef *I2C, uint8_t addr, uint8_t reg, uint8_t bits) {
+bool i2c_set_reg_bits(I2C_TypeDef *I2C, uint8_t address, uint8_t regis, uint8_t bits) {
   uint8_t value;
-  bool ret = i2c_read_reg(I2C, addr, reg, &value);
+  bool ret = i2c_read_reg(I2C, address, regis, &value);
   if(ret) {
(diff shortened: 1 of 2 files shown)


0x41head:
the 2.7 check seems to be broken:
```
def misra_2_7(self, data):
        for func in data.functions:
            # Skip function with no parameter
            if len(func.argument) == 0:
                continue
            # Setup list of function parameters
            func_param_list = list()
            for arg in func.argument:
                func_arg = func.argument[arg]
                if func_arg.typeStartToken and func_arg.typeStartToken.str == '...':
                    continue
                func_param_list.append(func_arg)
            # Search for scope of current function
            for scope in data.scopes:
                if (scope.type == "Function") and (scope.function == func):
                    # Search function body: remove referenced function parameter from list
                    token = scope.bodyStart
                    while token.next is not None and token != scope.bodyEnd and len(func_param_list) > 0:
                        if token.variable is not None and token.variable in func_param_list:
                            func_param_list.remove(token.variable)
                        token = token.next
                    # Emit a warning for each unused variable, but no more that one warning per line
                    reported_linenrs = set()
                    for func_param in func_param_list:
                        if func_param.nameToken:
                            linenr = func_param.nameToken
                            if linenr not in reported_linenrs:
                                self.reportError(func_param.nameToken, 2, 7)
                                reported_linenrs.add(linenr)
                        else:
                            linenr = func.tokenDef.linenr
                            if linenr not in reported_linenrs:
                                self.reportError(func.tokenDef, 2, 7)
                                reported_linenrs.add(linenr)
```
My guess is `reg` and `addr` were being considered as data types and not data variables that could be represented by `token.variable`? I would need to understand how to debug the misra addon to provide a conclusive answer.

adeebshihadeh:
Nice job tracking this down, probably worth opening an issue with cppcheck.

PR #1814, opened 2024-01-16
Honda Radarless: only forward buttons frame when controls are off.

csouers:
Related to: https://github.com/commaai/openpilot/pull/31022

**Changes**:

- Forward SCM_BUTTONS frame on Honda Bosch Radarless w/ stock ACC when !controls_allowed
- Add on to button spam test confirming tx of the untested cruise buttons
- Test the forwarding block of SCM_BUTTONS when controls are on.


jyoung8607:
@csouers I'm looking at this one, as a prereq to some other PRs that @vanillagorillaa had open. I see in the openpilot PR history that cancel/resume spamming wasn't reliable without this change.

Is there any reason why we couldn't always block at the Panda safety layer, and always proxy at the openpilot layer, rewriting in openpilot as needs dictate? Your current approach could be made to work in the usual case, but is a little more complicated than what seems necessary, and might break in off-nominal situations.

Consider the case when Panda could transition to controls-allowed, but openpilot rejects the engagement. This can happen if you hit Set while openpilot is calibrating, or a seatbelt isn't fastened, or any other `noEntry` event is present.

We can still apply safety logic to the transmitted messages.

csouers:
> Is there any reason why we couldn't always block at the Panda safety layer, and always proxy at the openpilot layer

Nope! Just an attempt to retain as much of a pcm_enable policy as possible. Subaru pre-global does a full proxy anyway. I'll rework this.

sshane:

PR #1815, opened 2024-01-16
enable misra-c2012-14.2

adeebshihadeh:


board/main.c
```diff
@@ -385,8 +385,8 @@ int main(void) {
 
   // LED should keep on blinking all the time
   uint64_t cnt = 0;
-
-  for (cnt=0;;cnt++) {
+  while (true) {
+    cnt++;
```
devtekve, line 389: Does the LED depend on `cnt` increasing? or is there any other reason to keep it around? while the max is quite high, I think having the risk of overflowing is enough to remove it if it's not being actually used until it is needed, then we can rethink our strategy on it if needed...

PR #1817, opened 2024-01-17
CI: Enable MISRA tests caching

bongbui321:
Enable caching both the scons build and the MISRA test (test_misra.sh and test_mutation.py), reducing the MISRA CI test from ~2m20 normally to [~40s](https://github.com/bongbui321/panda/actions/runs/7551529023/job/20558894429). This should be a good speedup for MISRA mutation test cases


adeebshihadeh:
This really only helps in the case where we're not touching the firmware, and most commits to panda touch the firmware. I don't think this is worth the extra complexity.

PR #1818, opened 2024-01-18
Add more MISRA mutation tests

bongbui321:
Randomly patch pattern now. with CI env variable saves about ~30s. I did 4 tests: 

- 2 tests without CI env var: [2m46s](https://github.com/commaai/panda/actions/runs/7592133440/job/20681117997), [2m54s](https://github.com/commaai/panda/actions/runs/7592226200/job/20681323218)
- 2 tests with CI env var: [2m11s](https://github.com/commaai/panda/actions/runs/7592190342/job/20681241177), [2m8s](https://github.com/commaai/panda/actions/runs/7592261430/job/20681406197)

Need pytest-randomly for pytest with random obj possible ([similar issue](https://github.com/pytest-dev/pytest-xdist/issues/432)). Need to update the ghcr

### DIFF
--- a/tests/misra/test_mutation.py
+++ b/tests/misra/test_mutation.py
@@ -16,8 +16,13 @@ IGNORED_PATHS = (
   'board/jungle',
   'board/stm32h7/inc',
   'board/stm32fx/inc',
+  'board/fake_stm.h',
+
+  # bootstub only files
+  'board/flasher.h',
   'board/bootstub.c',
   'board/bootstub_declarations.h',
+  'board/stm32fx/llflash.h'
 )
 
 mutations = [
--- a/tests/misra/test_mutation.py
+++ b/tests/misra/test_mutation.py
@@ -10,39 +10,36 @@ import random
 HERE = os.path.abspath(os.path.dirname(__file__))
 ROOT = os.path.join(HERE, "../../")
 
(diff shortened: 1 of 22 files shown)


tests/misra/test_mutation.py
```diff
@@ -13,27 +13,50 @@
...
 # - come up with a pattern for each rule (cppcheck tests probably have good ones?)
 mutations = [
+  # default
   (None, None, False),
-  # F4 only
-  ("board/stm32fx/llbxcan.h", "s/1U/1/g", True),
-  # H7 only
-  ("board/stm32h7/llfdcan.h", "s/return ret;/if (true) { return ret; } else { return false; }/g", True),
-  # general safety
-  ("board/safety/safety_toyota.h", "s/is_lkas_msg =.*;/is_lkas_msg = addr == 1 || addr == 2;/g", True),
```
adeebshihadeh, line 22: let's keep these (I won't count it against the 2m). it's very easy to see that we cover at least those three cases by having this

adeebshihadeh (commented):
Let's combine the two listed bounties in that issue and do this for the full $350:
* move all the patterns you added to a pattern list
* use the list of patterns to generate test cases by pairing each one with a randomly chosen file

PR #1819, opened 2024-01-18
enable misra-c2012-21.2

0x41head:
For #1794 

Due to numerous implicit calls, I was unable to rename `memset` and `memcpy`.
Changing their names in `flasher.h` breaks the tests.
Suppressed them for now.


adeebshihadeh:
Suppression is correct for the libc.h functions, since this is our own implementation of them in lieu of including libc.

PR #1823, opened 2024-01-19
enable misra-c2012-21.15

0x41head:
For #1794 
Not sure if there are negative side effects with these changes


maxime-desroches:
Fixed with https://github.com/commaai/panda/pull/2030

PR #1829, opened 2024-01-21
enable misra-c2012-8.2

samrum:
This is a bit of a workaround for what seems like a bug with cppcheck dumps dropping rx_msg_safety_check def arguments  after the first const function pointer type.


adeebshihadeh:
Ah, this is a nice change regardless.

PR #1838, opened 2024-01-26
Removed 'goto' statements as per rule 8.4

KrishnaTeja2121:
Kindly review my PR.


board/bootstub.c
```diff
@@ -19,11 +19,13 @@
 #include "obj/gitversion.h"
 #include "flasher.h"
 
-void __initialize_hardware_early(void) {
+void __initialize_hardware_early(void)
+{
```
adeebshihadeh, line 23: revert all changes that are irrelevant too the MISRA rule or for style

adeebshihadeh (changes requested):

PR #1845, opened 2024-02-03
Hardware connections cleanup

dzid26:
0 is CAN1 as noted here:
https://github.com/commaai/panda/blob/457e3b262d798aa6e400033c92d12a0b0f52a7ed/board/drivers/can_common.h#L169-L170


adeebshihadeh:
Two issues with this PR: it has no clear goal and it's way too big. If you'd like to get these changes in, break them into smaller PRs, each with a clear goal.

PR #1856, opened 2024-02-05
Hardware pin names refactor

dzid26:
1. Why I changed "sbu relay" to "switch"? 
SBUs lines are driving the relay, yes, but it is more logical to use "switch" since each SBU has its own transistor activated depending on cable rotation. Yes, transistor can be called relay, but it's better to say "switch" which separates it from the mechanical relay in the harness.

3. Why I changed "Ignition relay" to "Ignition test probe"?
Because by the same logic as with "harness relay", this implied there was a second relay (or a transistor) that can be driven. But there isn't anything that can be switched.  All this does, is pulls harness sbu2 line voltage higher.

### DIFF
--- a/board/boards/black.h
+++ b/board/boards/black.h
@@ -107,8 +107,8 @@ void black_init(void) {
   set_gpio_output(GPIOC, 5, 0);
   set_gpio_output(GPIOC, 12, 0);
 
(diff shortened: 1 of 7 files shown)


adeebshihadeh:
Closing since this name is also on a bunch of schematics. It'll be more confusing (at least to us) to have a different name here.

PR #1852, opened 2024-02-05
enable misra-c2012-10.3

0x41head:
For #1794

Issue #1847, opened 2024-02-03
MISRSA results different in CI and workstation


adeebshihadeh:
Running `test_misra.sh` on my workstation, I get no errors, but CI reports a few style violations and doesn't fail. We need to figure out why the results are different and why CI doesn't fail even though errors are reported. Also need to fix these violations.

https://github.com/commaai/panda/actions/runs/7770060952/job/21189783553

```
** PANDA F4 CODE **
Checking /tmp/openpilot/panda/board/main.c ...
Checking /tmp/openpilot/panda/board/main.c: CAN3=1;PANDA=1;STM32F4=1;UID_BASE=1;STM32F423xx...
/tmp/openpilot/panda/board/health.h:24:9: style: misra violation (use --rule-texts=<file> to get proper output) [misra-c2012-5.8]
  float interrupt_load;
        ^
/tmp/openpilot/panda/board/drivers/interrupts.h:35:7: style: misra violation (use --rule-texts=<file> to get proper output) [misra-c2012-5.8]
float interrupt_load = 0.0f;
      ^
/tmp/openpilot/panda/board/health.h:27:12: style: misra violation (use --rule-texts=<file> to get proper output) [misra-c2012-5.8]
  uint16_t spi_checksum_error_count;
           ^
/tmp/openpilot/panda/board/drivers/spi.h:43:10: style: misra violation (use --rule-texts=<file> to get proper output) [misra-c2012-5.8]
uint16_t spi_checksum_error_count = 0;
         ^
/tmp/openpilot/panda/board/main_comms.h:71:58: style: misra violation (use --rule-texts=<file> to get proper output) [misra-c2012-5.8]
int comms_control_handler(ControlPacket_t *req, uint8_t *resp) {
                                                         ^
/tmp/openpilot/panda/board/drivers/usb.h:82:9: style: misra violation (use --rule-texts=<file> to get proper output) [misra-c2012-5.8]
uint8_t resp[USBPACKET_MAX_SIZE];
        ^
/tmp/openpilot/panda/board/health.h:26:11: style: misra violation (use --rule-texts=<file> to get proper output) [misra-c2012-5.8]
  uint8_t safety_rx_checks_invalid;
          ^
/tmp/openpilot/panda/board/safety_declarations.h:227:6: style: misra violation (use --rule-texts=<file> to get proper output) [misra-c2012-5.8]
bool safety_rx_checks_invalid = false;
     ^
[...]

PR #1859, opened 2024-02-07
update cppcheck 

0x41head:
Updates cppcheck version to remove false positive for misra 10.3
For  #1852

### DIFF
--- a/board/can_comms.h
+++ b/board/can_comms.h
@@ -68,7 +68,7 @@ void comms_can_write(const uint8_t *data, uint32_t len) {
       can_write_buffer.ptr += can_write_buffer.tail_size;
       pos += can_write_buffer.tail_size;
 
-      void *addr= &to_push;
+      void *addr = &to_push;
       // send out
       (void)memcpy(addr, can_write_buffer.data, can_write_buffer.ptr);
       can_send(&to_push, to_push.bus, false);
--- a/board/can_comms.h
+++ b/board/can_comms.h
@@ -68,9 +68,8 @@ void comms_can_write(const uint8_t *data, uint32_t len) {
       can_write_buffer.ptr += can_write_buffer.tail_size;
       pos += can_write_buffer.tail_size;
 
-      void *addr = &to_push;
       // send out
-      (void)memcpy(addr, can_write_buffer.data, can_write_buffer.ptr);
+      (void)memcpy(&to_push, can_write_buffer.data, can_write_buffer.ptr);
       can_send(&to_push, to_push.bus, false);
 
       // reset overflow buffer
--- a/board/can_comms.h
+++ b/board/can_comms.h
@@ -68,8 +68,9 @@ void comms_can_write(const uint8_t *data, uint32_t len) {
       can_write_buffer.ptr += can_write_buffer.tail_size;
       pos += can_write_buffer.tail_size;
 
+      void *addr= &to_push;
       // send out
-      (void)memcpy(&to_push, can_write_buffer.data, can_write_buffer.ptr);
+      (void)memcpy(addr, can_write_buffer.data, can_write_buffer.ptr);
       can_send(&to_push, to_push.bus, false);
 
       // reset overflow buffer
--- a/board/can_definitions.h
+++ b/board/can_definitions.h
@@ -15,7 +15,7 @@ const uint8_t PANDA_BUS_CNT = 4U;
   #define CANPACKET_DATA_SIZE_MAX 8U
 #endif
 
-// cppcheck-suppress [misra-c2012-2.3,misra-c2012-2.4] ; False Positive
+#ifndef STM32F2
 typedef struct {
   unsigned char reserved : 1;
   unsigned char bus : 3;
@@ -27,6 +27,7 @@ typedef struct {
   unsigned char checksum;
   unsigned char data[CANPACKET_DATA_SIZE_MAX];
 } __attribute__((packed, aligned(4))) CANPacket_t;
+#endif
 
 const unsigned char dlc_to_len[] = {0U, 1U, 2U, 3U, 4U, 5U, 6U, 7U, 8U, 12U, 16U, 20U, 24U, 32U, 48U, 64U};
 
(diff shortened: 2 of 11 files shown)


tests/misra/install.sh
```diff
@@ -9,8 +9,8 @@ if [ ! -d "$CPPCHECK_DIR" ]; then
 fi
 
 cd $CPPCHECK_DIR
-VERS="2.13.0"
-git fetch --tags origin $VERS
+VERS="f6b538e855f0bacea33c4074664628024ef39dc6"
+# git fetch --tags origin $VERS
```
adeebshihadeh, line 13:
```suggestion
git fetch --all
```

adeebshihadeh:
It seems this may solve part of #1847.

CI seems to fail now on those errors, but we still need to understand why the results are different on different machines.

board/main_comms.h
```diff
@@ -28,14 +28,14 @@ int get_health_pkt(void *dat) {
   health->alternative_experience_pkt = alternative_experience;
   health->power_save_enabled_pkt = power_save_status == POWER_SAVE_STATUS_ENABLED;
   health->heartbeat_lost_pkt = heartbeat_lost;
-  health->safety_rx_checks_invalid = safety_rx_checks_invalid;
+  health->safety_rx_chks_invalid = safety_rx_checks_invalid;
 
-  health->spi_checksum_error_count = spi_checksum_error_count;
+  health->spi_checksum_error_count = spi_checksum_err_count;
```
adeebshihadeh, line 33: let's just add `_pkt` like the rest of the fields

board/drivers/interrupts.h
```diff
@@ -32,7 +32,7 @@ uint8_t interrupt_depth = 0U;
 uint32_t last_time = 0U;
 uint32_t idle_time = 0U;
 uint32_t busy_time = 0U;
-float interrupt_load = 0.0f;
+float intr_load = 0.0f;
```
adeebshihadeh, line 35: let's revert this one and add `_pkt` to the other one

tests/misra/install.sh
```diff
@@ -9,8 +9,8 @@ if [ ! -d "$CPPCHECK_DIR" ]; then
 fi
 
 cd $CPPCHECK_DIR
-VERS="2.13.0"
-git fetch --tags origin $VERS
+VERS="f6b538e855f0bacea33c4074664628024ef39dc6"
```
adeebshihadeh, line 12:
```suggestion
# 2.13.0 plus some fixes
VERS="f6b538e855f0bacea33c4074664628024ef39dc6"
```

board/drivers/spi.h
```diff
@@ -40,7 +40,7 @@ uint8_t spi_state = SPI_STATE_HEADER;
 uint8_t spi_endpoint;
 uint16_t spi_data_len_mosi;
 uint16_t spi_data_len_miso;
-uint16_t spi_checksum_error_count = 0;
+uint16_t spi_checksum_err_count = 0;
```
adeebshihadeh, line 43: same here, add _pkt to the other one

adeebshihadeh (changes requested):

board/can_comms.h
```diff
@@ -68,8 +68,9 @@ void comms_can_write(const uint8_t *data, uint32_t len) {
       can_write_buffer.ptr += can_write_buffer.tail_size;
       pos += can_write_buffer.tail_size;
 
+      void *addr = &to_push;
```
adeebshihadeh, line 71: this looks like a workaround, rather than a real fix. it's also possible we may need to suppress something for memcpy

board/can_definitions.h
```diff
@@ -15,6 +15,7 @@ const uint8_t PANDA_BUS_CNT = 4U;
   #define CANPACKET_DATA_SIZE_MAX 8U
 #endif
 
+// cppcheck-suppress [misra-c2012-2.3,misra-c2012-2.4] ; False Positive
```
adeebshihadeh, line 18: can you make an issue for this on cppcheck's issue tracker?

board/can_comms.h
0x41head: It is indeed a workaround. This fixes the error that happens because we pass an uninitialized variable to a function. My justification for this change was that this was an oddity rather than a regular occurrence for the memcpy function. If you believe that we may need to call uninitialized variables from memcpy, we can suppress that check on the function.

[Reference](https://trac.cppcheck.net/ticket/1203)

board/can_definitions.h
0x41head: After further introspection I realized it wasn't a false positive. My apologies for the trouble.

board/can_comms.h
0x41head: After further introspection and digging deeper into the cppchecker code and issues, this seems to be a false positive caused by the return statement in the memcpy function.
An [issue](https://trac.cppcheck.net/ticket/12414#no1) has already been created for this in the cppchecker's issue tracker.

We seem to have three options here:
- suppress the check for memcpy function
-  change the return statement for memcpy to a const value
- fix the false positive in cppchecker

@adeebshihadeh let me know how to move forward with this.

adeebshihadeh: hmm, seems like this is a regression in cppcheck and we should just stick to the releases. I propose we go back to the 2.13 release and just `git cherry-pick` your fix onto it

board/health.h
```diff
@@ -21,10 +21,10 @@ struct __attribute__((packed)) health_t {
   uint8_t power_save_enabled_pkt;
   uint8_t heartbeat_lost_pkt;
   uint16_t alternative_experience_pkt;
-  float interrupt_load;
+  float interrupt_load_pkt;
   uint8_t fan_power;
-  uint8_t safety_rx_checks_invalid;
-  uint16_t spi_checksum_error_count;
+  uint8_t safety_rx_checks_invalid_pkt;
+  uint16_t spi_checksum_error_count_pkt;
```
sshane, line 27: @0x41head make sure to grep openpilot when making changes, these were used in boardd

PR #1869, opened 2024-02-15
cuatro power readout

robbederks:


board/stm32h7/lladc.h
```diff
@@ -18,17 +15,19 @@ void adc_init(void) {
 }
 
 uint16_t adc_get_raw(uint8_t channel) {
+  uint16_t res = 0U;
   ADC1->SQR1 &= ~(ADC_SQR1_L);
   ADC1->SQR1 = ((uint32_t) channel << 6U);
 
   ADC1->SMPR1 = (0x2U << (channel * 3U));
+  // cppcheck-suppress misra-c2012-12.2; don't know how to fix this. if statement to check channel range does not work
```
adeebshihadeh, line 23: suppressions should only be used when absolutely necessary. see https://github.com/commaai/panda/commit/39671c3dd663b9041564ec577c01ebd2009786df

robbederks: wait how does that fix it? tried a bunch of stuff. also don't know why it's triggering now since that code didn't change

adeebshihadeh: max value is 8, from here https://github.com/commaai/panda/blob/39671c3dd663b9041564ec577c01ebd2009786df/board/boards/cuatro.h#L53

PR #1876, opened 2024-02-17
MISRA: only true/false allowed for bools

AlexandreSato:
0 -> false


adeebshihadeh:
I don't really want to merge this without cppcheck enforcing this. @0x41head do you think there's a way to make cppcheck check this somehow? I'd put up a new bounty for adding upstream support for that.

0x41head:
Can I get some context ? What is this commit for ? Why is this commit a problem since false in stdbool is just `#define false 0`

adeebshihadeh:
We want `bool tx = 0/1` to fail and `bool tx = false/true` to pass for readability, even if 0/1 are the underlying values.

0x41head:
I am not sure if we can push this upstream. But we can definitely make our own addon for this check. Should be easy as all addons work on a dump file generated by cppchecker, which is just an XML.

adeebshihadeh:
I'd prefer to upstream it, but that may be an option if the addon is small (<20 lines?)

0x41head:
addon would be pretty small. Probably something like the [findcasts](https://github.com/danmar/cppcheck/blob/main/addons/findcasts.py) addon

adeebshihadeh:
That would be totally fine, but let's try to upstream it first.

adeebshihadeh:
Cool, let's do a $150 bounty for getting all of these fixed and checked with cppcheck, ideally inside cppcheck, but ok as an addon if they reject it

simrit1 (commented):
ok

PR #1877, opened 2024-02-18
refactor GET_BIT to be MISRA friendly

jyoung8607:
Refactor GET_BIT to return a `bool`, which makes it easier to use while maintaining MISRA compliance.

Cast directly from an int to a bool doesn't work, but old-school double-negation does the job and seems to be permitted. It did leave a [single breakage in the Chrysler port](https://github.com/commaai/panda/pull/1877/commits/a2e87257a8a220fd0bca707957c34e1759b479d3). While MISRA 10.4 does make specific mention of matching essential types on both sides of a ternary op, it's not clear why it didn't think both sides were the same here. An additional cast to bool worked, so there may be a subtlety between Boolean and _essentially_ Boolean types, or perhaps a bug in the type checker.


adeebshihadeh:
Great improvement!

PR #1879, opened 2024-02-18
enable misra-config checks

adeebshihadeh:


dzid26:
Did you notice `-DSTM32F413xx` option triggered more rule violations. (10.1, 10.4, 11.8, 12.2 ) 

Shouldn't `-DSTM32F413xx` option be mainlined and each rule tackled separately again?

adeebshihadeh:
> Did you notice `-DSTM32F413xx` & `-DSTM32H725xx` option triggered more rule violations. (10.1, 10.4, 11.8, 12.2 ) [rules.txt](https://github.com/commaai/panda/files/14326526/rules.txt) Shouldn't `-DSTM32F413xx` & `-DSTM32H725xx` options be mainlined and each rule tackled separately again?

I'm including those as part of enabling this.

PR #1888, opened 2024-02-24
UDS: add some basic type hints

incognitojam:


python/uds.py
```diff
@@ -553,7 +553,7 @@ def _isotp_rx_next(self, rx_data: bytes) -> ISOTP_FRAME_TYPE:
 FUNCTIONAL_ADDRS = [0x7DF, 0x18DB33F1]
 
 
-def get_rx_addr_for_tx_addr(tx_addr, rx_offset=0x8):
+def get_rx_addr_for_tx_addr(tx_addr: int, rx_offset: int = 0x8) -> Optional[int]:
```
incognitojam, line 556: oh, this might actually be more annoying than useful in openpilot (having to check for None everywhere)...

incognitojam: There isn't a good way to restrict tx_addr without throwing an error but I'm not sure whether that would break anything

PR #1889, opened 2024-02-24
use pyupgrade to update syntax

incognitojam:
pyupgrade automatically upgrades syntax for newer versions of python. I ran the tool with `--py311-plus` ~~and setup a pre-commit hook (which seems fast)~~. Also used `ruff --fix` to remove the unecessary imports.

~~I had to rename the `list` method of Panda and PandaDFU to `list_all`, because it clashes with using `list` as a type now.~~

Steps:
- Run `pyupgrade --py311-plus`
- Run `ruff --fix`
- Moved `PandaDFU.list()` method to end of file

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -18,3 +18,8 @@ repos:
   rev: v0.2.2
   hooks:
     - id: ruff
+- repo: https://github.com/asottile/pyupgrade
+  rev: v3.15.1
+  hooks:
+    - id: pyupgrade
+      args: [--py311-plus]
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -18,8 +18,3 @@ repos:
   rev: v0.2.2
   hooks:
     - id: ruff
-- repo: https://github.com/asottile/pyupgrade
-  rev: v3.15.1
-  hooks:
-    - id: pyupgrade
-      args: [--py311-plus]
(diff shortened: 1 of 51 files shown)


incognitojam:
pyupgrade does seem to apply the fixes automatically - not sure if we want pre-commit to make changes without asking since we haven't setup Ruff to do that.

There is an option to keep "percent formatting" strings but not much configuration beyond that.
image

.pre-commit-config.yaml
```diff
@@ -18,3 +18,8 @@ repos:
   rev: v0.2.2
   hooks:
     - id: ruff
+- repo: https://github.com/asottile/pyupgrade
```
adeebshihadeh, line 21: I don't think we want to run this in pre-commit. we'll just do these upgrades as they come up and make sense for us

board/flash.py
```diff
@@ -9,7 +9,7 @@
 if __name__ == "__main__":
   subprocess.check_call(f"scons -C {board_path}/.. -j$(nproc) {board_path}", shell=True)
 
-  serials = Panda.list()
+  serials = Panda.list_all()
```
adeebshihadeh, line 12: any way to avoid this? list is nice and short

adeebshihadeh (commented):
For these PRs, make sure to list all the commands that produce the diff.

board/flash.py
incognitojam: I don't know when this would be useful but it picks up the method over the list type when adding type hints
image

Maybe there is a way to implement `list(Panda)` instead? that seems python-ish

incognitojam: oh! I might have made a mistake. it might only affect the @staticmethod in PandaDFU, not the @classmethod in Panda

incognitojam: moving the method to the end of the file is a workaround

PR #1890, opened 2024-02-27
Enable misra-c2012-10.6

savojovic:
for #1794


adeebshihadeh:
Already done in https://github.com/commaai/panda/pull/1811

PR #1895, opened 2024-03-09
Honda: Nidec COMPUTER_BRAKE_ALT safety

AravindaDP:
Safety mode for COMPUTER_BRAKE_ALT used by Honda Clarity PHEV and Honda Fit Hybrid JDM vehicles.

 - define flag

 - add tests

 - fix tx and rx hooks


adeebshihadeh:
We've decided the Clarity port isn't worth it to us to upstream - see #20268

AravindaDP:
@adeebshihadeh Noted.

Just a quick question. Is decision for Clarity fork because of

a) It needing this alternate safety mode and needing logic change at hondacan.py to accommodate COMPUTER_BRAKE_ALT?

b) Because it needs a proxy panda gateway to copy messages from extra CAN bus? (i.e. Additional H/W requirement/different harness connector)

c) Both of above requirement?

d) Simply because there aren't enough Clarity users irrespective of above complexities? Is there a minimum number of users that warrants a model port being accepted?

As a note I was trying my luck getting Honda Fit Hybrid JDM up streamed. Following is full details.

1. Both models need this Panda safety mode. If it is not welcomed no matter how many models out their using it, no worries I understand.

2. Both models need change at hondacan.py to accommodate COMPUTER_BRAKE_ALT in hondacan.py apart from usual model port additions.

3. I guess Clarity will need Different harness connector. For Honda Fit Hybrid JDM I submitted here: https://github.com/commaai/neo/pull/80 perhaps same might work for Clarity as well. (Along with proxy panda gateway)

However complexities ends there for Honda Fit Hybrid JDM.

It does not need a proxy panda gateway like Clarity.

adeebshihadeh:
C and D. The calculation in terms of difficulty to get merged here is roughly: new/custom HW > panda code > openpilot code. Requiring two or three of those means we expect to support lots of models and users.

PR #1896, opened 2024-03-11
Enabling misra-c2012-1.2

mpetkovic96:
for #1794
The problem was with the __typeof__ expression, which is not supported by MISRA-C2012-1.2 rule. Instead of this expression, shorts static inline functions were written. There was a problem when an unsigned value is compared with zero; it reported that the value is always greater than zero. Because of that, arguments of functions are unsigned int. For signed values, conversions would be made (for example, if you want the maximum between -2 and -3, first it would be converted to unsigned, which is 254 for -2 and 253 for -3, where 254 is greater than 253, which is correct). In safety_ford.h file, instead of the MAX macro, I use the ternary operator for float values because if I use the MAX macro, floats would be converted to unsigned, where 0.1 becomes 0, and I think that could be a problem.


mpetkovic96:
@adeebshihadeh
Could you check if there is any potential?

PR #1923, opened 2024-04-02
Defining types in return dictionaries

MarinkoMagla:
It was under "TODO" in python/ccp.py to define types for couple of function returns.

### DIFF
--- a/python/ccp.py
+++ b/python/ccp.py
@@ -3,6 +3,30 @@ import time
 import struct
 from enum import IntEnum, Enum
 
+from typing import TypedDict, Optional
+
+class ExchangeStationIdsReturn(TypedDict):
+    id_length: int
+    data_type: int
+    available: int  
+    protected: int  
+
+class GetDaqListSizeReturn(TypedDict):
+    list_size: int
+    first_pid: int
+
+class GetSessionStatusReturn(TypedDict):
+    status: int
+    info: Optional[int]   
+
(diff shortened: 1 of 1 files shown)


python/ccp.py
```diff
@@ -3,6 +3,30 @@
...
+    status: int
+    info: Optional[int]   
+
+class DiagnosticServiceReturn(TypedDict):
+    length: int
+    type: int
+
+class ActionServiceReturn(TypedDict):
+    length: int
+    type: int
```
adeebshihadeh, line 28: if we're gonna do this, might as well just return the class instead of a dict. what do you think @gregjhogan?

python/ccp.py
```diff
@@ -3,6 +3,30 @@
 import struct
 from enum import IntEnum, Enum
 
+from typing import TypedDict, Optional
+
+class ExchangeStationIdsReturn(TypedDict):
+    id_length: int
```
adeebshihadeh, line 9: indent is wrong

MarinkoMagla:
Hi @adeebshihadeh, can you take another look on this?

adeebshihadeh:
This comment still applies https://github.com/commaai/panda/pull/1923#discussion_r1552267692

adeebshihadeh:
Linter is still failing

MarinkoMagla:
Hi @adeebshihadeh, linter is ok now, can you take another look?

PR #1928, opened 2024-04-11
Enable misra-c2012-8.4

mpetkovic96:
for #1794 
Where it was possible, external and static linking was used. Otherwise, suppression was utilized. Additionally, global variable scopes were reduced to function scope.


mpetkovic96:
@adeebshihadeh Could you check if this approach is correct?

maxime-desroches:
Closing because 8.4 is solved now

PR #1930, opened 2024-04-13
WIP: Acura integra safety

vanillagorillaa:
Opening as a starting point to start the discussion. The Integra does not have a ENGINE_DATA message, so the panda RX checks fail. For the moment I have removed this check for the port, but will want to come up with a clean way to do this. 

Instead of using ENGINE_DATA (which uses the XMISSION_SPEED) signal for vEGO in openpilot, I've used the CAR_SPEED signal for vEGO instead. 

@sshane Thoughts?


sshane:
Have a route? Will need a flag, but can we switch over the safety to use CAR_SPEED? If that always exists, are there any issues with it on existing cars?

vanillagorillaa:
Neither the CVT or MT integra's have the ENGINE_DATA message (looked for it on different ID's but didn't find it at all). 0x309 will be used for CAR_SPEED

PR #1933, opened 2024-04-18
IsoTpMessage: extend timeout on first frame response

sshane:
Try 2 of https://github.com/commaai/panda/pull/1487 and https://github.com/commaai/panda/pull/1610. Potentially fixes https://github.com/commaai/openpilot/issues/32242

In short, this adds checks to first and single ISO-TP frame types. If you receive a first frame, the next message must be consecutive, or the new assertions in first and single are raised. You can receive as many single frames as you want without issue.

An additional rx length check is added to ensure that `rx_dat` isn't unset in the first frame hook (it previously accepted `m._isotp_rx_next(b'\x10\x08')`) which isn't a [valid CAN frame size](https://en.wikipedia.org/wiki/ISO_15765-2).

---

This would cause the EPS to now *successfully* respond for:

- `297b4b460f361603/2023-06-14--16-55-16`
- `297b4b460f361603/2023-06-14--16-48-52`
- `297b4b460f361603/2023-06-21--04-09-15`

### DIFF
--- a/python/uds.py
+++ b/python/uds.py
@@ -441,8 +441,6 @@ class IsoTpMessage():
       if self.debug and not setup_only:
         print(f"ISO-TP: TX - first frame - {hex(self._can_client.tx_addr)}")
       msg = (struct.pack("!H", 0x1000 | self.tx_len) + self.tx_dat[:self.max_len - 2]).ljust(self.max_len - 2, b"\x00")
-    print('sent msg', msg)
-    print(f'{self.tx_dat=}, {self.tx_len=}, {self.tx_idx=}, {self.tx_done=}, {self.rx_dat=}, {self.rx_len=}, {self.rx_idx=}, {self.rx_done=}')
     if not setup_only:
       self._can_client.send([msg])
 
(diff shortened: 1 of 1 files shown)


python/uds.py
```diff
@@ -477,7 +477,7 @@ def _isotp_rx_next(self, rx_data: bytes) -> ISOTP_FRAME_TYPE:
       assert self.rx_len < self.max_len, f"isotp - rx: invalid single frame length: {self.rx_len}"
       self.rx_dat = rx_data[1:1 + self.rx_len]
       self.rx_idx = 0
-      self.rx_done = True
+      self.rx_done = error_code != 0x78  # todo me
```
sshane, line 480: needs to handle response pending. you get a single frame, then a first or single frame which is valid

sshane:
Cases we need to handle for VW:

case 1:
- single frame: error code 0x78
- some time later: first frame, consecutive frame, etc.

case 2:
- single frame: error code 0x78
- some time later, single frame

cases that can pass/doesn't matter:
- single frame
- single frame
- etc.

cases that should fail:
case 1:
- single frame
- first frame

case 2:
- first frame
- first frame
- etc.

PR #1934, opened 2024-04-18
CI: fix misra mutation tests

jnewb1:
this has been failing but pytest.main doesn't exit, you need to exit manually...

```
 _____________________ test_misra_mutation[None-None-False] _____________________
[gw0] linux -- Python 3.11.4 /root/.pyenv/versions/3.11.4/bin/python3

fn = None, patch = None, should_fail = False

    @pytest.mark.parametrize("fn, patch, should_fail", mutations)
    def test_misra_mutation(fn, patch, should_fail):
      with tempfile.TemporaryDirectory() as tmp:
        shutil.copytree(ROOT, tmp, dirs_exist_ok=True)
    
        # apply patch
        if fn is not None:
          r = os.system(f"cd {tmp} && sed -i '{patch}' {fn}")
          assert r == 0
    
        # run test
        r = subprocess.run("tests/misra/test_misra.sh", cwd=tmp, shell=True)
        failed = r.returncode != 0
>       assert failed == should_fail
E       assert True == False

test_mutation.py:81: AssertionError
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -8,5 +8,8 @@ select = ["E", "F", "W", "PIE", "C4", "ISC", "RUF100", "A"]
 ignore = ["W292", "E741", "E402", "C408", "ISC003"]
 flake8-implicit-str-concat.allow-multiline=false
 
+[tool.ruff.lint.flake8-tidy-imports.banned-api]
+"pytest.main".msg = "pytest.main requires special handling that is easy to mess up!"
+
 [tool.pytest.ini_options]
 addopts = "-n auto"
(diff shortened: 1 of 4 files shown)


pyproject.toml
```diff
@@ -8,5 +8,8 @@ select = ["E", "F", "W", "PIE", "C4", "ISC", "RUF100", "A"]
 ignore = ["W292", "E741", "E402", "C408", "ISC003"]
 flake8-implicit-str-concat.allow-multiline=false
 
+[tool.ruff.lint.flake8-tidy-imports.banned-api]
+"pytest.main".msg = "pytest.main requires special handling that is easy to mess up!"
```
adeebshihadeh, line 12: good catch! let's add this to openpilot too

jnewb1: we added it here when we had similar problems in `test_onroad`: https://github.com/commaai/openpilot/pull/31623

PR #1935, opened 2024-04-19
fix misra race condition

jnewb1:


tests/misra/test_misra.sh
```diff
@@ -9,9 +9,10 @@ NC='\033[0m'
 
 : "${CPPCHECK_DIR:=$DIR/cppcheck/}"
 
-# install cppcheck if missing
-if [ -z "${SKIP_BUILD}" ]; then
-  $DIR/install.sh
```
adeebshihadeh, line 14: `./test_misra.sh` should always get your cppcheck in the right state

PR #1940, opened 2024-04-27
[$100 bounty] Match stock Hyundai button logic

ksfi:
Safety checks related to bounty which implements the pause/resume button HKG cars logic
Cruise speed is taken from SCC11 that uses here the same bus logic as SCC12 but i do not have any device so couldn't check

OP PR: https://github.com/commaai/openpilot/pull/32306


sshane:

PR #1947, opened 2024-05-08
use pytest instead of unittest

BBBmau:
Matches the changes that are being done in openpilot repo


adeebshihadeh:
This has gone stale, feel free to reopen if you pick it back up.

PR #1948, opened 2024-05-14
Toyota LTA: fix the way the torque `sample_t`'s are used

sshane:


sshane:

PR #1950, opened 2024-05-17
Subaru: Increase Global Gen 2 steering torque

CheckYourSix:
Increases the max steering torque for Subaru Global Gen 2 vehicles

**Prerequisite of**
* [Subaru: Increase Global Gen 2 steering torque openpilot#32451](https://github.com/commaai/openpilot/pull/32451)


adeebshihadeh:
Responded in the openpilot PR.

PR #1830, opened 2024-01-22
enable misra-c2012-8.7

samrum:

PR #1953, opened 2024-05-22
Enable misra c2012 8.7

dzid26:
Bounty https://github.com/commaai/panda/issues/1794
I picked up after https://github.com/commaai/panda/pull/1830., rebased on the head and made minor fixes.


dzid26:
### It fixes the rule, but it's pointless in principal. 

In the context of rule 8.7, using `static` is meant to reduce visibility within a given translation unit. 
But Panda has only one translation unit (`main.c`) as it doesn't use the common source/header pair architecture, so it requires `static` in front of every function and object defined externally. 

Cppcheck 2.13 doesn't detect all of them as violations (maybe it gets confused by the hierarchical includes).  cppcheck 2.14 detects more, but not all off them.  For example it detects [dos_check_ignition()](https://github.com/commaai/panda/blob/788f92a77662dfc67b018228d91b8a19d0b0d0b1/board/boards/dos.h#L88) but not [dos_read_som_gpio()](https://github.com/commaai/panda/blob/788f92a77662dfc67b018228d91b8a19d0b0d0b1/board/boards/dos.h#L105)


There is one benefit of solving 8.7 that I can see. - It allows to detect 8.9 violations. 8.9 is about narrowing scope of the objects - which is good. It sometimes makes the code look more messy though, when constants need to be moved from top of the file to a function body.

PR #1960, opened 2024-05-28
Generate cppcheck checkers list

dzid26:
### Generate and validate cppcheck checker list.

Checker list is generated for each test run variant and appended to `checkers_report.txt`.
`checkers_report.txt` is then compared against `checkers.txt` in git.

### DIFF
--- a/tests/misra/test_misra.sh
+++ b/tests/misra/test_misra.sh
@@ -78,6 +78,6 @@ printf "\n${GREEN}Success!${NC} took $SECONDS seconds\n"
 # ensure list of checkers is up to date
 cd $DIR
 if ! git diff --quiet $CHECKLIST; then
-  echo -e "\n${YELLOW}Warning: Cppcheck checkers.txt report has changed. Review and commit...${NC}"
+  echo -e "\n${YELLOW}WARNING: Cppcheck checkers.txt report has changed. Review and commit...${NC}"
   exit 1
 fi
--- a/tests/misra/test_misra.sh
+++ b/tests/misra/test_misra.sh
@@ -5,6 +5,8 @@ DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
 PANDA_DIR=$(realpath $DIR/../../)
 
 GREEN="\e[1;32m"
+YELLOW="\e[1;33m"
+RED="\e[1;31m"
 NC='\033[0m'
 
 : "${CPPCHECK_DIR:=$DIR/cppcheck/}"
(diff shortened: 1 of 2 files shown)


tests/misra/test_misra.sh
```diff
@@ -63,3 +74,12 @@ cppcheck $PANDA_OPTS -DSTM32H7 -DSTM32H725xx $PANDA_DIR/board/main.c
 
 printf "\n${GREEN}Success!${NC} took $SECONDS seconds\n"
 
+
+# ensure list of checkers is up to date
+cd $DIR
+if ! cmp -s checkers_report.txt checkers.txt; then
```
adeebshihadeh, line 80: I think it would be nicer to just overwrite the file, then use `git` to check for a diff. Then, it's just `git add` to update it.

dzid26: I agree, I was thinking that might be better.

dzid26: Another issue is that I accidentally pass an absolute path `$PANDA_DIR` to the report.  
I think I could skip $PANDA_DIR and just call `cppcheck $PANDA_OPTS -DSTM32F4 -DSTM32F413xx /oard/main.c`, but that somehow enables bunch of new violations.

adeebshihadeh: oh - let's get those fixed :)

tests/misra/test_misra.sh
```diff
@@ -28,23 +29,33 @@ if [ -z "${SKIP_BUILD}" ]; then
...
+          --platform=arm32-wchar_t4 $COMMON_DEFINES --checkers-report=$CHECKLIST.tmp \
           "$@" |& tee $OUTPUT
-
+  
+  cat $CHECKLIST.tmp >> $CHECKLIST
+  rm $CHECKLIST.tmp
   # cppcheck bug: some MISRA errors won't result in the error exit code,
   # so check the output (https://trac.cppcheck.net/ticket/12440#no1)
   if grep -e "misra violation" -e "error" -e "style: " $OUTPUT > /dev/null; then
+    printf "${RED}** FAIL!!!${NC}\n"
```
adeebshihadeh, line 58:
```suggestion
    printf "${RED}** FAILED: MISRA violations found!${NC}\n"
```

tests/misra/test_misra.sh
```diff
@@ -63,3 +74,10 @@ cppcheck $PANDA_OPTS -DSTM32H7 -DSTM32H725xx $PANDA_DIR/board/main.c
 
 printf "\n${GREEN}Success!${NC} took $SECONDS seconds\n"
 
+
+# ensure list of checkers is up to date
+cd $DIR
+if ! git diff --quiet $CHECKLIST; then
+  echo -e "\n${YELLOW}Warning: Cppcheck checkers.txt report has changed. Review and commit...${NC}"
```
adeebshihadeh, line 81:
```suggestion
  echo -e "\n${YELLOW}WARNING: Cppcheck checkers.txt report has changed. Review and commit...${NC}"
```

PR #1963, opened 2024-05-30
remove void return

dzid26:
It looks that's the only one


adeebshihadeh:
Is there a cppcheck or compiler flag we can enable to catch this?

dzid26:
It would be `-Wpedantic` or some other ISO-C option which in turn ignores `-std=gnu11` and throws bunch of other errors.    I tried selectively suppressing them with #pragmas, but that didn't want to work.  It would need to be investigated more, but so far I am pessimistic.

PR #1968, opened 2024-06-04
fdcan: don't log bad CAN messages

sshane:
For https://github.com/commaai/panda/pull/1967. Now messages that we send to a missing comma power won't come back as "returned"


sshane:
Igor reminded me we have retransmission enabled, so a message that comes back with an error code is still potentially in the hardware queue and will be transmitted when it can, so we can't know if it actually sent or not:

https://github.com/commaai/panda/blob/4df996cfd7ffaef5ad48a6ae8245d33a1414befa/board/stm32h7/llfdcan.h#L196-L197

PR #1972, opened 2024-06-05
drivers: log can core reset when switching multiplexing

sshane:
fdcan: log lost tx messages when resetting can core for multiplexing and can core reset
bxcan: log can core reset

follow up to https://github.com/commaai/panda/pull/1967


adeebshihadeh:
Is this good to merge? or close?

PR #1978, opened 2024-06-30
no serial sometimes

adeebshihadeh:


adeebshihadeh:
This device had some other issues going on.

PR #1982, opened 2024-07-10
Convert to pytest

UkuLoskit:
There is an issue with the usbprotocol tests, they depend on a particular ordering of the tests. I used pytest-order to preserve the default ordering of unittest. Additionally, these tests cannot be run in parallel. Used this ordering as a crutch to work around this.


maxime-desroches:
@UkuLoskit Do you plan on finishing this up soon?

UkuLoskit:
@maxime-desroches hey, ready for review now with some caveats described in the PR description.

UkuLoskit:
There are actually more tests ... fixing these

adeebshihadeh:
trigger-jenkins

tests/safety/test_elm327.py
```diff
@@ -26,23 +26,23 @@ def test_tx_hook(self):
...
+      assert should_tx == self._tx(common.make_msg(0, GM_CAMERA_DIAG_ADDR, dat=bytes([byte] * 8)))
 
   def test_tx_hook_on_wrong_safety_mode(self):
     # No point, since we allow many diagnostic addresses
     pass
 
 
 if __name__ == "__main__":
-  unittest.main()
+  pytest.main()
```
adeebshihadeh, line 48: the linter should be configured to block this. how is it passing?

pyproject.toml
```diff
@@ -10,6 +10,3 @@ flake8-implicit-str-concat.allow-multiline=false
 
 [tool.ruff.lint.flake8-tidy-imports.banned-api]
 "pytest.main".msg = "pytest.main requires special handling that is easy to mess up!"
-
-[tool.pytest.ini_options]
-addopts = "-n auto"
```
adeebshihadeh, line 15: why was this removed? we want parallel by default

tests/usbprotocol/test_comms.py
```diff
@@ -99,31 +102,30 @@ def test_comms_reset_tx(self):
...
+      pkt = libpanda_py.ffi.new('CANPacket_t *')
+      while lpp.can_pop(TX_QUEUES[bus], pkt):
+        queue_msgs.append(unpackage_can_msg(pkt))
 
-          self.assertEqual(len(queue_msgs), len(msgs))
-          self.assertEqual(queue_msgs, msgs)
+      assert len(queue_msgs) == len(msgs)
+      assert queue_msgs == msgs
 
+  @pytest.mark.order(1)
```
adeebshihadeh, line 128: why is this needed? the tests should be independent

adeebshihadeh (changes requested):

tests/usbprotocol/test_comms.py
UkuLoskit: @adeebshihadeh I have explained this in PR description The tests are dependent on a definite ordering which be reproduced with unittests if you change the method names. E.g test_zzz... for the order(5) five test for example.

UkuLoskit: @adeebshihadeh just bumping this since I haven't heard anything back in a while. I think the easiest way to allow parallelism would be to mark the tests that depend on a particular order to run serially. that way parallelism could still be used for other tests. Fixing the tests themselves, seems like a more complicated task.

adeebshihadeh:
Closing, I left feedback and it wasn't addressed. Feel free to open a new one without any "crutches".

PR #1983, opened 2024-07-13
workflow: update readme with straightforward setup [ubuntu]

BBBmau:
dependencies didn't seem to match the setup that CI has, with the readme having missing apt packages such as `python3-dev` which is important for setup.

their is also no mention of the python version which must at least be 3.9

I've also included a section for updating permissions since even after using the appropriate permissions this was still the last step needed to be able to successfully run `python setup.py install`

### DIFF
--- a/install_ubuntu_dependencies.sh
+++ b/install_ubuntu_dependencies.sh
@@ -1,0 +1,40 @@
+#!/usr/bin/env bash
+set -e
+
+SUDO=""
+
+# Use sudo if not root
+if [[ ! $(id -u) -eq 0 ]]; then
+  if [[ -z $(which sudo) ]]; then
+    echo "Please install sudo or run as root"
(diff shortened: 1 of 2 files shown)


install_ubuntu_dependencies.sh
```diff
@@ -0,0 +1,40 @@
...
+  fi
+  SUDO="sudo"
+fi
+
+$SUDO apt-get update 
+$SUDO apt-get install -y --no-install-recommends \
+    make \
+    bzip2 \
+    ca-certificates \
+    capnproto \
```
adeebshihadeh, line 20: most of this stuff isn't needed, closing

adeebshihadeh: panda shouldn't have enough dependencies to even warrant a script like this

PR #1988, opened 2024-07-30
Detect H723 in firmware

robbederks:
(another attempt at #1975)

Unfortunately, I couldn't find a good way to determine if the chip has the SMPS (725) or not (723) [without multiple writes to](https://github.com/commaai/panda/blob/d56bad2b7f2309e7490720b3b32341de9644096a/board/stm32h7/clock.h#L22) `PWR->CR3`, and this appears to be the only real difference between the two.

This detection is based on the package type, which is nicely exposed in `SYSCFG->PKGR`. This does mean we can't have any overlap between 723 and 725 /735 packages in future designs, but it's unlikely we'll need to.

### DIFF
--- a/board/stm32h7/clock.h
+++ b/board/stm32h7/clock.h
@@ -21,7 +21,7 @@ PCLK1: 60MHz (for USART2,3,4,5,7,8)
 // The package will do for now, since we have only used TFBGA100 for H723
 bool is_h723(void) {
   uint8_t package = SYSCFG->PKGR & 0xFU;
-  return package == 0b0001U || package == 0b0011U; // TFBGA100 Legacy || TFBGA100
+  return (package == 0b0001U) || (package == 0b0011U); // TFBGA100 Legacy || TFBGA100
 }
 
 void clock_init(void) {
--- a/board/stm32h7/clock.h
+++ b/board/stm32h7/clock.h
@@ -17,6 +17,13 @@ APB4 per: 60MHz
 PCLK1: 60MHz (for USART2,3,4,5,7,8)
 */
 
+// TODO: find a better way to distinguish between H723 and H725
+// The package will do for now, since we have only used TFBGA100 for H723
+bool is_h723(void) {
+  uint8_t package = SYSCFG->PKGR & 0xFU;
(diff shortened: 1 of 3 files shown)


board/stm32h7/clock.h
```diff
@@ -17,6 +17,13 @@ APB4 per: 60MHz
 PCLK1: 60MHz (for USART2,3,4,5,7,8)
 */
 
+// TODO: find a better way to distinguish between H723 and H725
+// The package will do for now, since we have only used TFBGA100 for H723
+bool is_h723(void) {
+  uint8_t package = SYSCFG->PKGR & 0xFU;
```
adeebshihadeh, line 23: is `SYSCFG` enabled at this point? I don't think this is expected to be valid until the peripheral is enabled.

in the bootstub:
https://github.com/commaai/panda/blob/f6375848ca393a9483921665b6a2d131d7ec9b20/board/bootstub.c#L37-L43

in the app (which the CR3 behavior makes irrelevant):
https://github.com/commaai/panda/blob/f6375848ca393a9483921665b6a2d131d7ec9b20/board/main.c#L300-L303

robbederks: I assumed this wasn't clock gated, but I might be wrong (will check the reference manual).

It does seem to read out correctly at this point after a power cycle, but good to make sure.

robbederks: enabled it before reading out. APB4 should have some frequency on it already from boot

board/stm32h7/clock.h
```diff
@@ -29,11 +37,12 @@ void clock_init(void) {
   */
 
   // Set power mode to direct SMPS power supply(depends on the board layout)
-#ifndef STM32H723
-  register_set(&(PWR->CR3), PWR_CR3_SMPSEN, 0xFU); // powered only by SMPS
-#else
-  register_set(&(PWR->CR3), PWR_CR3_LDOEN, 0xFU);
-#endif
+  if (is_h723()) {
```
adeebshihadeh, line 40: let's at least whitelist both

robbederks:
Tested on:
- tres
- quatro
- red panda
- jungle v2

All worked, including after power cycle. Feel free to merge

PR #2011, opened 2024-09-05
Fix USB errors

dzid26:
Panda needed more time to be detected by USB host after reset.

USB errors would would show for the devices with USB hubs.

### DIFF
--- a/python/__init__.py
+++ b/python/__init__.py
@@ -255,7 +255,7 @@ class Panda:
     self._handle = None
     while self._handle is None:
       # try USB first, then SPI
-      self._context, self._handle, serial, self.bootstub, bcd = self.usb_connect(self._connect_serial, claim=claim)
+      self._context, self._handle, serial, self.bootstub, bcd = self.usb_connect(self._connect_serial, claim=claim, no_error=wait)
       if self._handle is None:
         self._context, self._handle, serial, self.bootstub, bcd = self.spi_connect(self._connect_serial)
       if not wait:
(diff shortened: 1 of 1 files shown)


adeebshihadeh:
A `time.sleep` is rarely the right fix. Why doesn't the polling below that line take care of this?

dzid26:
I got rid of the extra delay

board/jungle/scripts/reset_test.py
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env python3
+
```
adeebshihadeh, line 2: this is a good test to evaluate this change, but I don't think it has value going forward. I'd remove it

python/__init__.py
```diff
@@ -451,7 +452,7 @@ def reconnect(self):
     # wait up to 15 seconds
     for _ in range(15*10):
       try:
-        self.connect()
+        self.connect(False, True)
```
adeebshihadeh, line 455: use the kwargs
```suggestion
        self.connect(claim=False, wait=True)
```

python/__init__.py
```diff
@@ -347,7 +347,7 @@ def spi_connect(cls, serial, ignore_version=False):
     return None, handle, spi_serial, bootstub, None
 
   @classmethod
-  def usb_connect(cls, serial, claim=True):
+  def usb_connect(cls, serial, claim=True, wait=False):
```
adeebshihadeh, line 350: call it `silent`? wait doesn't mean anything in the context of this function

adeebshihadeh (changes requested):
LGTM aside from the comments

python/__init__.py
```diff
@@ -357,7 +357,8 @@ def usb_connect(cls, serial, claim=True):
           try:
             this_serial = device.getSerialNumber()
           except Exception:
-            logger.exception("failed to get serial number of panda")
+            if not wait:
```
adeebshihadeh, line 360: we should leave a good comment about why this is here

python/__init__.py
dzid26: I called it `no-error`

PR #2012, opened 2024-09-06
fix test_mutation.py performace

dzid26:
Makes `test_mutation.py` 3x faster when run locally:
```
real    3m0.221s
user    25m4.861s
sys     1m2.424s
vs
real    1m7.962s
user    8m0.304s
sys     0m33.491s
```

This is due to naming of `test_misra.sh`.  It was executed by pytest workers on their respective workpsace copy. They were compiling cppcheck and running all misra tests during test collection phase.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -12,4 +12,4 @@ flake8-implicit-str-concat.allow-multiline=false
 "pytest.main".msg = "pytest.main requires special handling that is easy to mess up!"
 
 [tool.pytest.ini_options]
-addopts = "-n auto"
+addopts = "-n auto --ignore-glob='*.sh'"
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -12,4 +12,4 @@ flake8-implicit-str-concat.allow-multiline=false
 "pytest.main".msg = "pytest.main requires special handling that is easy to mess up!"
 
 [tool.pytest.ini_options]
-addopts = "-n auto"
+addopts = "-n auto --ignore=test_misra.sh"
```
adeebshihadeh, line 15: we should only consider files that look like `test_*.py`. does this fix it?
```
python_files = ["test_*.py"]
```

dzid26: It didn't work.   `test_*.py` is the default anyway. 

[Shane's issue](https://github.com/pytest-dev/pytest/issues/11974) points to plugins interaction-  `pytest-cpp` in particular. 

We can fix it either with:
- `addopts = "-n auto --ignore-glob='*.sh'"` or
- `cpp_files = "test_*.py"`

PR #2020, opened 2024-09-13
Subaru LKAS Angle safety model

bravochar:
Newer (3-camera) EyeSight systems need angle-based steering controls, which means new safety models. Most of this is lifted from jnewb1's abandoned branch.

The gist is that we're adding a new message and bus for the LKAS angle controls and using the Gen2 alt bus for most everything else.


bravochar:
I just rebased this so it's more up-to-date. My OpenDBC and OpenPilot PRs will be blocked until this is merged. Is there anything I'm missing here?

sshane:

PR #2022, opened 2024-09-14
misra8.4: safety.h + safety_declarations.h

maxime-desroches:


tests/misra/test_misra.sh
```diff
@@ -36,7 +36,7 @@ echo "Cppcheck checkers list from test_misra.sh:" > $CHECKLIST
 
 cppcheck() {
   # get all gcc defines: arm-none-eabi-gcc -dM -E - < /dev/null
-  COMMON_DEFINES="-D__GNUC__=9 -UCMSIS_NVIC_VIRTUAL -UCMSIS_VECTAB_VIRTUAL"
+  COMMON_DEFINES="-D__GNUC__=9 -UCMSIS_NVIC_VIRTUAL -UCMSIS_VECTAB_VIRTUAL -DALLOW_DEBUG"
```
adeebshihadeh, line 39: what's this for? we want to test the release build

board/safety_declarations.h
```diff
@@ -208,39 +212,39 @@ void pcm_cruise_check(bool cruise_engaged);
...
-bool brake_pressed_prev = false;
-bool regen_braking = false;
-bool regen_braking_prev = false;
-bool cruise_engaged_prev = false;
-struct sample_t vehicle_speed;
-bool vehicle_moving = false;
-bool acc_main_on = false;  // referred to as "ACC off" in ISO 15622:2018
-int cruise_button_prev = 0;
-bool safety_rx_checks_invalid = false;
+extern bool controls_allowed;
```
adeebshihadeh, line 215: why do we need to specify extern?

maxime-desroches: `controls_allowed` is a global variable that is declared in safety_declarations.h and defined with external linkage in  safety.h

Misra 8.4 says that when defining with external linkage, you must have an equivalent declaration somewhere (so including the `extern` keyword)

tests/misra/test_misra.sh
maxime-desroches: cppcheck emits a false positive like[ this](https://github.com/commaai/panda/actions/runs/10859980022/job/30140003121#step:5:23) because of this [if statement](https://github.com/commaai/panda/blob/65f7852bb2362bc76c146a5bbac79dfcfedfaec4/board/safety/safety_ford.h#L412)

cppcheck does not understand the #ifdef ALLOW_DEBUG just above, which might change the variable.

I can add a suppression instead for that line only

board/safety/safety_chrysler.h
```diff
@@ -40,7 +17,8 @@ typedef struct {
 } ChryslerAddrs;
 
 // CAN messages for Chrysler/Jeep platforms
-const ChryslerAddrs CHRYSLER_ADDRS = {
+// cppcheck-suppress misra-c2012-8.9
```
adeebshihadeh, line 20: why are we adding new suppressions?

board/safety/safety_chrysler.h
```diff
@@ -1,32 +1,9 @@
...
-  .max_steer = 350,
-  .max_rt_delta = 112,
-  .max_rt_interval = 250000,
-  .max_rate_up = 6,
-  .max_rate_down = 6,
-  .max_torque_error = 80,
-  .type = TorqueMotorLimited,
-};
+#include <stdint.h>
+#include <stdbool.h>
```
adeebshihadeh, line 4: these really needed for every file?

board/safety/safety_body.h
```diff
@@ -34,6 +33,14 @@ static bool body_tx_hook(const CANPacket_t *to_send) {
 }
 
 static safety_config body_init(uint16_t param) {
+  static RxCheck body_rx_checks[] = {
+    {.msg = {{0x201, 0, 8, .check_checksum = false, .max_counter = 0U, .frequency = 100U}, { 0 }, { 0 }}},
+  };
+
+  static const CanMsg BODY_TX_MSGS[] = {{0x250, 0, 8}, {0x250, 0, 6}, {0x251, 0, 5},  // body
+                                 {0x350, 0, 8}, {0x350, 0, 6}, {0x351, 0, 5},  // knee
+                                 {0x1, 0, 8}}; // CAN flasher
```
adeebshihadeh, line 42: indentation is off

board/safety.h
```diff
@@ -52,11 +54,156 @@
...
+    const uint32_t shift = i * 8;
+    ret |= (((uint32_t)msg->data[start + i]) << shift);
+  }
+  return ret;
+}
+
+const int MAX_WRONG_COUNTERS = 5;
+
+// This can be set by the safety hooks
+bool controls_allowed = false;
```
adeebshihadeh, line 69: can't we just let the initial SAFETY_SILENT init take care of these?

board/safety/safety_ford.h
```diff
@@ -394,13 +344,71 @@ static int ford_fwd_hook(int bus_num, int addr) {
...
   UNUSED(param);
 #ifdef ALLOW_DEBUG
+  const uint16_t FORD_PARAM_LONGITUDINAL = 1;
+  const uint16_t FORD_PARAM_CANFD = 2;
   ford_longitudinal = GET_FLAG(param, FORD_PARAM_LONGITUDINAL);
   ford_canfd = GET_FLAG(param, FORD_PARAM_CANFD);
 #endif
 
   safety_config ret;
+  // cppcheck-suppress knownConditionTrueFalse
```
adeebshihadeh, line 411: why is this needed?

adeebshihadeh (commented):
@sshane want to take a quick look as well?

board/safety.h
maxime-desroches: No because you would get a bunch of `undefined reference to ...` since they are only declared in `safety_declarations.h` and not defined anywhere.

PR #2028, opened 2024-09-18
misra12.2

maxime-desroches:


adeebshihadeh:
Edited manually? I'd prefer not to do that if possible. I think I found some compliant ones though don't remember exactly where: #1917.

maxime-desroches:
Yeah I edited manually since we did it in the past to add the FSMC global Interrupt.

There is a bunch of headers on STMicro's github : https://github.com/STMicroelectronics/cmsis_device_f4/blob/master/Include/stm32f413xx.h
This one seems compliant with the rule, I will try it

adeebshihadeh (approved):

PR #2034, opened 2024-09-20
misra2.5: more removal

maxime-desroches:
Only removed unused ones. Will have to plan how to deal with false positives from cppcheck


board/drivers/gpio.h
```diff
@@ -7,7 +7,6 @@
 #define PULL_UP 1
 #define PULL_DOWN 2
 
-#define OUTPUT_TYPE_PUSH_PULL 0U
 #define OUTPUT_TYPE_OPEN_DRAIN 1U
```
adeebshihadeh, line 10: I'd just make these an enum

dzid26:
> Will have to plan how to deal with false positives from cppcheck

That's where I got stuck. i think the best way would be to use `--project` option with _compile database json_ from scons 310f1aa49341151cf6149673fb052219d7848f7a to give cppcheck all the information about the program configurations.

But first [this test](https://github.com/danmar/cppcheck/pull/6551#issuecomment-2351922904) needs to be fixed.

PR #2036, opened 2024-09-21
 Safety for Tesla Model 3 / Model Y

lukasloetkolben:
Add support for Tesla Model 3/Y

Prerequisite of: https://github.com/commaai/opendbc/pull/1263


tests/safety/common.py
```diff
@@ -950,9 +950,9 @@ def test_not_allow_user_brake_when_moving(self, _user_brake_msg=None, get_brake_
       _user_brake_msg = self._user_brake_msg
 
     # Brake was already pressed
+    self._rx(self._vehicle_moving_msg(self.STANDSTILL_THRESHOLD))
```
gregjhogan, line 953: This was moved to match the ordering of `test_allow_user_brake_at_zero_speed()` and fixes this test failing due to the vehicle moving message being the same message as the cruise state message, so when we set the vehicle in motion it also calls `pcm_cruise_check` which changes `controls_allowed`.

sshane:
- [ ] user lateral override safety
- [ ] double check bit parsing logic

PR #2041, opened 2024-09-25
compile libpanda with clang

maxime-desroches:
Split this from the mutation test PR since we have to fix a bunch of line coverage errors.

We currently only check line coverage, but not branch coverage. We should also enforce that in the future.


board/safety/safety_hyundai.h
```diff
@@ -49,16 +49,19 @@ static uint8_t hyundai_get_counter(const CANPacket_t *to_push) {
   uint8_t cnt = 0;
   if (addr == 0x260) {
     cnt = (GET_BYTE(to_push, 7) >> 4) & 0x3U;
-  } else if (addr == 0x386) {
+  } 
+  if (addr == 0x386) {
```
adeebshihadeh, line 53: it's important that these tools don't dictate our style. let's revert and revisit this

PR #1841, opened 2024-01-29
Safety: 2021+ Rav4 Prime

pd0wm:
TODO:

- [ ] Add unit tests for new flags
- [ ] Check if any of the new messages have checksum that can be verified by the panda
- [ ] Ensure blocking 0x131 from camera doesn't affect any existing Toyotas (or just put behind flag).

PR #2042, opened 2024-09-26
Toyota: 2021+ RAV4 Prime

jyoung8607:
Supersedes #1841.

- [ ] ~~Message naming struct cleanup, similar to Chrysler~~ later
- [x] Consolidate alt flags if possible
- [X] ~~Add unit tests for new flags~~ flags consolidated
- [x] Check if any of the new messages have checksum that can be verified by the panda
- [x] ~~Ensure blocking 0x131 from camera doesn't affect any existing Toyotas (or just put behind flag)~~ put behind flag


board/safety/safety_toyota.h
```diff
@@ -203,6 +226,19 @@ static bool toyota_tx_hook(const CANPacket_t *to_send) {
...
+    // SecOC cars block any form of LTA actuation for now
+    if (toyota_secoc_car && (addr == 0x131)) {
+      bool lta_request = GET_BIT(to_send, 3U);
+      bool lta_request2 = GET_BIT(to_send, 0U);
+      int lta_angle_msb = GET_BYTE(to_send, 2);
+      int lta_angle_lsb = GET_BYTE(to_send, 3);
+
+      bool actuation = lta_request || lta_request2 || (lta_angle_msb != 0) || (lta_angle_lsb != 0);
+      if (actuation) {
+        tx = false;
```
jyoung8607, line 238: @maxime-desroches Can you help me understand why the new mutation test dislikes this?

https://github.com/commaai/panda/actions/runs/11135924813/job/30946827139?pr=2042

maxime-desroches: Should be fixed with a rebase

tests/safety/test_toyota.py
```diff
@@ -324,5 +326,36 @@ def setUp(self):
...
+  RELAY_MALFUNCTION_ADDRS = {0: (0x2E4,)}
+  FWD_BLACKLISTED_ADDRS = {2: [0x2E4, 0x412, 0x191, 0x131]}
+
+  def setUp(self):
+    self.packer = CANPackerPanda("toyota_rav4_prime_generated")
+    self.safety = libpanda_py.libpanda
+    self.safety.set_safety_hooks(Panda.SAFETY_TOYOTA, self.EPS_SCALE | Panda.FLAG_TOYOTA_STOCK_LONGITUDINAL | Panda.FLAG_TOYOTA_SECOC_CAR)
+    self.safety.init_tests()
+
+  # This platform also has alternate brake and PCM messages, but same naming in the DBC, so same packers orks
```
adeebshihadeh, line 341: typo

python/__init__.py
```diff
@@ -187,6 +187,7 @@ class Panda:
   FLAG_TOYOTA_ALT_BRAKE = (1 << 8)
   FLAG_TOYOTA_STOCK_LONGITUDINAL = (2 << 8)
   FLAG_TOYOTA_LTA = (4 << 8)
+  FLAG_TOYOTA_SECOC_CAR = (8 << 8)
```
adeebshihadeh, line 190:
```suggestion
  FLAG_TOYOTA_SECOC = (8 << 8)
```

board/safety/safety_toyota.h
```diff
@@ -87,14 +98,21 @@ static void toyota_rx_hook(const CANPacket_t *to_push) {
...
 
     // enter controls on rising edge of ACC, exit controls on ACC off
-    // exit controls on rising edge of gas press
-    if (addr == 0x1D2) {
-      // 5th bit is CRUISE_ACTIVE
+    if (!toyota_secoc_car && (addr == 0x1D2)) {
+      bool cruise_engaged = GET_BIT(to_push, 5U);
+      pcm_cruise_check(cruise_engaged);
+    }
+    if (toyota_secoc_car && (addr == 0x176)) {
```
adeebshihadeh, line 105: so many of these are different, I'd consider a big `if secoc; else`. not sure that it's cleaner though, up to you

adeebshihadeh (approved):

board/safety/safety_toyota.h
jyoung8607: I did that in the car port, but in Panda safety, it's nice having all the alternate-gas, alternate-brake, etc code in a place where it can be seen and understood together.

I might feel differently if Panda had the ability (and test coverage) to reject entry into a safety mode if openpilot sent nonsense flag combinations.

board/safety/safety_toyota.h
```diff
@@ -87,14 +98,21 @@ static void toyota_rx_hook(const CANPacket_t *to_push) {
...
 
-      // sample gas pedal
-      gas_pressed = !GET_BIT(to_push, 4U);
+    // exit controls on rising edge of gas press
+    if (!toyota_secoc && (addr == 0x1D2)) {
+      gas_pressed = !GET_BIT(to_push, 4U); // GAS_PEDAL.GAS_RELEASED
+    }
+    if (toyota_secoc && (addr == 0x116)) {
+      gas_pressed = GET_BYTE(to_push, 1) != 0U; // GAS_PEDAL.GAS_PEDAL_USER
     }
```
sshane, line 116:
```suggestion
    // exit controls on rising edge of gas press
    if (!toyota_secoc) {
      if (addr == 0x1D2) {
        gas_pressed = !GET_BIT(to_push, 4U); // GAS_PEDAL.GAS_RELEASED
      }
    } else {
      if (addr == 0x116) {
        gas_pressed = GET_BYTE(to_push, 1) != 0U; // GAS_PEDAL.GAS_PEDAL_USER
      }
    }
```

board/safety/safety_toyota.h
sshane: I also think combining these if `toyota_secoc_car` checks in one large if else is cleaner

board/safety/safety_toyota.h
```diff
@@ -203,6 +226,19 @@ static bool toyota_tx_hook(const CANPacket_t *to_send) {
       }
     }
 
+    // SecOC cars block any form of LTA actuation for now
+    if (toyota_secoc && (addr == 0x131)) {
```
sshane, line 230: I would now label these three blocks with message names (ideally we make a struct that we can use instead of addresses soon)

python/__init__.py
sshane: not resolved?

board/safety/safety_toyota.h
```diff
@@ -298,14 +338,23 @@ static safety_config toyota_init(uint16_t param) {
...
   const uint32_t TOYOTA_PARAM_LTA = 4UL << TOYOTA_PARAM_OFFSET;
 
+#ifdef ALLOW_DEBUG
+  const uint32_t TOYOTA_PARAM_SECOC = 8UL << TOYOTA_PARAM_OFFSET;
+  toyota_secoc = GET_FLAG(param, TOYOTA_PARAM_SECOC);
+#endif
+
   toyota_alt_brake = GET_FLAG(param, TOYOTA_PARAM_ALT_BRAKE);
   toyota_stock_longitudinal = GET_FLAG(param, TOYOTA_PARAM_STOCK_LONGITUDINAL);
   toyota_lta = GET_FLAG(param, TOYOTA_PARAM_LTA);
```
sshane, line 348: Are there sec oc cars that aren't LTA?

tests/safety/test_toyota.py
```diff
@@ -324,5 +326,36 @@ def setUp(self):
...
+
+  # This platform also has alternate brake and PCM messages, but same naming in the DBC, so same packers work
+
+  def _user_gas_msg(self, gas):
+    values = {"GAS_PEDAL_USER": gas}
+    return self.packer.make_can_msg_panda("GAS_PEDAL", 0, values)
+
+  def _lta_msg(self, req, req2, angle_cmd, torque_wind_down=100):
+    values = {"STEER_REQUEST": req, "STEER_REQUEST_2": req2, "STEER_ANGLE_CMD": angle_cmd}
+    return self.packer.make_can_msg_panda("STEERING_LTA_2", 0, values)
```
sshane, line 349: If we still use `STEERING_LTA`, do we no longer test that?

tests/safety/test_toyota.py
```diff
@@ -324,5 +326,36 @@ def setUp(self):
...
+class TestToyotaSecOcSafety(TestToyotaStockLongitudinalBase):
+
+  TX_MSGS = TOYOTA_SECOC_TX_MSGS
+  RELAY_MALFUNCTION_ADDRS = {0: (0x2E4,)}
+  FWD_BLACKLISTED_ADDRS = {2: [0x2E4, 0x412, 0x191, 0x131]}
+
+  def setUp(self):
+    self.packer = CANPackerPanda("toyota_rav4_prime_generated")
+    self.safety = libpanda_py.libpanda
+    self.safety.set_safety_hooks(Panda.SAFETY_TOYOTA, self.EPS_SCALE | Panda.FLAG_TOYOTA_STOCK_LONGITUDINAL | Panda.FLAG_TOYOTA_SECOC_CAR)
```
sshane, line 338: Missing `TOYOTA_PARAM_LTA`, so this is testing torque and sec oc?

python/__init__.py
jyoung8607: my bad, missed these on my search-and-replace sweep

board/safety/safety_toyota.h
jyoung8607: Do you mean trailing comments, like `// STEERING_LTA_2.LTA_REQUEST`?

tests/safety/test_toyota.py
jyoung8607: We do send both. Both are present on the car, not sure why. Your point is correct. Reverted to Draft, I'll fix up the tests in the AM to verify no-actuation on both.

sshane: Did we ever check what the stock system uses? We were able to send LKA on some 2023 RAV4 but because the stock system used LTA for lane keep we chose to switch off the unused API.

pd0wm: If I recall correctly the stock system uses the LTA for lane keep steering, but the torque message works fine too. So it's like most TSS2 cars in this regard.

They split up the LTA message into two parts, the old one which contains a bunch of values and flags (with no space for a MAC), and a new one that contains just the actual requested steering angle + a SecOC MAC. They did something similar for longitudinal.

board/safety/safety_toyota.h
jyoung8607: Added comments, I'm assuming that's what you were looking for if we're deferring an overall refactor of the entire safety to use message naming macros.

tests/safety/test_toyota.py
jyoung8607: Now testing both messages, thanks for spotting that!

Verified by breaking the 0x191 check offline:
```C
bool steer_control_enabled = lta_request; //|| lta_request2;
```

Which was properly picked up by both normal and SecOC tests:

```
FAILED test_toyota.py::TestToyotaSecOcSafety::test_lta_steer_cmd - AssertionError: False != True : req=0 req2=1 angle=0.0 torque_wind_down=0
FAILED test_toyota.py::TestToyotaStockLongitudinalTorque::test_lta_steer_cmd - AssertionError: False != True : req=0 req2=1 angle=0.0 torque_wind_down=0
FAILED test_toyota.py::TestToyotaSafetyTorque::test_lta_steer_cmd - AssertionError: False != True : req=0 req2=1 angle=0.0 torque_wind_down=0
```

board/safety/safety_toyota.h
```diff
@@ -203,6 +225,19 @@ static bool toyota_tx_hook(const CANPacket_t *to_send) {
...
     }
 
+    // SecOC cars block any form of LTA actuation for now
+    if (toyota_secoc && (addr == 0x131)) {
+      bool lta_request = GET_BIT(to_send, 3U);  // STEERING_LTA_2.STEER_REQUEST
+      bool lta_request2 = GET_BIT(to_send, 0U);  // STEERING_LTA_2.STEER_REQUEST_2
+      int lta_angle_msb = GET_BYTE(to_send, 2);  // STEERING_LTA_2.STEER_ANGLE_CMD (MSB)
+      int lta_angle_lsb = GET_BYTE(to_send, 3);  // STEERING_LTA_2.STEER_ANGLE_CMD (LSB)
+
+      bool actuation = lta_request || lta_request2 || (lta_angle_msb != 0) || (lta_angle_lsb != 0);
```
jyoung8607, line 235: Just a note for review: I originally wrote this to build `lta_angle`, like the other LTA message. Unfortunately the [mutation test doesn't like it](684bf3aab21b4bb6e1fcc8f52ab8e4d76c3f2c9b). We never have actuation via this path, so the angle is never valid while non-zero, so the mutation test sees it can introduce faults in the bit-shift operation without consequence.

board/safety/safety_toyota.h
jyoung8607: Gave it a try and I think it does look better, so I'm onboard.

tests/safety/test_toyota.py
sshane: @adeebshihadeh do you want to merge this with LKA when we specifically reverted LKA for the new RAV4s which don't use but could? https://github.com/commaai/openpilot/pull/29978

jyoung8607: It's probably possible to switch to LTA, the SecOC part should just carry over. However, it will take multiple round-trips with test owners to plot how the stock camera drives the two LTA messages in tandem and then replicate, test, and validate safety for that behavior.

I don't have a preference on whether to make the switch, the openpilot team knows more depth and breadth of Toyota history than I do. If we do want to make the switch, my preference is to ship what we have and create an issue to track migrating to LTA.

pd0wm: > If I recall correctly the stock system uses the LTA for lane keep steering, but the torque message works fine too. So it's like most TSS2 cars in this regard.

Sorry, forgot lane keep is what you should call the safety nudges to put you back in the lane, not lane centering. I did *not* check what the stock system uses for LKA.

jyoung8607: image

Verified the stock system actuates on both STEERING_LKA (torque) and STEERING_LTA_2 (angle) depending on conditions.

board/safety/safety_toyota.h
```diff
@@ -87,14 +98,31 @@ static void toyota_rx_hook(const CANPacket_t *to_push) {
     }
 
     // enter controls on rising edge of ACC, exit controls on ACC off
-    // exit controls on rising edge of gas press
-    if (addr == 0x1D2) {
-      // 5th bit is CRUISE_ACTIVE
```
sshane, line 92: keep these comments?

board/safety/safety_toyota.h
```diff
@@ -203,6 +225,19 @@ static bool toyota_tx_hook(const CANPacket_t *to_send) {
...
+      int lta_angle_msb = GET_BYTE(to_send, 2);  // STEERING_LTA_2.STEER_ANGLE_CMD (MSB)
+      int lta_angle_lsb = GET_BYTE(to_send, 3);  // STEERING_LTA_2.STEER_ANGLE_CMD (LSB)
+
+      bool actuation = lta_request || lta_request2 || (lta_angle_msb != 0) || (lta_angle_lsb != 0);
+      if (actuation) {
+        tx = false;
+      }
+    }
+
     // LTA angle steering check
```
sshane, line 241:
```suggestion
    // STEERING_LTA angle steering check
```

sshane (commented):
Looks much cleaner!

board/safety/safety_toyota.h
jyoung8607: The "rising edge of gas press" comment was still there. I added signal annotations for each signal inside the block.

The same brake-pressed naming was used for three different messages, so I had to annotate the DBC as well there.

PR #2048, opened 2024-09-27
move safety to opendbc

maxime-desroches:
need to fix misra now


maxime-desroches:
Let's revisit this later when we have more mutation coverage and panda in a more stable state

PR #2049, opened 2024-09-28
unittest -> pytest

bongbui321:
resolve https://github.com/commaai/openpilot/issues/32941


tests/usbprotocol/test_comms.py
```diff
@@ -121,9 +123,10 @@ def test_can_send_usb(self):
           while lpp.can_pop(TX_QUEUES[bus], pkt):
             queue_msgs.append(unpackage_can_msg(pkt))
 
-          self.assertEqual(len(queue_msgs), len(msgs))
-          self.assertEqual(queue_msgs, msgs)
+          assert len(queue_msgs) == len(msgs)
+          assert queue_msgs == msgs
 
+  @pytest.mark.order(2)
```
maxime-desroches, line 129: Why do we need this?

bongbui321:
The three remaining points that you pointed out can be explained by this:

in the test_comms.py, we only have one instance of [libpanda](https://github.com/commaai/panda/blob/08c95bf47ba6d0c5d11a7616e42e10b8dbde19eb/tests/usbprotocol/test_comms.py#L8), and all the tests in that file share that one instance. since we are using multithreading, we have 6 tests in total for usbprotocol and only 4 runners, there is a high chance that while running one test, it got switched out to another one while the state of the switched out test remains (same libpanda). This explains why we need pytest-order (to ensure completion of one test before executing the next one).

One failure instance of running without pytest-order is that while executing `reset_rx` test case it got switched out and test `receive_usb` is run instead, this will create a failed test, since  now the `receive usb` test now reading the rx_q data from the `reset_rx`

The same apply for this:
> Why do we need to set a different safety mode here?

Since when we used to use unitttest -> send_usb is run first before the reset_tx so the safety tx hook is already defined from the send_usb test so you don't need to redefine it in reset_tx anymore (you can confirm this by [disable](https://github.com/bongbui321/panda/actions/runs/11127156136/job/30918807052) running send_usb test) and run the reset _tx only and that test would fail. pytest doesn't follow that test execution order anymore and it might run the reset_tx first, so without the safety hook definition, then it wouldn't get any tx messages

Happy to clarify more if that's not clear

adeebshihadeh:
We either need to:
1. reset all state such that the order doesn't matter
2. fix this so it can run in parallel with their own instances

Ideally, we can do #2.

adeebshihadeh:
Closing due to inactivity.

PR #2050, opened 2024-10-01
honda safety: fix mutations failures

maxime-desroches:
Currently, this line https://github.com/commaai/panda/blob/08c95bf47ba6d0c5d11a7616e42e10b8dbde19eb/board/safety/safety_honda.h#L281 
can change without the tests catching it


sshane:
These are all good to get in, can you open PRs?

PR #2052, opened 2024-10-02
CAN print improvement

dzid26:
- print all busses by default
- add colors for can_health
- scripts panda/jungle

### DIFF
--- a/board/jungle/scripts/can_health.py
+++ b/board/jungle/scripts/can_health.py
@@ -1,14 +1,29 @@
 #!/usr/bin/env python3
-
 import time
+import re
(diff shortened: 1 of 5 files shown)


board/jungle/scripts/can_health.py
```diff
@@ -1,13 +1,29 @@
...
-
 import time
+import re
 from panda import PandaJungle
 
+RED = '\033[91m'
+GREEN = '\033[92m'
+
+def colorize_errors(value):
+    if isinstance(value, str):
```
maxime-desroches, line 10: We generally use 2 spaces

PR #2057, opened 2024-10-02
mutation tests: make coverage gradual 

maxime-desroches:
Here is my proposal for making the mutation tests coverage increase over time while still being practical.

We have 2 testing modes that are run in parallel in CI:

1. DIFF_COVERAGE. This mode verifies mutations with `test_xx.py` (all of them) on the intersection between its code coverage and the current git diff. This mode is very powerful since it will cover basically all the files if they are changed and will still run super fast.

2. SAFETY_ONLY. This mode verifies mutations on `safety_xx.h` with `test_xx.py` if safety_xx.h is in `MUTATION_SAFETY_FILES`. Overtime, the goal is to add all the safety modes in that list. We can split those nicely, add them one by one and even put them on bounties.

Regardless of the testing mode, we are using all the ops in the `MUTATION_OPS` list to mutate the code. Overtime, the goal is to add  [most ops listed here](https://mull.readthedocs.io/en/latest/SupportedMutations.html) in that list.

In the end, we can increase coverage in 3 ways: add files to `MUTATION_SAFETY_FILES`, add ops to `MUTATION_OPS` and add a third testing mode for other files we want to always check.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -118,7 +118,7 @@ jobs:
         run: eval "$BUILD"
       - name: Mutation tests
         timeout-minutes: 5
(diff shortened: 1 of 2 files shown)


sshane:

PR #2060, opened 2024-10-05
Remove retry loop 

lukasloetkolben:
Current implementation could potentially result in an endless loop, which cannot be caught or handled from outside the function.

### DIFF
--- a/python/__init__.py
+++ b/python/__init__.py
@@ -831,18 +831,13 @@ class Panda:
   @ensure_can_packet_version
   def can_send_many(self, arr, timeout=CAN_SEND_TIMEOUT_MS):
     snds = pack_can_buffer(arr)
-    while True:
-      try:
-        for tx in snds:
-          while True:
-            bs = self._handle.bulkWrite(3, tx, timeout=timeout)
-            tx = tx[bs:]
-            if len(tx) == 0:
-              break
-            logger.error("CAN: PARTIAL SEND MANY, RETRYING")
-        break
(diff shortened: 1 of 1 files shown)


python/__init__.py
```diff
@@ -831,18 +831,13 @@ def can_reset_communications(self):
...
-        break
-      except (usb1.USBErrorIO, usb1.USBErrorOverflow):
-        logger.error("CAN: BAD SEND MANY, RETRYING")
+    for tx in snds:
+      while True:
+        bs = self._handle.bulkWrite(3, tx, timeout=timeout)
+        tx = tx[bs:]
+        if len(tx) == 0:
+          break
+        logger.error("CAN: PARTIAL SEND MANY, RETRYING")
```
adeebshihadeh, line 840: looks like this should be removed?

python/__init__.py
```diff
@@ -831,18 +831,13 @@ def can_reset_communications(self):
...
-            logger.error("CAN: PARTIAL SEND MANY, RETRYING")
-        break
-      except (usb1.USBErrorIO, usb1.USBErrorOverflow):
-        logger.error("CAN: BAD SEND MANY, RETRYING")
+    for tx in snds:
+      while True:
+        bs = self._handle.bulkWrite(3, tx, timeout=timeout)
+        tx = tx[bs:]
+        if len(tx) == 0:
+          break
```
adeebshihadeh, line 839: I'd just put this in the while condition

adeebshihadeh (changes requested):

PR #2064, opened 2024-10-19
ford - reset steering angle state when driver stops overriding steering

pfeiferj:
Currently if the driver manually steers faster than the angle rate limits for Ford (and others) then openpilot will not be able to resume from the current curvature/steering angle until the rate limit catches up to the actual state.

This resets the steering angle with the previous measured angle upon detecting the user has stopped overriding. This means that rate limit clips will be centered on the currently measured angle for the first frame after done overriding.


sshane:
I don't know if we want this in panda, you can achieve something similar by making latActive go low under high angle rate events in openpilot.

PR #2065, opened 2024-10-21
Ford: higher curvature rate limits at low speed

sshane:
Estimating the max EPS allowed rate using some old plots from last year:

image

image


sshane:
Same plot but expanded:

image

PR #2061, opened 2024-10-18
Toyota: Add SECOC 

chrispypatt:
Creates a SECOC variant of common long messages. This is basically a copy of the `TOYOTA_COMMON_LONG_TX_MSGS` but:
1. swapped `TOYOTA_COMMON_TX_MSGS` for `TOYOTA_COMMON_SECOC_TX_MSGS` 
2. added a the new secoc acc command address 0x183

This is still a WIP. Still need to:
- [ ] Verify all the addresses in `TOYOTA_COMMON_SECOC_TX_MSGS` are needed
- [ ] Testing to verify the change
- [ ] Anything else? Lmk

Accompanies: https://github.com/commaai/opendbc/pull/1385

PR #2072, opened 2024-11-01
Toyota: SecOC longitudinal control safety

jyoung8607:
Complement to commaai/opendbc#1385, intended to progress (not necessarily replace) commaai/panda#2061.

- [x] opendbc needs a corresponding change to send inactive accel in ACC_CONTROL
- [ ] common test override is lame, can we do better?
- [x] do we actually need the radar diagnostic address when we're able to MITM?
  * Radar diagnostic is permitted for all Toyota long right now, should scope better but not in this PR


sshane:

PR #2085, opened 2024-11-21
Control over FD flag of CAN messages

pd0wm:
Currently once a CAN-FD message is seen on a bus, all future messages are sent out as FD. This probably made porting the first CAN-FD cars easier, as openpilot does need to know about this.

However, with Toyota CAN-FD cars I've already ran into an issue where diagnostic messages are only accepted if they are non-FD. Also in future car ports it might be necessary to mix FD and non-FD messages.

In this PR I'm suggesting to use the reserved bit in the USB/SPI message to set the FD state of an incoming/outgoing message. I've also added an endpoint to enable the current behavior to make a smooth migration possible, when this is merged boardd should probably set this.

My questions:
- Are you interested in adding functionality like this?
- The FD flag is probably the most necessary to add, and does not require making the packet larger. But if you want to be 100% correct you also want to add flags for BRS and RTR. Is this something you'd like to add if we're messing with the `CANPacket_t` anyway?
- I've added the `fd` argument to `can_send`/`can_send_many`, but it's probably cleaner to create something like a dataclass to hold a CAN frame. If you want that I can do that in a separate PR first, but that would require many changes.

Behavior verified with a Red Panda and PeakCAN SocketCAN. But should probably tested together with the indented openpilot change before merging. Otherwise this PR is ready to merge.

### DIFF
--- a/board/can_declarations.h
+++ b/board/can_declarations.h
@@ -13,7 +13,7 @@
 #endif
 
 typedef struct {
-  unsigned char reserved : 1;
+  unsigned char fd : 1;
   unsigned char bus : 3;
   unsigned char data_len_code : 4;  // lookup length with dlc_to_len
   unsigned char rejected : 1;
--- a/board/drivers/bxcan.h
+++ b/board/drivers/bxcan.h
@@ -89,6 +89,7 @@ void process_can(uint8_t can_number) {
       if ((CANx->TSR & CAN_TSR_RQCP0) == CAN_TSR_RQCP0) {
         if ((CANx->TSR & CAN_TSR_TXOK0) == CAN_TSR_TXOK0) {
           CANPacket_t to_push;
+          to_push.fd = 0U;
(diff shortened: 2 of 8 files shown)


adeebshihadeh:
> * Are you interested in adding functionality like this?

Always interested in anything to make more upstream-able cars supported. We'll always find a clean.

> * The FD flag is probably the most necessary to add, and does not require making the packet larger. But if you want to be 100% correct you also want to add flags for BRS and RTR. Is this something you'd like to add if we're messing with the `CANPacket_t` anyway?

I think we'd rather not. Let's just leave a good comment and revisit if it becomes relevant later?

> * I've added the `fd` argument to `can_send`/`can_send_many`, but it's probably cleaner to create something like a dataclass to hold a CAN frame. If you want that I can do that in a separate PR first, but that would require many changes.

Yeah, it's probably time we got away from the tuple. I'm fine with the flag for now though.

PR #2087, opened 2024-11-22
Ensure auto switching FD is off in Python

pd0wm:
Disable auto switching (= default state) on connecting using the Python API. This state is persistent, and therefore may have been left in the enabled state by openpilot.


adeebshihadeh:
I suspect this may cause issues somewhere, but I'll merge for now and revert if it does.

PR #2096, opened 2024-12-06
test: Allow running test_misra.sh easy on macos

devtekve:
simple fix to be able to run misra with macos's default bash and it should have no impact on linux / unix


devtekve:
_@adeebshihadeh you might be interested on this one (macOS fixes)_

PR #2100, opened 2024-12-10
cuatro: remove beeper

adeebshihadeh:


dzid26:
Is the siren going to be removed entirely or is OP going to play the sound?
Either way, seems like this would worsen the safety model, more or less.

PR #2106, opened 2024-12-13
GPIO: Explicitly casts constants to `uint32_t`

sunnyhaibin:
**Description**

Explicitly casts constants (`3U` and `pin * 2U`) to `uint32_t` to prevent type promotion or mismatches.

Noticed this MISRA violation when we were working on sunnypilot's panda using upstream's CI: https://github.com/sunnypilot/panda/actions/runs/12314225225/job/34369895304

image


adeebshihadeh:
Why doesn't this fail on ours?

devtekve:
> Why doesn't this fail on ours?

That's the million dollar question. It started happening to us on the pipeline when adding a new header and one function call on `safety.h`. It's so strange because we did not touch anywhere near gpio.h and yet MISRA decided to go for a validation run on that path. 

Maybe something on the static analyzer made with an optimization or something decided to go that route. But this is very odd. However, from what I can tell, the warning is a valid one. So not sure why it wasn't flagged already when MISRA was introduced

PR #2121, opened 2025-01-13
remove uds from panda

sshane:
For https://github.com/commaai/opendbc/pull/1598


pd0wm:
What was the exact reason to remove this from panda? Are you going to move over `isotp.py`, `xcp.py` and `ccp.py` al well to keep all car-protocol related code together?

I'm a bit sad about how many of my scripts will break with import errors now, but that shouldn't stop you from putting things in a more logical place for the future 😄

adeebshihadeh:
The goal is to make opendbc the only repo you need to touch for a car port (DBC, interface code, safety, etc.). We plan to have everything settled before the hackathon next month, so you might want to wait a bit before fixing stuff up.

pd0wm:
Would you ever need to touch `uds.py` for a car port though? But happy to wait to see what else you have planned to clean up to make car porting easier.

I had some thoughts on putting CI around the UDS stuff now that greg upstreamed the SocketPanda stuff. But guess that can still be done in the opendbc repo.

PR #2128, opened 2025-01-22
Ford: long safety is default for CAN

sshane:
For https://github.com/commaai/openpilot/issues/33944

### DIFF
--- a/tests/safety/test_ford.py
+++ b/tests/safety/test_ford.py
@@ -60,7 +60,6 @@ class Buttons:
 
 
 # Ford safety has four different configurations tested here:
-#  * CAN with stock longitudinal
 #  * CAN with openpilot longitudinal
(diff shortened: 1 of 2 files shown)


tests/safety/test_ford.py
```diff
@@ -457,7 +441,8 @@ class TestFordLongitudinalSafety(TestFordLongitudinalSafetyBase):
   def setUp(self):
     self.packer = CANPackerPanda("ford_lincoln_base_pt")
     self.safety = libpanda_py.libpanda
-    self.safety.set_safety_hooks(Panda.SAFETY_FORD, Panda.FLAG_FORD_LONG_CONTROL)
+    # Makes sure we enforce long safety even without long flag for CAN
```
sshane, line 444:
```suggestion
    # Make sure we enforce long safety even without long flag for CAN
```

PR #2130, opened 2025-01-24
Add socketcan and python-can support to panda

ccdunder:
This provides a bidi connection between Panda and socketcan (and/or python-can).

My initial motivation was to use some opensource tools for discovery, fuzzing, and analysis of higher level CAN msg protocols. The existing socketcan driver doesn't support SPI Panda, and this is 1/4 the size of a kernel driver.

I'm opening this as a draft to solicit early feedback. If there's interest, I'll add cmdline args and threaded msg passing before moving out of draft.

### DIFF
--- a/python/__init__.py
+++ b/python/__init__.py
@@ -9,7 +9,7 @@ import binascii
 from functools import wraps, partial
 from itertools import accumulate
 
-from .base import BaseHandle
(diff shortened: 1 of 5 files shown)


adeebshihadeh:
Closing due to inactivity, feel free to reopen when it's ready

PR #2131, opened 2025-01-29
bxcan: we don't reset can core

sshane:


sshane:
nvm, we do consider it a can core reset: https://github.com/commaai/panda/commit/a35e9a1edb1e685f00f4dd9d874043f56ea771fc

PR #2134, opened 2025-01-29
fdcan: don't reset CAN cores w/ disabled transceivers

sshane:


sshane:
This won't work. After ignition off, the transmit error counter builds and our new check will prevent the CAN core from being reset, however once ignition is back, the messages in the STM buffer will be one by one sent to the CAN core after each reset.

image

Issue #2132, opened 2025-01-29
fdcan: high number of CAN core resets w/ ACK error


sshane:
The route in https://github.com/commaai/openpilot/issues/33840 had its camera bus spam CAN core resets due to ACK errors and a high transmit error count on ignition off.

This PR https://github.com/commaai/panda/pull/1502 switched the logic from only resetting once when an error counter reached 100 to resetting continually while the transmit_error_counter > 127. Once this happens, any future ACK errors no longer need this tolerance to reset the CAN core. This causes the interrupt load of the panda to hover around 90%, slowing down/hanging SPI communication with pandad.

Things that mitigate this:
- removing this busywait delay significantly lowers the interrupt load and communication is kept: https://github.com/commaai/panda/blob/12905886bafb9bf3fbe7200fa632a5501b1e1cb5/board/stm32h7/llfdcan.h#L18
- not sure why, but keeping both CAN 1 and CAN 3 up also keep SPI communication happy (now we only keep one): https://github.com/commaai/panda/blob/12905886bafb9bf3fbe7200fa632a5501b1e1cb5/board/power_saving.h#L36
- ~Giving the CAN irqs priorities of 1 and SPI1 and SPI2 (setting SPI1_IRQn and SPI2_IRQn irqs to -1 that)~: https://github.com/commaai/panda/blob/12905886bafb9bf3fbe7200fa632a5501b1e1cb5/board/stm32h7/llfdcan.h#L135-L142
- debounce CAN core resets
- don't reset CAN cores when their transceivers are off due to ignition/power saving mode.
- reset transit error counter

TODOs:
- [ ] This should be common, figure out why this is rare. On another random Bronco route I see 2 CAN core resets on camera bus (flipped canState0) due to 100 busOffCnts. Also figure out how it didn't spam CAN core resets in this case. Figuring out these two questions should make it clearer which fix is appropriate.

Related PRs that touch this code:
- https://github.com/commaai/panda/pull/1431
- https://github.com/commaai/panda/pull/1502
- https://github.com/commaai/panda/pull/1967

PR #2137, opened 2025-01-30
fdcan: rate limit CAN core reset

sshane:
Closes https://github.com/commaai/panda/issues/2132 and https://github.com/commaai/openpilot/issues/33840.

Ignition off w/ canState1 and canState2 active receiving messages w/ forced resetting on canState0 (bus 2, similar case to right after ignition off in https://github.com/commaai/openpilot/issues/33840):

- No resetting: 0.13 interrupt load
- Resetting at 10Hz: 0.15 interrupt load (**this PR**)
- Resetting at 100Hz: 0.35 interrupt load
- Resetting w/o restrictions (~286Hz): 0.8 interrupt load (**similar to issue**)

### DIFF
--- a/board/drivers/fdcan.h
+++ b/board/drivers/fdcan.h
@@ -23,7 +23,7 @@ void can_clear_send(FDCAN_GlobalTypeDef *FDCANx, uint8_t can_number) {
   uint32_t time = microsecond_timer_get();
 
   // Resetting CAN core is a slow blocking operation, limit frequency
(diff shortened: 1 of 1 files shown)


sshane:
No issues on ignition off with some transmit errors!

image

Before, we would get stuck resetting in a loop for ~500 resets: https://github.com/commaai/panda/pull/2134#issuecomment-2623440427

adeebshihadeh (approved):

PR #2116, opened 2024-12-17
Remove car specific flags

nworb-cire:
Clean up after https://github.com/commaai/panda/pull/2115 and https://github.com/commaai/opendbc/pull/1589 are merged

PR #2142, opened 2025-02-11
add opendbc as proper dependency

sshane:
enabled by https://github.com/commaai/panda/pull/2116


sshane:
we need CANPacker with is cython/c++, so cannot merge yet

PR #2150, opened 2025-02-13
Remove hw type from safety tests

sshane:


sshane:
should keep this for the test_comms.py test, so going to remove it for libsafety in https://github.com/commaai/panda/pull/2149 instead

PR #2154, opened 2025-02-17
safety: don't silently skip classes with missing abstract methods

sshane:
@adeebshihadeh did you know pytest does this?? unittest.main() catches these and throws a TypeError, but pytest just skips the class. can't believe this hasn't happened yet, just hit it in: https://github.com/commaai/panda/pull/2036

```
TypeError: Can't instantiate abstract class TestTeslaLongitudinalSafety without an implementation for abstract method '_accel_msg'
```

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -1,0 +1,9 @@
+import pytest
+import unittest
+
+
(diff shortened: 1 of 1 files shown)


sshane:

PR #2157, opened 2025-02-20
Rivian: add CAN ignition

lukasloetkolben:
We should verify that again. 
`ESPiB1_IgnitionOn` also sounds good.
I can't remember why I used this signal back then.


sshane:
This message exists on a few Subaru Pre-global, however the bits are always zero. https://github.com/commaai/panda/issues/2164 should make this less sketchy

Issue #2171, opened 2025-03-13
[$500 Bounty] Refactor panda to use source + header file structure


robbederks:
It's time we move the repo from the current style (everything is a header file) to having separate source and header files. This should clean up all the import orders and `*_definitions.h` files, as well as make it a lot easier to add new drivers etc.

Bounty for a clean PR with minimal changes.

robbederks:
Locked to #2172 for now.

aubsw:
I spent some time trying to break things down into smaller PRs, but now I'm not so sure how feasible thi will be because I'm struggling with the MISRA checks.

Here's what we have right now:

1. https://github.com/commaai/opendbc/pull/2076 which the pre-factor to opendbc to make our changes to panda happy
2. https://github.com/commaai/panda/pull/2175 the whole panda change in one PR

I need to write up some more notes on the PR description themselves.

@robbederks do you have any guidance on how much I should care about he MISRA checks? Are there any kinds of checks I can fudge?

aubsw:
@robbederks Here's where we stand:

- I was able to make do the refactor without https://github.com/commaai/panda/pull/2181, but it did require changes to opendbc. 

   Crucially, we need to put matching guards on can.h and utils.h, as these files provide conflicting definitions at compile time. We are forced to address the namespace conflict in these declarations as long as we need to link safety.h from opendbc, i.e.

```diff
diff --git a/opendbc/safety/board/can_declarations.h b/opendbc/safety/board/can_declarations.h
index 3824f01..589d466 100644
--- a/opendbc/safety/board/can_declarations.h
+++ b/opendbc/safety/board/can_declarations.h
@@ -1,4 +1,5 @@
-#pragma once
+#ifndef CAN_DEFINITION_H
+#define CAN_DEFINITION_H
 
 #define CANPACKET_HEAD_SIZE 6U
 
@@ -24,3 +25,6 @@ typedef struct {
 
 #define GET_BUS(msg) ((msg)->bus)
 #define GET_ADDR(msg) ((msg)->addr)
+
+#endif // CAN_DEFINITION_H
+
diff --git a/opendbc/safety/board/utils.h b/opendbc/safety/board/utils.h
index 9bb9b64..985df67 100644
--- a/opendbc/safety/board/utils.h
+++ b/opendbc/safety/board/utils.h
@@ -1,3 +1,5 @@
+#ifndef UTILS_DEFINITION_H
+#define UTILS_DEFINITION_H
 #pragma once
[...]

PR #2172, opened 2025-03-13
Refactor panda

savojovic:
Is this the kind of refactor you had in mind? If so, would you consider this a minimal change? Note that this change is just for proof of concept. If folder structure is good, it should be added to the build file so linker can find new files.
@robbederks


robbederks:
Yes, similar to this.

I would prefer the header files to either be kept in the same folder, or to mimic the source code tree inside of `include/` (similar to linux), whatever is the cleanest.

To claim the bounty, obviously everything needs to be moved over and panda / openpilot CI needs to pass.

robbederks:
(linking to bounty issue #2171)

savojovic:
Great, I'll see what would be the cleanest approach and see how it goes.

savojovic:
Pushing progress. CI is expected to fail for now, and resolving conflicts is unnecessary since more will arise before completion.

robbederks:
Good progress so far! Let's maybe not put the .c files in `src/` folders, unless there's a good reason for it?
Will lock the bounty to you for the time being.

savojovic:
Sounds good!
There is no particular reason for the src/ folder. I'll refactor it out.

savojovic:
Major refactoring should be done. I'm going step by step and fixing build issues caused by implicit inclusions. At this moment panda/board/boards folder is successfully being built.

SConscript
```diff
@@ -66,69 +66,95 @@ def build_project(project_name, project, extra_flags):
...
-    "-Wstrict-prototypes",
-    "-Werror",
-    "-mlittle-endian",
-    "-mthumb",
-    "-nostdlib",
-    "-fno-builtin",
-    "-std=gnu11",
-    "-fmax-errors=1",
-    f"-T{linkerscript_fn}",
+      "-Wall",
```
robbederks, line 69: let's make sure we don't have unnecessary indentation changes like this (we use 2 spaces)

savojovic: Forgot to change indentation, reverted changes.

board/boards/board_dos.c
```diff
@@ -1,6 +1,6 @@
-#pragma once
-
+#include "unused_funcs.h"
```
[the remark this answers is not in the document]
savojovic: Yes, we can. Reverted.

savojovic:
This is already taking more time than I expected it to. I'll try to finalize this by the end of the week.

savojovic:
I hoped that going for bulk refactoring that the dependencies would solve themselves. But this problem requires systematic approach and probably going one by one build.
Let's close this and give others a chance. Maybe someone has a better idea.

PR #2174, opened 2025-03-17
Consistent power measurements

robbederks:
Replaces the one-shot ADC readouts with a continuous readout that is low-pass filtered.
- pre-PR: current stdev ~17mA, voltage stdev ~10mV
<s>- post-PR: current stdev ~8mA, voltage stdev ~4mV</s> TODO: update measurements

Also implements VDDA reference voltage calibration based on VREFINT and its factory calibration value.
- pre-PR bias: > 2%
- post-PR bias: < 0.5%


robbederks:
Don't think we need it to run continuously. Splitting out the VDDA calibration in a new PR (#2247)

PR #2175, opened 2025-03-18
Leave circular dependency hell (adopt source/header model)

aubsw:
This is to address https://github.com/commaai/panda/issues/2171

Unfortunately, for this to work in the end, I think we need to do the same to opendbc, since that library being header-only breaks our build. 

For local testing, I patched opendbc, and it was a fairly small patch.


aubsw:
@robbederks I think for this work, we will need source-and-header-ify https://github.com/commaai/opendbc/tree/master/opendbc/safety, since we would need to #include those headers here.

I got build working locally by patching safety.h & co.

I'm also happy to try to break this work up in to several smaller PRs, tho that may take some time because it's difficult to  break out individual headers without generating conflicts at the linking stage.

robbederks:
Thanks. For now, the bounty is locked to #2172, if that attempt stalls it'll be unlocked again.

robbederks:
@aubsw the bounty is unlocked if you want to continue!

aubsw:
@robbederks I'm prototyping my merge plan and am a bit behind. I will follow-up on the merge plan with a clear head tomorrow morning.

SConscript
```diff
@@ -111,14 +119,127 @@ def build_project(project_name, project, extra_flags):
     env.Object(f"rsa-{project_name}", f"{panda_root}/crypto/rsa.c"),
```
robbederks, line 119: let's also re-use `make_object` for the existing ones here

board/bootstub.c
```diff
@@ -1,10 +1,18 @@
-#define BOOTSTUB
```
robbederks, line 1: can't immediately find where this gets defined again, it's still used in a bunch of `ifdef`s

SConscript
```diff
@@ -83,8 +83,16 @@ def build_project(project_name, project, extra_flags):
     '.',
     '..',
     panda_root,
-    f"{panda_root}/board/",
-    f"{panda_root}/../opendbc/safety/",
+    f"{panda_root}/../opendbc",
+    f"{panda_root}/../opendbc/safety",
+    f"{panda_root}/include/",
```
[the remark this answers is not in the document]
aubsw: Suprisingly, niether GCC nor SCons provides recursive directory includes. We have a few options:

1. leave it how it is
2. write a helper function to recursively get search paths
3. add only top-level directories to the search path, adjust all header includes to use the full path. I.e. `#include "registers.h"` would become `#include "drivers/registers.h"` everywhere.

I kind of prefer option #3 because it clarifies the directory structure of included headers in the source code.

SConscript
```diff
@@ -111,14 +119,127 @@ def build_project(project_name, project, extra_flags):
...
+      if not _is_panda_jungle:
+          sources.extend([
+            ("stm32f4_llfan", f"{panda_root}/board/stm32f4/llfan.c"),
+          ])
+
+  # Create a bootstub-specific environment with -DBOOTSTUB flag
+  bootstub_env = env.Clone()
+  bootstub_env.Append(CFLAGS=["-DBOOTSTUB"])
+  bootstub_env.Append(ASFLAGS=["-DBOOTSTUB"])
+  bootstub_env.Append(LINKFLAGS=["-DBOOTSTUB"])
```
aubsw, line 192: @robbederks I actually sneakily moved the BOOTSTUB define from the impl to here... Is this a bad idea?

board/bootstub.c
aubsw: See L188 in the root SConscript–lmk what you think.

SConscript
aubsw: I actually just remembered why I did this: we now need to generate seperate objects for bootstub. Hence, we add BOOTSTUB to their compilation environment, since they're not guaranteed to get bootstub.c anymore.

robbederks: should be fine! just couldn't immediately find it.

PR #2180, opened 2025-04-03
Untangle config dependencies, misc prefactors

aubsw:
A pre-factor for https://github.com/commaai/panda/issues/2171 in the interest of reducing diff size.

# Config files refactor

We split the `board/config.h` into the following headers:

- `include/board/stm_config.h`: declarations + headers that depend on which STM microcontroller we're targetting.
- `include/board/platform_definitions.h`: declarations ONLY. To be included in source files that need to know about STM-dependent definitions, but don't need to know about all existing headers (IWYU).
- `include/board/board_config.h`: global value declarations

There will be additional de-spaghetti-ification in subsequent PRs.

# Misc refactors
- Add header guard to utils.h + some tweaks to ensure compatibility with opendebc (we must land this PR before https://github.com/commaai/opendbc/pull/2076)
- Add declarations for gpio.h
- Rework can_common declarations a bit


board/drivers/gpio.h
```diff
@@ -68,12 +55,14 @@ int get_gpio_input(const GPIO_TypeDef *GPIO, unsigned int pin) {
   return (GPIO->IDR & (1UL << pin)) == (1UL << pin);
 }
 
+// cppcheck-suppress[misra-c2012-9.3,misra-c2012-8.7]
```
robbederks, line 58: try not to suppress any misra errors like these, let's fix them instead

aubsw:
> why is this needed exactly? 

I did this way because when I first started working this, it was the only way to unambiguously break the circular dependency chain.

I disgree with you that this is adding complexity–I think that this is a strict reduction in complexity overall because it flattens the dependency heirarchy. You also need to keep track of less state when reading the code this way.

> can we keep https://github.com/commaai/panda/pull/2175 to a minimal diff to just get everything building

Definitely a fair consideration.

 At the very least we need to keep a dedicated header whose purpose is to  plex on `stm32h7xx.h`, `stm32h7xx_hal_gpio_ex.h`, etc, because this is currently only exposed in config.h, which causes circular depedencies. I think this is hopefully a fair compromise. Working on proving that this works!

aubsw:
Closing in favor of https://github.com/commaai/panda/pull/2185

PR #2181, opened 2025-04-04
Dedupe opendbc code

aubsw:
Needed for https://github.com/commaai/panda/issues/2171

Dedupes the code in `panda` that was cribbed from `opendbc`.

**details:**

1. As a convention, we now include headers from `opendbc` using with the relative path from `opendbc`. 
    E.g. `#include "safety.h"` now becomes `#include "safety/safety.h"` 
2. Delete duplicated code.
3. Fix MISRA test invocation (now it can see opendbc headers).

**why?** 

1. Conflicts

    The duplicated code was causing issues during compilation when trying to set up the source/header model for #2171. 
    
    The duplicated macros defined at `opendbc/safety/board/utils.h` and `board/utils.h` were confusing the compiler during preprocessing, e.g.:
 

> ```zsh
> panda % arm-none-eabi-gcc -o board/bootstubx-safety-panda.o -c -mcpu=cortex-m4 -mhard-float -DSTM32F4 -DSTM32F413xx -Iboard/stm32f4/inc # [... truncated for brevity]
> In file included from /Users/aub/programming/panda-cpy/opendbc/safety/safety.h:6,
>                  from /Users/aub/programming/panda-cpy/opendbc/safety/main.c:1:
> /Users/aub/programming/panda-cpy/opendbc/safety/board/utils.h:38:21: error: expected identifier or '(' before 'void'
>    38 | #define UNUSED(x) ((void)(x))
>       |                     ^~~~
> /Users/aub/programming/panda-cpy/opendbc/safety/main.c:6:1: note: in expansion of macro 'UNUSED'
>     6 | UNUSED(heartbeat_engaged);
>       | ^~~~~~
> ```
>  

I don't think there's a sensible work-around aside from deleting the duplicated macros outright.

 
 2. Comprehensibility
 
     The change in import paths for headers clarifies where the code is coming from. May help protect us against namespace collisions in the future.


aubsw:
@robbederks WDYT of this approach as a first step? This help shrink the diff in subsequent PRs and also makes things a bit easier to reason about, IMO.

I still need to figure out the MISRA checks.

robbederks:
Splitting up the code like this is pretty ugly (especially things like the FAULT list being split over the two repos). The duplicated code we have now in panda / opendbc is obviously bad already, but this seems like a step in the wrong direction to me. Requires a bit more thought / refactoring to abstract away cleanly I think (@sshane any ideas?)

Any way we can finish your refactor without these changes in the meantime?

PR #2186, opened 2025-04-11
Delete duplicated code from panda

aubsw:
- Must be merged after https://github.com/commaai/opendbc/pull/2076
- This PR deletes duplicated code from opendbc. 

# details

- `utils.h` and `can.h` are approximately duplicated from opendbc, which is making https://github.com/commaai/panda/issues/2171 hard to land because it results in duplicated symbols
- since opendbc has not yet adopted source/header style, we can't include many headers from opendbc in panda without getting multiple defintions of the same symbols (compilation errors)


robbederks:
Diff looks good to me, can merge once the tests are fixed. I've bumped the opendbc ref too.

tests/misra/test_misra.sh
```diff
@@ -64,12 +72,16 @@ cppcheck() {
 }
 
 PANDA_OPTS="--enable=all --disable=unusedFunction -DPANDA --addon=misra"
+PANDA_SOURCES="$PANDA_DIR/board/ --enable=all --file-filter=*.c \
+         -i $PANDA_DIR/board/safety.c \
+         -i $PANDA_DIR/board/jungle/main.c \
+         -i $PANDA_DIR/board/bootstub.c"
```
aubsw, line 78: @robbederks should we ignore all non-main.c files? 

If we don't ignore, we'll have a lot more cppchecks failures. The old behavior was to only check `board/main.c`.

board/flasher.h
```diff
@@ -130,18 +130,22 @@ void soft_flasher_start(void) {
 
   flasher_peripherals_init();
 
+#if defined(ENABLE_SPI) || defined(BOOTSTUB)
```
aubsw, line 133: @robbederks is this what we should do? or should be just suppress unused function check?

board/drivers/watchdog.h
```diff
@@ -7,6 +7,7 @@ static void watchdog_feed(void) {
   IND_WDG->KR = 0xAAAAU;
 }
 
+// cppcheck-suppress [unusedFunction] - No clue why this exists.
```
aubsw, line 10: @robbederks what should we do about this?

aubsw (commented):
@robbederks not sure why, but these changes have exposed some new cppcheck failures. 

Mostly just conditionally unused functions (i.e. some functions only get used when ENABLE_SPI is true). Pls advise (see comments).

tests/misra/test_misra.sh
robbederks: Ideally we check everything I think, but we should especially check safety.c and main.c if that covers all of the panda app code

board/flasher.h
robbederks: If it's only in a few spots I like the ifdefs better

tests/misra/test_misra.sh
aubsw: Kk. Looking into it. It may require some more cahnges to opendbc.

aubsw:
Marked it a draft until https://github.com/commaai/panda/pull/2188 and https://github.com/commaai/panda/pull/2189 land.

tests/misra/test_misra.sh
aubsw: I cleared up `board/main.c` and `safety.c`. 

Currently both `bootstub.c` and `jungle/main.c` are in a MISRA-incompliant state. I think we should address those later for the sake of progress.

tests/misra/suppressions.txt
```diff
@@ -15,6 +15,14 @@ unmatchedSuppression
 # All interrupt handlers are defined, including ones we don't use
 unusedFunction:*/interrupt_handlers*.h
 
+# This header is generated by the SConscript, is working as intended.
+misra-c2012-8.9:*/gitversion.h
```
robbederks, line 19: why is this an issue now? don't think anything changed for `gitversion.h`?

robbederks:
trigger-jenkins

aubsw:
Anything for me to do regarding this failed hitl run? I don't think I have access to https://jenkins.comma.life/job/panda/job/tmp-jenkins-2186/2/display

robbederks:
Jungle v2 doesn't seem to boot with this PR (gets stuck resetting after flashing). Do you have any hardware to test this out on?

aubsw:
trigger-jenkins

aubsw:
trigger-jenkins

aubsw:
Sorry for the n00b question, but do you think I could debug this with a [blue pill stm board](https://www.amazon.com/ALMOCN-STM32F103C8T6-Development-Downloader-Programmer/dp/B0C9PKJF2T/ref=sr_1_2?crid=3UE4X5QZQOT7U&dib=eyJ2IjoiMSJ9.BFbJhaRttaM-fkO3E65hlnb5Hb77OYZcIbrl8ssKSUJ_wd60K1ZgMI9ic40aPbExNgOglbDfk_nB0QlYg_r4xCCsyzRw2_hSsr5dnE5WFV4LWozM0O6ZS3FSf4Rf5vDaRWtDW_9RqBhoIBlT8iA64dpR9svhuqrsEPrdfvowm3g.Kyh7YkNrfyXFfcy8o5RXc8GD1nzZtlP41-X8vWUXfDw&dib_tag=se&keywords=blue%2Bpill%2Bstm&qid=1745332561&sprefix=blue%2Bpill%2Bst%2Caps%2C190&sr=8-2&th=1)?

robbederks:
trigger-jenkins

robbederks:
Sorry, I'm OOO this week so reply time might be a bit slower

robbederks:
trigger-jenkins

aubsw:
@robbederks is the HITL test still failing?

aubsw:
Adeeb was unhappy with this PR in the dev call today, which is understandable as it's not the ideal solution.

I think that the ideal solution would probably be to make opendbc use source/header style so that we can easily link the safety code into panda.

aubsw:
OK I took another hard look. Found a waaay simpler way to do this step.

We may still need to make another PR to deal with linking safety code properly for the final PR, but if we can land this one first, that PR will be way less risky.

PR #2211, opened 2025-05-20
same size CANPacket_t across platforms

adeebshihadeh:


adeebshihadeh:
Ok maybe not...
```
region `RAM' overflowed by 164452 bytes
```

PR #2216, opened 2025-05-21
Add PWM brightness control for Cuatro LEDs

adeebshihadeh:


robbederks:
real implementation: https://github.com/commaai/panda/pull/2221

Issue #2105, opened 2024-12-13
cppcheck: catch ints as bools


adeebshihadeh:
Context in #1876

This is three steps:
1. add support for this to cppcheck
2. upstream that support to cppcheck
3. PR to enable it and fix all the violations

While we're waiting for them to merge the support upstream, you can open a draft PR using the patch.

anshgoyalevil:
Hey @adeebshihadeh I am willing to resolve this issue. Can you please assign it to me?

anshgoyalevil:
I have created a pull request for getting the required changes merged into upstream cppcheck. Lets see how it goes.
https://github.com/danmar/cppcheck/pull/7110

eckertliam:
@adeebshihadeh it is unclear to me whether this issue is still open/active. If it is I am willing to resolve this if you assign it to me.

B-L-A-N-K-official:
is this open ?

varshneydevansh:
@rbxict I am workign on this and that my upstream merge 
still have a bug which I found after runnning my changes.
and I have found the places which needs to be fix in the panda :)

varshneydevansh:
@adeebshihadeh I identified the remaining panda-side cppcheck cleanup after 
the upstream fixes

- https://github.com/danmar/cppcheck/pull/8340
- https://github.com/danmar/cppcheck/pull/8388 (still pending)

 and already made the change here:

https://github.com/varshneydevansh/panda/commit/286147eedc9da60a14228f470a4c42a1704d39d5

I’m blocked by GitHub fork/PR creation issues on `commaai/panda`. 
**If needed, could you open/apply this tiny panda patch on my behalf?**




```sh

 Devanshs-MacBook-Pro    \panda  ➜ (  zero_and_one)   25ms   3:55 PM   
 ⚡devanshvarshney ❯❯ source .venv/bin/activate
SKIP_BUILD=1 SKIP_TABLES_DIFF=1 CPPCHECK_DIR=/Users/devanshvarshney/cppcheck tests/misra/test_misra.sh


** PANDA H7 CODE **
Checking /Users/devanshvarshney/panda/board/main.c ...
/Users/devanshvarshney/panda/board/main.c:23:2: information: Include file: "board/obj/gitversion.h" not found. [missingInclude]
#include "board/obj/gitversion.h"
 ^
Checking /Users/devanshvarshney/panda/board/main.c: __GNUC__=9;STM32H7=1;STM32H725xx=1...
[...]

PR #2240, opened 2025-07-26
Fix cppcheck warning: Make 'status != 0' explicit in spidev_open

eshreyareddy:
### Summary
This PR resolves the cppcheck warning:
> style: Condition 'status' is always true [knownConditionTrueFalse]

### What I changed
- Replaced `if (status)` with `if (status != 0)` in `spidev_open()` to make the check explicit
- This removes false positives from static analysis tools like cppcheck

### Verification
- Ran `cppcheck --enable=all` locally
- The warning is now gone

Closes: commaai/panda#2105

### DIFF
--- a/drivers/spi/spidev_panda.c
+++ b/drivers/spi/spidev_panda.c
@@ -600,7 +600,7 @@ static int spidev_open(struct inode *inode, struct file *filp)
 		}
 	}
 
-	if (status) {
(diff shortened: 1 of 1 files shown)


adeebshihadeh:
We don't run cppcheck on drivers/ and don't think it's necessary.

PR #2252, opened 2025-08-03
Keep supporting White/Black pandas from Python side

pd0wm:
Noticed support for white and black pandas was dropped. Does it make sense to support them from the Python side only while they still work? I still use white pandas quite often with the UdsClient for interacting with ECUs on the bench. 

Feel free to close if you're not interested in supporting this anymore and/or have bigger changes planned on the panda side that would break support anyway. I guess red pandas are only $100 these days :smile:


pd0wm:
Thanks!

PR #2258, opened 2025-08-19
Reset bus_config defaults in can_init_all

adeebshihadeh:


board/drivers/can_common.h
```diff
@@ -130,18 +130,24 @@ void can_clear(can_ring *q) {
 
 // Helpers
 // Panda:       Bus 0=CAN1   Bus 1=CAN2   Bus 2=CAN3
-bus_config_t bus_config[BUS_CONFIG_ARRAY_SIZE] = {
-  { .bus_lookup = 0U, .can_num_lookup = 0U, .forwarding_bus = -1, .can_speed = 5000U, .can_data_speed = 20000U, .canfd_auto = false, .canfd_enabled = false, .brs_enabled = false, .canfd_non_iso = false },
-  { .bus_lookup = 1U, .can_num_lookup = 1U, .forwarding_bus = -1, .can_speed = 5000U, .can_data_speed = 20000U, .canfd_auto = false, .canfd_enabled = false, .brs_enabled = false, .canfd_non_iso = false },
-  { .bus_lookup = 2U, .can_num_lookup = 2U, .forwarding_bus = -1, .can_speed = 5000U, .can_data_speed = 20000U, .canfd_auto = false, .canfd_enabled = false, .brs_enabled = false, .canfd_non_iso = false },
-  { .bus_lookup = 0xFFU, .can_num_lookup = 0xFFU, .forwarding_bus = -1, .can_speed = 333U, .can_data_speed = 333U, .canfd_auto = false, .canfd_enabled = false, .brs_enabled = false, .canfd_non_iso = false },
```
jyoung8607, line 137: I'm unclear when bus_num/can_num would have been 0xFF. Perhaps it's a GMLAN remnant?

I ask before sending this out to users because:
* We're losing that bus_config entry in this PR
* There are still four checks for `can_number != 0xffU` in the current codebase
* Jenkins/HITL failed on this but I don't have visibility into why

board/drivers/can_common.h
```diff
@@ -130,18 +130,24 @@ void can_clear(can_ring *q) {
 
 // Helpers
 // Panda:       Bus 0=CAN1   Bus 1=CAN2   Bus 2=CAN3
-bus_config_t bus_config[BUS_CONFIG_ARRAY_SIZE] = {
```
jyoung8607, line 133: If discarding the 0xFF bus is correct, then `BUS_CONFIG_ARRAY_SIZE` may be removed in favor of `PANDA_CAN_CNT`.

Also, the comment block above is [duplicated verbatim](https://github.com/commaai/panda/blob/3dc21386239e3073a623156b75901aa302340d6c/board/drivers/can_common_declarations.h#L59-L69) in can_common_declarations.h. Probably makes sense to remove the copy here due to otherwise misleading references to CAN MCU interface numbers.

PR #2260, opened 2025-08-26
ci: stop power_monitor

maxime-desroches:


tests/setup_device_ci.sh
```diff
@@ -36,6 +36,8 @@ while true; do
   if ! sudo systemctl is-active -q ssh; then
     sudo systemctl start ssh
   fi
+
+  awk '{print \$1}' /proc/uptime > /var/tmp/power_watchdog
```
adeebshihadeh: can just kill it too

PR #2264, opened 2025-08-31
flashing speeeed

adeebshihadeh:
A full reflash is currently 26s on master 🤯
* 0.7s is import time
* 4s is sleeps
* 14s is `erase_sector` in the bootstub flash
* 2s is the sector erase in the app flash
* 2s is flashing the bootstub 256 bytes at a time (max of ST bootloader)
* 2.5s was flashing the app 16 bytes at a time

This gets it down to 9.5s. Fundamental limit here is the flash erasing speed on the STM32.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -126,6 +126,7 @@ pipeline {
               }
             }
 
+            /*
             stage('bootkick tests') {
               steps {
                 script {
@@ -133,6 +134,7 @@ pipeline {
                 }
               }
             }
+            */
           }
         }
       }


python/dfu.py
```diff
@@ -116,10 +116,11 @@ def reset(self):
   def program_bootstub(self, code_bootstub):
     self._handle.clear_status()
 
-    # erase all sectors
-    for i in range(len(self._mcu_type.config.sector_sizes)):
+    # erase bootstub + app sectors
+    for i in (0, 1):
```
robbederks, line 120: this is going to break silently again once we get a binary that's larger than these sectors, why not revert to the generic one?

robbederks: maybe we never had that calc in dfu: https://github.com/commaai/panda/blob/18230831f311a553ef64e55c1f418cb155eb3d7d/python/__init__.py#L432

adeebshihadeh: why would it break? we do the full erase in the bootstub; this erase just ensures the bootstub won't jump to the app

PR #2267, opened 2025-09-03
Cuatro fan fixes

probablyanasian:
Set correct settings and bypass PID controller for cuatro.


adeebshihadeh:
Let’s see if we can just do straight PWM for all.

board/boards/cuatro.h
```diff
@@ -132,8 +132,8 @@ static harness_configuration cuatro_harness_config = {
 board board_cuatro = {
   .harness_config = &cuatro_harness_config,
   .has_spi = true,
-  .fan_max_rpm = 12500U,
-  .fan_max_pwm = 99U, // it can go up to 14k RPM, but 99% -> 100% is very non-linear
+  .fan_max_rpm = 5000U,
+  .fan_max_pwm = 100U,
```
adeebshihadeh: we can also remove this now. it was only used here

board/drivers/fan.h
```diff
@@ -32,26 +34,16 @@ void fan_tick(void) {
     #endif
 
     // Cooldown counter
-    if (fan_state.target_rpm > 0U) {
+    if (fan_state.power > 0U) {
       fan_state.cooldown_counter = current_board->fan_enable_cooldown_time * FAN_TICK_FREQ;
     } else {
       if (fan_state.cooldown_counter > 0U) {
         fan_state.cooldown_counter--;
```
adeebshihadeh, line 40: we still need the cooldown?

adeebshihadeh (changes requested):

board/drivers/fan.h
probablyanasian: In tests/hitl/7_internal.py, comments say cooldown is to prevent noise on the tach line, so I think we still need it.

adeebshihadeh: hmm, shouldn't need without the stall recovery though?

probablyanasian: Measurement of the tach line (yellow) with cooldown removed, note noise after 12v fan power (pink) is turned off.

tests/hitl/7_internal.py
```diff
@@ -4,18 +4,21 @@
 from panda import Panda
 
 pytestmark = [
-  pytest.mark.test_panda_types(Panda.INTERNAL_DEVICES)
+  # TODO: re-enable this once we update the HITL devices
```
adeebshihadeh, line 7: FYI @maxime-desroches

PR #2275, opened 2025-09-10
robust SPI protocol check

adeebshihadeh:


adeebshihadeh:
It's always after the harness test...
```python
8_misc.py .                                                              [ 16%]
9_harness.py .                                                           [ 33%]
2_health.py EEEE                                                         [100%]
```

PR #2277, opened 2025-09-11
change pwm to 25kHz

probablyanasian:
Changed to align with fan pwm specifications.


adeebshihadeh:
Let's add a HITL test to check that the fan PWM-RPM curve is roughly linear?

probablyanasian:
Already implemented in hitl/7_internal.py unless you want smaller power intervals.

adeebshihadeh:
Ah nice, LGTM then.

PR #2294, opened 2025-10-24
Fix siren

briskspirit:


board/drivers/fake_siren.h
```diff
@@ -2,7 +2,36 @@
 
 #define CODEC_I2C_ADDR 0x10
 
-void fake_siren_init(void);
+// 1Vpp sine wave with 1V offset
+// cppcheck-suppress misra-c2012-8.9
```
robbederks: let's try to not any suppressions. doesn't look like there's a fundamental reason to not keep the LUT defined in the init function?

adeebshihadeh:
What happens to anything streamed over I2S? Does this properly ignore it and play the siren?

briskspirit:
> What happens to anything streamed over I2S? Does this properly ignore it and play the siren?

Yes, siren has higher priority, normal audio gets interrupted

robbederks:
This still has the "clicking" issue when the sine gets turned off, because the DMA isn't disabled at the start/end of a buffer. Can be resolved in a future PR

PR #2295, opened 2025-10-27
skip 1024 samples to settle, around 22ms

briskspirit:
Skip 21ms of a recording to let mic settle, removes "clipping" in the beginning of every recording.

### DIFF
--- a/board/stm32h7/sound.h
+++ b/board/stm32h7/sound.h
@@ -10,8 +10,7 @@ __attribute__((section(".sram4"))) static uint32_t mic_rx_buf[2][MIC_RX_BUF_SIZE
 #define MIC_SKIP_BUFFERS 2U // Skip first 2 buffers (1024 samples = ~21ms at 48kHz)
 static uint8_t sound_idle_count;
 static uint8_t mic_idle_count;
-static bool mic_recording_active;
-static uint8_t mic_skip_buffer_count;
+static uint8_t mic_buffer_count;
 
 void sound_tick(void) {
   if (sound_idle_count > 0U) {
@@ -26,7 +25,7 @@ void sound_tick(void) {
     mic_idle_count--;
     if (mic_idle_count == 0U) {
       register_clear_bits(&DFSDM1_Channel0->CHCFGR1, DFSDM_CHCFGR1_DFSDMEN);
-      mic_recording_active = false;  // Recording session ended
+      mic_buffer_count = 0U;
     }
   }
 }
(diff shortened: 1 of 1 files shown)


board/stm32h7/sound.h
```diff
@@ -23,6 +26,7 @@ void sound_tick(void) {
     mic_idle_count--;
     if (mic_idle_count == 0U) {
       register_clear_bits(&DFSDM1_Channel0->CHCFGR1, DFSDM_CHCFGR1_DFSDMEN);
+      mic_recording_active = false;  // Recording session ended
```
robbederks: should be able to use just a counter instead of a counter and a bool, no?

- set counter to 0 here when the mic is idle
- increment after each buffer has been processed (cap at MIC_SKIP_BUFFERS)
- in the process samples for loop, check if the counter is equal to MIC_SKIP_BUFFERS

should reduce the diff quite a bit I think

board/stm32h7/sound.h
```diff
@@ -31,13 +34,21 @@ void sound_tick(void) {
...
-  // process samples
-  uint8_t buf_idx = (((DMA1_Stream0->CR & DMA_SxCR_CT) >> DMA_SxCR_CT_Pos) == 1U) ? 0U : 1U;
-  for (uint16_t i=0U; i < MIC_RX_BUF_SIZE; i++) {
-    tx_buf[2U*i] = ((mic_rx_buf[buf_idx][i] >> 16U) & 0xFFFFU);
-    tx_buf[(2U*i)+1U] = tx_buf[2U*i];
+  if (mic_buffer_count < MIC_SKIP_BUFFERS) {
+    // Send silence during settling
+    mic_buffer_count++;
+    for (uint16_t i = 0U; i < MIC_TX_BUF_SIZE; i++) {
+      tx_buf[i] = 0U;
```
robbederks, line 43: Don't think `0U` is correct here? it's a signed signal inside of an unsigned buffer, so we probably want the actual signal "zero" point which should be roughly `(1U << 15)`

robbederks: other than that lgtm, if you confirm that the recordings actually have the zeroed out section for the expected duration

briskspirit: Your logic feels correct at first and that's what I was thinking too, but tests showed that it's incorrect.
With 0U I am getting nice silence at 0 level, but with 1U << 15 it's clipping to negative value.
Screenshot 2025-10-29 at 1 49 11 PM

robbederks: Not sure how, but okay

PR #2296, opened 2025-10-28
Fix mcu_type for deprecated pandas

pd0wm:
A recent refactor broke white pandas again. `get_mcu_type()` asserts on `F4` devices, but the `_mcu_type` variable is only referenced in the flashing functions.

In general this mcu_type thing seems a bit redundant now, since it's always `H7`, which then leads to an `H7Config`. I would also be happy to clean this up a bit more, but that requires quite a few changes.

### DIFF
--- a/python/__init__.py
+++ b/python/__init__.py
@@ -209,7 +209,6 @@ class Panda:
     self._serial = serial
     self._connect_serial = serial
     self._handle_open = True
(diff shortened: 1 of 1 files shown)


robbederks:
Seems fine as an intermediate fix. Agree that we can probably rip out `mcu_type` completely too, sounds like a good cleanup PR, will open an issue

PR #2306, opened 2025-12-15
PSA CAN ignition

elkoled:
Add CAN ignition for PSA vehicles.
EV and ICE models use two separate bits.

### DIFF
--- a/board/drivers/can_common.h
+++ b/board/drivers/can_common.h
@@ -201,6 +201,21 @@ void ignition_can_hook(CANPacket_t *msg) {
       ignition_can_cnt = 0U;
     }
 
+  } else if (msg->bus == 2) {
+    int len = GET_LEN(msg);
+
+    // PSA exception
+    if ((msg->addr == 0x432) && (len == 8)) {
+      int counter = msg->data[1] & 0xFU;
+
+      static int prev_counter_psa = -1;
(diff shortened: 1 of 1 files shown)


board/drivers/can_common.h
```diff
@@ -201,6 +201,13 @@ void ignition_can_hook(CANPacket_t *msg) {
       ignition_can_cnt = 0U;
     }
 
+    // PSA exception
+    if ((msg->addr == 0x348U) && (len == 8)) {
+      // bit 41: EV running, bit 42: ICE running
+      ignition_can = (msg->data[5] & 0x6U) != 0U;
+      ignition_can_cnt = 0U;
```
adeebshihadeh: does this message have a counter? ideally all new CAN ignition signals have a counter since we're seeing some collisions across brands now

adeebshihadeh (changes requested):

board/drivers/can_common.h
elkoled: it does not unfortunately.
image

adeebshihadeh: hmm is there another one we can use that has one?

elkoled: I will check, this one works with all PSA models I have seen.

PR #2328, opened 2026-02-12
update pyproject.toml: include panda.python and panda.board

pd0wm:
Installing the panda repo as a pip package (e.g. using `uv add git+https://github.com/commaai/panda.git`) would result in only the `__init__.py` in the root being installed.

This should also make importing consistent between openpilot and external programs, e.g. `from panda.python import Panda` now works everywhere.


pd0wm:
I think before https://github.com/commaai/panda/commit/cb11a40852d27910b14f254d4f13d7f6355af34b it kinda worked, since the python folder was included, but not the body/jungle code.

PR #2329, opened 2026-02-12
windows: fix fcntl import

pd0wm:
Like spidev, we need to gate the fcntl import as it's only available on *nix.

This can also be combined into a single try/except if needed. If one of them is not available, it's probably fine to mock out the other one too. E.g.

```python
try:
    import fcntl # No fcntl on Windows
    import spidev # No spidev on Windows/MacOS
except ImportError:
    fcntl = None # type: ignore
    spidev = None

```

### DIFF
--- a/python/spi.py
+++ b/python/spi.py
@@ -13,9 +13,9 @@ from .utils import logger
 
 # No fcntl on Windows
 try:
-    import fcntl
+  import fcntl
 except ImportError:
-    fcntl = None # type: ignore
+  fcntl = None # type: ignore
 
 # No spidev on MacOS/Windows
 try:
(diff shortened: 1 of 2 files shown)


python/spi.py
```diff
@@ -12,6 +11,13 @@
 from .constants import McuType, MCU_TYPE_BY_IDCODE, USBPACKET_MAX_SIZE
 from .utils import logger
 
+# No fcntl on Windows
+try:
+    import fcntl
```
adeebshihadeh: wrong indentation :P

adeebshihadeh (commented):
can you also add some simple windows CI that maybe just installs the package? it shouldn't be too aggressive, just enough that it's clearly a problem if it fails

PR #2380, opened 2026-04-04
body v1 updater

stefpi:
Body V1 offline flash script

add a version of the body v1 can flash in panda for dev flashing. Body v1 does not support usb flashing


adeebshihadeh:
Still relevant?

stefpi:
Not anymore, since updated the flasher in the body repo

PR #2387, opened 2026-04-10
make can hook v2 specific

stefpi:
make can ignition hook v2 specific

having this increases the time it takes for v1 ignition shut off time

### DIFF
--- a/board/drivers/can_common.h
+++ b/board/drivers/can_common.h
@@ -202,8 +202,8 @@ void ignition_can_hook(CANPacket_t *msg) {
     }
   }
 
-  // body exception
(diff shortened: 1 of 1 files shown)


adeebshihadeh (approved):

robbederks:
This new check matches traffic on my model S causing ignition to stay high. not sure if we can pick another address or be more specific in this hook?

stefpi:
@robbederks thanks for the catch, I should be more specific with length of message: https://github.com/commaai/panda/pull/2388

can you please check if this works? If not I will change the address as well

robbederks:
will have to double check, but pretty sure this is fine. There aren't a ton of 1-byte messages

sshane:
Why wasn't this addressed faster?

PR #2388, opened 2026-04-18
make body ignition hook more specific

stefpi:
add length checking


sshane:
Need a test in test_models! Too easy to write these bugs. But maybe first moving can ignition to opendbc

adeebshihadeh:
Moved into opendbc

PR #2390, opened 2026-04-22
Disable body ignition check temporarily

sshane:
Stops Teslas from generating routes every few seconds

Related: https://github.com/commaai/panda/pull/2388

### DIFF
--- a/board/drivers/can_common.h
+++ b/board/drivers/can_common.h
@@ -202,11 +202,12 @@ void ignition_can_hook(CANPacket_t *msg) {
     }
   }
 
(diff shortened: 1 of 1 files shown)


board/drivers/can_common.h
```diff
@@ -202,11 +202,12 @@ void ignition_can_hook(CANPacket_t *msg) {
     }
   }
 
+  // TODO: this is too loose, Teslas have 0x222
```
sshane, line 205:
```suggestion
  // FIXME: this is too loose, Teslas have 0x222
```

PR #2393, opened 2026-04-26
bootkick: wake on CAN

dzid26:
Separates SOM waking up from ignition (going on-road) logic. Optional, but lets the device start booting early. 

For now this is only wired for Tesla. The comma device will wake the moment a door is opened or the car is woken up remotely (e.g. with _Keep Accessory Power On_).

SOM shutdown is still controlled by openpilot - if ignition was not seen the device will turn off after [1hr](https://github.com/commaai/openpilot/blob/d1e069210fc38a2c0752f04bed05074eaf3704ff/system/hardware/power_monitoring.py#L125) - waiting for the driver to start driving.
If the car sits in accessory mode for long time (>1hr), the the ignition signal or car restart will be needed to boot again.

### DIFF
--- a/board/drivers/bootkick.h
+++ b/board/drivers/bootkick.h
@@ -2,18 +2,18 @@
 
 bool bootkick_reset_triggered = false;
 
-void bootkick_tick(bool ignition, bool recent_heartbeat) {
+void bootkick_tick(bool wake_up, bool recent_heartbeat) {
   static uint16_t bootkick_last_serial_ptr = 0;
   static uint8_t waiting_to_boot_countdown = 0;
   static uint8_t boot_reset_countdown = 0;
   static uint8_t bootkick_harness_status_prev = HARNESS_STATUS_NC;
-  static bool bootkick_ign_prev = false;
+  static bool bootkick_wake_up_prev = false;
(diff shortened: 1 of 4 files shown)


adeebshihadeh:
What's the use case here exactly?

dzid26:
Currently, main use case is quality of life improvement - wake the device up when the car wakes up - comma device is ready to use earlier.
Secondary use case for Tesla - waking up the car remotely turns on the device, useful for development.

-----

Additionally, I think it should always have been the case that panda is in charge _only_ of waking up the SOM. Today the CAN ignition bit is doing double duty (wake + onroad/offroad gating). Decoupling them would allow removing CAN ignition detection from panda entirely. We could potentially wake up early probably for most cars using CAN activity rate threshold (instead of specific message id) and let openpilot handle fingerprinting and onroad/offroad transitions using richer inputs for robustness and safety.

adeebshihadeh:
I'm going to close this for two reasons:
1) I don't really see a concrete benefit to this.
2) I think it's very likely that we'll have tons of false positive wake ups on some cars.

> Additionally, I think it should always have been the case that panda is in charge _only_ of waking up the SOM. 

This is an interesting idea in the abstract... though I can't immediately think of anything this improves.

PR #2395, opened 2026-04-30
VW MEB can ignition

elkoled:
TODO: 
- [ ] check if adding counter is feasible


adeebshihadeh:
Let's just get CAN ignition moved over to opendbc while we're at it?

PR #2405, opened 2026-05-21
Add MISRA check for integer bool literals

dhrubasumatary:
## Summary
- add a cppcheck addon to report integer 0/1 literals assigned to bools
- run the addon with the existing MISRA check
- fix existing Panda bool literal assignments and add a mutation case

## Verification
- tests/misra/test_misra.sh
- pytest -q tests/misra/test_mutation.py


adeebshihadeh:
Closing in favor of #2409

PR #2455, opened 2026-09-21
mic: clear output buffers before startup and after idle

sshane:
Microphone output buffers live in `.sram4 (NOLOAD)`, outside the BSS cleared at boot. Output DMA can transmit that stale memory before the microphone's settling buffers arrive. Clear both output buffers before enabling output DMA, and again when the microphone becomes idle so the next capture starts with silence.

Four added lines in `board/stm32h7/sound.h` (two `memset` calls, one comment, one blank line). This isolates the startup-memory defect from #2454; it does not change clock matching, gain, filtering, or audio format and does not fix recurring repeated-block clicks (handled separately in #2456). Openpilot integration is commaai/openpilot#38984.

Validation: H7 firmware build passes. An actual-header host check with mocked registers and poisoned buffers verifies both buffers are zero before output DMA is enabled and after idle. Earlier device captures in #2454 reproduced the startup artifact and its removal with buffer initialization; this newly isolated branch has not been reflashed for a fresh device test.


board/stm32h7/sound.h
```diff
@@ -191,6 +192,9 @@ void sound_init(void) {
   register_set(&DMAMUX2_Channel0->CCR, 16U, DMAMUX_CxCR_DMAREQ_ID_Msk); // SAI4_B_DMA
   register_set_bits(&BDMA_Channel0->CCR, BDMA_CCR_EN);
 
+  // SRAM4 is not cleared at startup; transmit silence until capture is ready.
```
sshane:
```suggestion
  // SRAM4 is not cleared at startup; transmit silence until capture is ready
```

robbederks:
lgtm

PR #2456, opened 2026-09-21
mic: match capture to the independent I2S output clock

sshane:
The microphone decimator produces approximately 47,952 samples/s while the independent I2S clock consumes approximately 48,000. Copying whole input blocks into whichever output buffer is inactive eventually retransmits an old 512-sample block. Baseline desk captures reproduced this about every 10.53 seconds; changing the nominal divider alone still failed a longer capture.

Fill output buffers on I2S DMA completion and continuously read the input ring at a rate corrected from buffer occupancy. Retain the previously tested cubic interpolation, saturation, and one-block fade when establishing a read position. Original microphone clock, decimation, gain, buffer sizes, and I2S format remain unchanged.

This extracts only the clock-matching implementation from #2454: **79 added / 14 removed lines in one file**, with no investigation tools or recordings. Startup output-memory initialization is separate in #2455; openpilot integration is commaai/openpilot#38984. This is a resampler and buffer-control change, not a one-line clock-divider fix.

Validation:
- H7 firmware build and the repository's intended MISRA checker pass on this extracted branch.
- Actual-C UBSan simulation passes 320 seconds at each of five input clock offsets (-2000, -1000, 0, +20, +2000 ppm), with producer-position jitter and a continuous 1 kHz test tone.
- The same clock-matching implementation was tested on-device during #2454, including a 300-second run without repeated blocks or resynchronization. Those captures also included the separate startup fix; no fresh device capture was made for this extraction.

Remaining review: concurrent DMA/interrupt behavior under normal on-road workload, added capture latency, and resampling response. This addresses the reproduced repeated-block defect; it does not claim to eliminate every acoustic transient or the separate unexplained driving discontinuity.


sshane:
Replaced by https://github.com/commaai/panda/pull/2457

PR #2457, opened 2026-09-21
mic: match capture to I2S using the audio PLL

sshane:
The microphone produces about 47,952 samples/s while I2S consumes about 48,000. The drift eventually replays old 512-sample buffers as clicks.

Use PLL2 as the microphone clock and adjust its fractional divider from the outgoing DMA position. This keeps buffer completions aligned while copying PCM unchanged. CPU/CAN clocks and microphone filter settings stay the same.

**+18 / -1 lines in sound.h.** Startup buffer clearing is separate in #2455. This is a smaller alternative to the software resampler in #2456.

Validation: a fresh baseline reproduced six buffer replays in 60 seconds. The PLL implementation had zero in a five-minute capture and four restart recordings, with no audio overruns or firmware faults. Those device tests also included #2455; this PR now excludes it. The clock-only branch passes the H7 build, MISRA and actual-handler UBSan checks. The simplified expression produces the same PLL setting at every valid DMA position (0–1,024). On-road CAN/interrupt-load validation remains.


sshane:
[index.html](https://github.com/user-attachments/files/32546035/index.html)

